Issue #163, opened 2022-04-19
Share to function?


poVoq:
Really strange that neither this nor Lemmur support integration with Android to share links to Lemmy from a browser etc., i.e. that I can use the share button of Firefox (or any other app) to open Jerboa to make a new post on Lemmy with the link (or the selected text) already pasted into the new post form.

At least with a Lemmy webapp that is really easy to add and probably one of the most useful features for posting stuff.

dessalines:
Lemmy-ui doesn't have this, and it would be difficult to add for any federated service where the domain name could be anything.

Right now you do the same thing you have to do for lemmy-ui, which is to copy-paste any link urls into the create post URL field.

I'm not sure how this would be possible in any way for unpredictable lemmy domains.

poVoq:
Uhm, maybe I am missing something but Lemmy-UI does have this. I used it in my uLemmy Ubuntu Touch app (currently unmaintaned as I broke my Nexus5).

It is a URL addition: `/create_post?body=` and `/create_post?url=`

dessalines:
I have no idea how its able to do this, I didn't code anything for it. That URL link would work, but android is not going to know what app to open when the URL could be literally anything.

poVoq:
Hmm, I guess that only works with webapps then?

Android knows of course which app to use as when Jerboa is registered as a "share to" end-point for URLs, then the user can select it from the list of apps that opens when you click on the share button.

dessalines:
@LunaticHacker completed this.

PR #178, opened 2022-06-29
Add deeplinks for createPost

LunaticHacker:
for #163,

* Jerboa appears in share sheet for plain text
* URLs and other text are handled in createPost Deeplink

we should be able to handle sharing images too but i don't understand how the PickImage thing works :)


dessalines (changes requested):
I tested this with a few browsers, and it seems to copy both the title and URL into the body field. It needs to copy the title -> title, and URL -> url

LunaticHacker:
> I tested this with a few browsers, and it seems to copy both the title and URL into the body field. It needs to copy the title -> title, and URL -> url

That would be ideal. but we can't determine what is title and url from plain text, only whether the text is a url or not

app/src/main/java/com/jerboa/ui/components/post/create/CreatePostActivity.kt
```diff
@@ -33,6 +33,8 @@ fun CreatePostActivity(
     navController: NavController,
     communityListViewModel: CommunityListViewModel,
     postViewModel: PostViewModel,
+    _url: String,
+    _body: String,
```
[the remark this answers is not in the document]
LunaticHacker: Yeah but passing empty string makes the implementation cleaner in this particular instance

app/src/main/java/com/jerboa/ui/components/post/create/CreatePost.kt
```diff
@@ -97,6 +97,9 @@ fun CreatePostBody(
     val nameField = validatePostName(name)
     val urlField = validateUrl(url)
 
+    // To fetch suggested title when opened from deeplink
+    onUrlChange(url)
+
```
LunaticHacker, line 102: This is probably wrong, but i couldn't find a better way to do this

app/src/main/java/com/jerboa/MainActivity.kt
```diff
@@ -168,14 +174,40 @@ class MainActivity : ComponentActivity() {
...
+                        val text = activity?.intent?.getStringExtra(Intent.EXTRA_TEXT) ?: ""
+                        val image =
+                            activity?.intent?.getParcelableExtra<Parcelable>(Intent.EXTRA_STREAM) as? Uri
+                        // url and body will be empty everytime except when there is EXTRA TEXT in the intent
+                        var url = ""
+                        var body = ""
+                        if (Patterns.WEB_URL.matcher(text).matches()) {
+                            url = text
+                        } else {
+                            body = text
```
dessalines, line 197: I'm in the process of testing this out, but from what I can tell, when using share via an android browser, it puts the title on the first line, and the URL on the second. Maybe you could do a line split for that case.

Also, I think adding _title would be better than _body, right?

app/src/main/java/com/jerboa/ui/components/common/PictrsImage.kt
```diff
@@ -152,6 +153,24 @@ fun PickImage(
...
+                    .Media.getBitmap(ctx.contentResolver, imageUri!!)
+            } else {
+                val source = ImageDecoder
+                    .createSource(ctx.contentResolver, imageUri!!)
+                bitmap.value = ImageDecoder.decodeBitmap(source)
+            }
+
+            Log.d("jerboa", imageUri.toString())
+            onPickedImage(image)
+        }
```
dessalines, line 171: Have you tested this? For me it doesn't actually seem to be doing the image upload.

app/src/main/java/com/jerboa/MainActivity.kt
LunaticHacker: What test data are you working with? The text could be any plain text from a single word to large chunks of paragraphs. it has no guaranteed structure. in most cases users will be using 'share link' which will just send the url that's what is handled on line 194.

> Also, I think adding _title would be better than _body, right?

If someone is copying an entire article to share you wouldn't want to fill the title with it.

I'm not sure how you managed to get both title and url, you might be using a share button from a website that uses some api to invoke the sharesheet and send formatted data

app/src/main/java/com/jerboa/ui/components/common/PictrsImage.kt
LunaticHacker: Tested it on firefox and from my gallery, note that when sharing from browsers you have to use 'share image' just sharing the link won't do, but it makes sense to detect if the link is an image, i'll add that shortly

app/src/main/java/com/jerboa/MainActivity.kt
dessalines: I'm using several android browser's share button, which to me seems like the most common use case. You're on a site, and you want to share it to lemmy. That to me seems much more likely than selecting random text to share as a body of a lemmy post.

But also, you could still differentiate the two, by counting lines and seeing if the 2nd line is a URL.

app/src/main/java/com/jerboa/ui/components/common/PictrsImage.kt
dessalines: My bad, this does work correctly. Not sure why it didn't the first time for me.

app/src/main/java/com/jerboa/MainActivity.kt
dessalines: Here's some findings:

- Chrome - only shares the URL.
- Via - Shares two lines, first one is title, 2nd is URL
- Puffin - only shares the URL
- Firefox - only URL

In that case, I'll say this is good, because via seems to be the only one that does the 2-line method.

app/src/main/java/com/jerboa/MainActivity.kt
```diff
@@ -168,14 +174,40 @@ class MainActivity : ComponentActivity() {
                             selectMode = it.arguments?.getBoolean("select")!!
                         )
                     }
-                    composable(route = "createPost") {
+                    composable(
+                        route = "createPost",
+                        // TODO:  Handle Image mimetype too
```
dessalines, line 179: You can remove this comment now.

app/src/main/java/com/jerboa/ui/components/common/PictrsImage.kt
```diff
@@ -152,6 +153,24 @@ fun PickImage(
         mutableStateOf<Bitmap?>(null)
     }
 
+    if (image != null) {
+        LaunchedEffect(image) {
+            imageUri = image
+            // Set the bitmap
```
dessalines, line 159: This looks like copy-pasted code, could you extract that to a function.

app/src/main/java/com/jerboa/ui/components/common/PictrsImage.kt
```diff
@@ -152,6 +153,24 @@ fun PickImage(
         mutableStateOf<Bitmap?>(null)
     }
 
+    if (image != null) {
```
dessalines, line 156: God I hate kotlin null safety. Rust options are so much better.

dessalines (changes requested):

dessalines:
Oops I forgot one more thing: make sure when doing a URL share, that it tries to fetch the post title

dessalines (approved):
Works great, thx a ton!

PR #180, opened 2022-07-02
fix signout

### LINKED ISSUE #179: singout isn't handled properly

LunaticHacker:
fix #179


LunaticHacker:
is there a way to force the lint check locally?

dessalines:
We use husky for lemmy-ui ( and used to use cargo-husky for lemmy ), but it doesn't look like there's a husky equivalent for gradle projects :( . 

You might just be able to add `./gradlew ktlintCheck` to `.git/hooks/pre-commit`

dessalines (approved):
Thanks a lot for this BTW, I'm feeling spread so thin, and jerboa needs a lot of work. I'd rather spend more time on jerboa than lemmy-ui ( kotlin is better to work with than typescript, and priority should be put on native apps and not the web .

Issue #88, opened 2022-02-07
Settings page


dessalines:
Settings ideas:

- Look and feel
  - Font Size
  - Theme
- Account
- About

PR #183, opened 2022-07-12
Add api settings

LunaticHacker:
for #88 

for now this is just the api settings, we can add app-specific settings and account actions(password reset, account deletion) to this page in separate PRs


LunaticHacker:
sorry for the force pushes, local linting doesn't pickup any codestyle errors is there some config file i need to add?

dessalines:
You can just run `./gradlew ktlintFormat` before your commit. I should be able to test this later today.

app/src/main/java/com/jerboa/ui/components/settings/Settings.kt
```diff
@@ -0,0 +1,407 @@
...
+
+    val backgroundColor = MaterialTheme.colors.primarySurface
+    val contentColor = contentColorFor(backgroundColor)
+
+    TopAppBar(
+        title = {
+            Text(
+                text = "Settings",
+            )
+        },
```
dessalines, line 59: Could you move the Save button into this header `actions` ? An example of this is in PostEdit.kt

app/src/main/java/com/jerboa/ui/components/settings/Settings.kt
```diff
@@ -0,0 +1,407 @@
...
+                    Icons.Filled.ArrowBack,
+                    contentDescription = "Back"
+                )
+            }
+        },
+    )
+}
+
+@Composable
+fun SettingsField(
```
dessalines, line 79: Maybe this should be SettingsTextField. 

Also, Bio is a markdown field, there's a composable for it in common/TextFields.kt .

app/src/main/java/com/jerboa/ui/components/settings/Settings.kt
```diff
@@ -0,0 +1,407 @@
...
+    var showBotAccount by rememberSaveable { mutableStateOf(luv?.local_user?.show_bot_accounts) }
+    var botAccount by rememberSaveable { mutableStateOf(luv?.person?.bot_account) }
+    var showReadPosts by rememberSaveable { mutableStateOf(luv?.local_user?.show_read_posts) }
+    var showNewPostNotifs by rememberSaveable { mutableStateOf(luv?.local_user?.show_new_post_notifs) }
+    var sendNotificationsToEmail by rememberSaveable { mutableStateOf(luv?.local_user?.send_notifications_to_email) }
+    val form = SaveUserSettings(
+        display_name = displayName,
+        bio = bio,
+        email = email,
+        auth = account?.jwt,
```
dessalines, line 258: Apologies, I just checked the API, auth is a required field, so change that form to `val auth: String`

app/src/main/java/com/jerboa/ui/components/settings/SettingsActivity.kt
```diff
@@ -0,0 +1,53 @@
...
+
+    Surface(color = MaterialTheme.colors.background) {
+        Scaffold(
+            scaffoldState = scaffoldState,
+            topBar = {
+                SettingsHeader(
+                    navController = navController,
+                )
+            },
+            content = {
```
dessalines, line 36: You might as well do a `account?.also { acct -> ...` here, account is required to save the settings anyway.

app/src/main/java/com/jerboa/ui/components/settings/SettingsViewModel.kt
```diff
@@ -0,0 +1,33 @@
...
+        form: SaveUserSettings,
+        ctx: Context,
+        siteViewModel: SiteViewModel,
+        account: Account?
+    ) {
+
+        viewModelScope.launch {
+            loading = true
+            saveUserSettingsWrapper(form, ctx)
+            siteViewModel.fetchSite(account?.jwt, ctx)
```
dessalines, line 29: Hrm... I guess this makes sense. It does need to re-fetch your info after a save.

dessalines (commented):
Looks really good, thanks a ton for this. Just a few things.

app/src/main/java/com/jerboa/ui/components/settings/Settings.kt
LunaticHacker: Is this necessary? To do this we will have to store the form data in the SettingsViewModel which will make the implementation very ugly because of the number of items in this form.

Also i was thinking of making this a tabbed page when adding other types of settings, this wouldn't work in that case

app/src/main/java/com/jerboa/ui/components/settings/Settings.kt
LunaticHacker: The MarkdownTextField is very specific to CommentReply, let's keep it like this for now and  make it generic some other time.

dessalines: mmk, but add a TODO.

app/src/main/java/com/jerboa/ui/components/settings/Settings.kt
dessalines: I wouldn't do tabs, most settings I see in apps have lists for different settings categories, and then those categories have their own activities. I spose its fine to leave as is, but it wouldn't be difficult to pass in the whole form as a single var to the viewmodel.

I'm looking at a few other apps, and the way they seem to do it, which I like a lot, is to not have any save buttons at all, simply do a save immediately after a change. The only field that would be weird for is bio, which we might wanna add a debounce for, but there are already examples of this in the code.

For now you can just put a TODO, to remove the save button.

dessalines:
Just add a few TODOs, but other than that, it looks good.

dessalines (approved):

PR #320, opened 2023-02-09
Fix the backspace bug by setting KeyboardType to 'Email'

### LINKED ISSUE #148: Backspacing on a word results sometimes deletes the space behind the word.
I’m on Android.

When I use backspace on a word, it sometimes removes the space behind the word.

### LINKED ISSUE #194: Issues with typing (backspace)
When pressing the backspace several times, sometimes the last word would go back one character and attach itself to the one behind it.

Say I've written the following sentence:

> I can't believe you've done this

If I delete the last 5-6 characters, the following would happen:

> I can't believe you've**don** 

A temporary bypass is to select the characters that you want and delete them all at once.

Kradyz:
A proposed approach to eliminate the backspace bug. Fixes #194, fixes #148 

The autoCorrect is triggering the following undesirable behavior for text fields:

Given:

"Hello world <Cursor position>"

Pressing backspace two times will result in:

"Helloworl"

An attempt was made to turn autoCorrect off, but the approach failed because the KeyboardType.Text does not respect the autoCorrect = false setting. ([See documentation](https://developer.android.com/reference/kotlin/androidx/compose/foundation/text/KeyboardOptions#autoCorrect()))

The proposed solution is to change the KeyboardType to KeyboardType.Email, which does respect the autoCorrect = false setting.

I have tested this approach over a period of a few days. The backspace bug disappears, and I did not experience any negative effects from using the Email keyboard type. 

I looked into how to apply the autoCorrect = false setting to KeyboardType.Text, but I could not find how to do that.

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/common/InputFields.kt
+++ b/app/src/main/java/com/jerboa/ui/components/common/InputFields.kt
@@ -122,8 +122,8 @@ fun MarkdownTextField(
                 modifier = modifier.focusRequester(focusRequester),
                 keyboardOptions = KeyboardOptions.Default.copy(
                     capitalization = KeyboardCapitalization.Sentences,
-                    keyboardType = KeyboardType.Text
-                    // autoCorrect = true,
+                    keyboardType = KeyboardType.Email,
+                    autoCorrect = false
                 )
             )
         } else {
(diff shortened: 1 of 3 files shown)


dessalines (changes requested):
Grep for `TextField(`, because there are a few more you missed here.

I'm seeing ones in `CommunityList.kt`, PostEdit.kt, CreatePost.kt, a few more in Login.kt, and InputFields.kt

Might also be good to create / use one in `InputFields.kt`, and just use that same simple one around.

dessalines:
@oscarnylander Do you have any experience with these strange keyboard autocorrect / backspace bugs ppl (including me) have been having with jetpack compose?

oscarnylander:
Unfortunately, no. I didn't encounter it during my time working professionally with Compose, and nowadays I only do Compose on the side.

dessalines:
Gotcha, thx!

PR #358, opened 2023-04-24
Adjust the status bar while using the system dark theme to not use dark icons

russjr08:
Hello!

I noticed while using Jerboa that while my phone is using the dark theme, the status bar icons within the app are also dark which makes it seem a bit out of place. I'm opening this pull request to try to resolve this issue. I'm pretty new to the Jetpack Compose toolkit, so hopefully this approach looks appropriate. The following are screenshots from before and after to demonstrate the change:

Before:

image

After:

image

I tested this on my physical Pixel 6a running Android 13, however as a precaution I went ahead and also tested this on a virtual Pixel 3a running Android 11, just to make sure there were no introduced incompatibility changes.

### DIFF
--- a/app/src/main/java/com/jerboa/ui/theme/Theme.kt
+++ b/app/src/main/java/com/jerboa/ui/theme/Theme.kt
@@ -1,15 +1,19 @@
 package com.jerboa.ui.theme
 
+import android.app.Activity
 import android.os.Build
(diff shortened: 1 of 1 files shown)


dessalines (commented):
Looks good! Run `./gradlew formatKotlin` for this to pass lint.

russjr08:
Ah, got it, thanks!

PR #375, opened 2023-06-03
Show settings in sidebar even when user is not logged in

twizmwazin:
Right now the only way to access the settings menu is to first log in to an account. This change shows the settings menu button in the drawer even when not logged in.


dessalines:
Seems fine, thx!

dessalines (approved):

PR #383, opened 2023-06-03
Remove Splash Screen

oscarnylander:
New default route is now `home`, which the splash screen used to navigate to automatically.

Motivation:

There's a bug causing the splash screen to relaunch when pressing the back button (#382). Removing the splash screen remedies this issue.

There's some cosmetic loss in that the current splash screen never shows any more, and some performance benefit in that users no longer need to wait 0.7 seconds before the app content shows.

The cosmetic benefits of a splash screen can be regained by implementing the `SplashScreen`-API: https://developer.android.com/develop/ui/views/launch/splash-screen


dessalines (approved):
Looks good, thx!

Run `./gradlew formatKotlin` and push that.

oscarnylander:
Done

Issue #384, opened 2023-06-03
Introduce automatically applicable code style standards


oscarnylander:
`ktlint` can provide automatically applicable code style standards, which will enhance the code workflow and prevent style-related bike shedding issues.

dessalines:
We already have kotlinter running in ci.

PR #385, opened 2023-06-03
Add `ktlint`

oscarnylander:
This solves #384

Motivation:

Automatically applicable code style standards:

- Help enhance the code workflow, in that you can automatically format any code added to an independently verifiable standard
- Prevent style-related bike shedding issues, as there is an independent tool responsible for all formatting rules

An additional potential enhancement would be to add a Github Action to check style on new PRs.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+root = true
+
+[*.{kt,kts}]
+
+indent_size = 4
+insert_final_newline = true
+max_line_length = 120
+end_of_line = lf
+
+ij_kotlin_allow_trailing_comma = true
+ij_kotlin_allow_trailing_comma_on_call_site = true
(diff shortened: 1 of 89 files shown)


oscarnylander:
Upon further inspection, I noticed that `kotlinter` was already added to the project, with a slightly different configuration.

I guess what we could do for this PR is:

- Discuss the merits of the difference in configuration
- Remake the PR to have an agreed upon configuration and reformat the project using that configuration

In particular, I guess we could discuss the following:

- Wildcard imports: I find they introduce some ambiguity in what gets resolved, and as such I've tended to avoid them
- Trailing commas: I prefer these, as they make introducing a new line in a git diff only cover one line (the new content) instead of two (the new content + the newly introduced comma)

dessalines:
Yup, this already uses kotlinter, which has trailing commas.

And I'll do any format action commits, but feel free to do prs to add the configs.

oscarnylander:
Fair enough. See #405

PR #387, opened 2023-06-03
Fix dropdowns on Account Settings-screen not working

oscarnylander:
Fixes #386

PR #389, opened 2023-06-03
Fixed ExposedDropdownMenuBox not expanding in Settings menu when clicked

a1studmuffin:
Small bugfix for the Settings menu dropdowns (Default Listing Type, Default Sort Type) not expanding when clicked. Used the solution provided here:

https://stackoverflow.com/questions/73946042/jetpack-compose-exposeddropdownmenu-not-showing-up-when-pressed


a1studmuffin:
Ha, looks like https://github.com/dessalines/jerboa/pull/387 beat me to it! Closing.

PR #392, opened 2023-06-04
Added Blue theme color

a1studmuffin:
Added Blue theme color. Screenshots compared to Pink for reference:

blue-dark
blue-light
pink-dark
pink-light


dessalines (approved):
Thx!

Issue #376, opened 2023-06-03
Black theme for OLED displays, and some more theme color options?


a1studmuffin:
I saw in a previous commit that there was once a Black theme alongside Light and Dark. This would be great to reinstate for OLED displays (and just for aesthetic preference). I imagine it would effectively look the same as Dark theme but with true black (#000000) as the background.

On a similar note it would be nice to have a greyscale theme color option in addition to Green and Pink.

If this isn't high on the priority list, I'm happy to have a go at putting together a PR, just let me know. Thanks!

dessalines:
Feel free to add as many themes as you like! jetpack compose handles material v4 themes really well, so its nbd to add these.

PR #393, opened 2023-06-04
Added Black and System Black theme modes, fixes https://github.com/dessalines/jerboa/issues/376

a1studmuffin:
Added Black theme mode. Looks exactly like the Dark theme mode except near-black background colors are true black. Also added theme mode SystemBlack, which uses Light/Black instead of Light/Dark depending on system setting. Fixes https://github.com/dessalines/jerboa/issues/376.

green-black

theme_mode


dessalines (approved):
Looks good. Run `./gradlew formatKotlin` for this to pass lint.

PR #404, opened 2023-06-04
Roll back AGP to `8.0.2`

oscarnylander:
This makes it so that an alpha-build of Android Studio is not necessary to contribute to the project.


dessalines (approved):
Mmmk this did work with android studio canary also, which surprised me, because it always gives me warnings about not using the most updated AGP.

PR #405, opened 2023-06-04
Add `.editorconfig`

oscarnylander:
`ktlint` follows the rules in this file.

For the sake of simplicity, the rules in the file are aligned with what's in place in the project already. No additional formatting is necessary after this PR is merged.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,7 @@ plugins {
     id 'com.android.library' version '8.0.2' apply false
     id 'org.jetbrains.kotlin.android' version '1.8.20' apply false
     id "com.github.ben-manes.versions" version "0.42.0"
-    id 'org.jmailen.kotlinter' version "3.13.0" apply false
+    id 'org.jmailen.kotlinter' version "3.15.0" apply false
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -16,9 +16,6 @@ plugins {
 
 subprojects {
   apply plugin: "org.jmailen.kotlinter" // Version should be inherited from parent
-  kotlinter {
```
dessalines, line 19: Upgrade kotlinter in L13 to `3.14.0`, and this should pass.

oscarnylander: Weird that this worked for me locally. Perhaps I didn't run `./gradlew lintKotlin` after removing the `kotlinter`-config.

Anyway, if I upgrade to either `3.14.0` or `3.15.0` we get other rule violations, as those versions enforce additional rules. All of those rules are automatically formattable, but as per the last PR, I'll leave that formatting commit to you.

You can create it by calling `./gradlew formatKotlin` - it mostly adds trailing commas to basically everything.

dessalines: K thx, I'll do that shortly.

PR #406, opened 2023-06-04
Fixing CI ntfy.sh notification link.

dessalines:


.woodpecker.yml
```diff
@@ -20,6 +20,6 @@ pipeline:
     image: alpine:3
     commands:
       - apk add curl
-      - "curl -d'Thumb-Key build ${CI_BUILD_STATUS}: ${CI_BUILD_LINK}' ntfy.sh/thumbkey_ci"
+      - "curl -d'Jerboa build ${CI_BUILD_STATUS}: ${CI_BUILD_LINK}' ntfy.sh/jerboa_ci"
```
binwiederhier, line 23: I dude, I love Lemmy. Let me know if you want a ntfy Pro subscription so you can reserve that topic. :heart:

dessalines: Whoa thanks! And thank you for creating ntfy! At some point we will add ntfy notifs into the lemmy back-end, so i'll get your expertise when we start on that.

Issue #400, opened 2023-06-04
Swipe-Refresh indicator gets obscured by top bar on Home-Screen


oscarnylander:
Expected behaviour is that the Swipe-Refresh indicator starts from the first point where it would be visible - in this case, from the top bar.

This might be fixable by introducing an offset.

oscarnylander:
Regression in 0.0.35 - this no longer works.

PR #408, opened 2023-06-04
Add padding to SwipeRefresh for PostListings (for #400)

russjr08:
This pull request attempts to partially resolve the issue listed in #400 - I mention partially because from what I can tell, this occurs in two places:

- Any place where there is a list of posts (the `PostListings` component)
- The list of comments on a post (the `PostActivity` composable)

I was able to fix the first occurrence by just simply passing in the padding from the parent `PostListings` component (assuming I'm using the correct phrasing here), but for the second occurrence that trick doesn't quite work as I can't really find a place to extract the padding from.

While it would be possible to hardcode a padding value, that seems quite... hacky - and additionally, when I took a look at the screen from the Layout Inspector on two different devices (my Pixel 6a, and a Pixel 3a XL emulator) the top header was two different sizes. I presume this is because of ["safe areas"](https://developer.android.com/jetpack/compose/layouts/insets) due to some devices (like my 6a) having the camera notch/cut-out whereas some devices don't (such as the 3a XL).

I'm not sure what the elegant way of fixing the second occurrence would be since I still have a long way to go with learning Compose, but I figured I could at least toss my hat in for fixing the first occurrence if that's alright 👍🏼


app/src/main/java/com/jerboa/ui/components/post/PostListings.kt
```diff
@@ -57,7 +57,8 @@ fun PostListings(
 ) {
     SwipeRefresh(
         state = rememberSwipeRefreshState(loading),
-        onRefresh = onSwipeRefresh
+        onRefresh = onSwipeRefresh,
+        indicatorPadding = padding
     ) {
```
dessalines, line 62: We've got to get rid of these at some point anyway, because swiperefresh is deprecated.

dessalines (approved):

PR #412, opened 2023-06-04
Add option to disable showing the bottom navigation bar

twizmwazin:
This adds a checkbox option to disable the bottom navigation bar throughout the app. Testing locally it seems to work correctly, including migrations.

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/common/AppBars.kt
+++ b/app/src/main/java/com/jerboa/ui/components/common/AppBars.kt
@@ -43,6 +43,7 @@ import com.jerboa.datatypes.api.GetUnreadCountResponse
 import com.jerboa.datatypes.samplePersonSafe
 import com.jerboa.datatypes.samplePost
 import com.jerboa.db.Account
+import com.jerboa.db.AppSettings
 import com.jerboa.loginFirstToast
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/jerboa/ui/components/common/AppBars.kt
```diff
@@ -76,6 +77,7 @@ fun SimpleTopAppBar(
 @Composable
 fun BottomAppBarAll(
     navController: NavController = rememberNavController(),
+    appSettings: AppSettings? = null,
```
dessalines, line 80: Since showBottomNav is the only one that's necessary, just pass that in rather than all the settings.

dessalines (commented):
Just that one thing. Thanks for this!

dessalines:
Oops, accidentally merged this early.  I'll fix that issue I mentioned.

twizmwazin:
I do plan to keep adding various UI settings to customize the UI to my liking, so I think for the time being passing the whole AppSettings object isn't a bad idea, just to reduce the amount of refractors that are necessary for each incremental option.

dessalines:
No problem, we can always change that back then.

Issue #395, opened 2023-06-04
App crashes when navigating to linked communities in other instances


Anna-log7:
The app crashes/restarts seemingly when clicking on a link to a community in a different instance. For example, as a beehaw user clicking on the link in this comment: https://lemmy.ml/comment/445530

- Fix a crash upon clicking links to communities in other instances

mounderfod:
I am able to reproduce this problem

dessalines:
I'm just going to have to disable deep linking, because of all the bugs its causing.

PR #413, opened 2023-06-04
Fix navigating to other instance communities and users

Anna-log7:
This fixes an issue where the app crashes/restarts when a user or community in a different instance is clicked within a comment.
I've changed the routes slightly so the name given ends up being `name@instance` instead of just `name`

Fixes [#395](https://github.com/dessalines/jerboa/issues/395)

### DIFF
--- a/app/src/main/java/com/jerboa/MainActivity.kt
+++ b/app/src/main/java/com/jerboa/MainActivity.kt
@@ -180,19 +180,23 @@ class MainActivity : ComponentActivity() {
                     }
                     // Only necessary for community deeplinks
                     composable(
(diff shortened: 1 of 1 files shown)


Anna-log7:
So I've got this building/running with no issues, but it looks like the pipeline is failing. I'd love to fix the issue, but I don't love giving read access to my private repos. Is there a way I can get around that requirement to log in with my account?

dessalines:
I don't think so, not with woodpecker at least. Run `./gradlew formatKotlin` to pass the lint.

dessalines (approved):

PR #424, opened 2023-06-05
Aesthetics improvements

pipe01:
* Disabled autocorrection in the login form's instance field.
* Added some padding to the top of the login button.
* Set the status and navigation bar colors depending on the chosen theme.
  * Also set the navigation color depending on the bottom app bar visibility.

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/common/AppBars.kt
+++ b/app/src/main/java/com/jerboa/ui/components/common/AppBars.kt
@@ -1,6 +1,7 @@
 package com.jerboa.ui.components.common
 
 import android.annotation.SuppressLint
+import android.app.Activity
 import androidx.compose.animation.core.animateFloatAsState
 import androidx.compose.animation.core.tween
 import androidx.compose.foundation.ExperimentalFoundationApi
@@ -18,6 +19,7 @@ import androidx.compose.material.icons.filled.Person
 import androidx.compose.material.icons.outlined.*
 import androidx.compose.material3.*
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.DisposableEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
@@ -25,6 +27,7 @@ import androidx.compose.ui.draw.drawWithContent
 import androidx.compose.ui.geometry.Offset
 import androidx.compose.ui.geometry.Size
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.toArgb
 import androidx.compose.ui.graphics.vector.ImageVector
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.text.TextStyle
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/jerboa/ui/components/common/AppBars.kt
```diff
@@ -85,6 +88,17 @@ fun BottomAppBarAll(
     val totalUnreads = unreadCounts?.let { unreadCountTotal(it) }
 
     if (showBottomNav == true) {
+        val window = (LocalContext.current as Activity).window
+        val colorScheme = MaterialTheme.colorScheme
+
+        DisposableEffect(Unit) {
```
7heo, line 94: I'm guessing the check failed because of a missing space between `DisposableEffect` and the following parenthesis.

app/src/main/java/com/jerboa/ui/components/login/Login.kt
```diff
@@ -124,6 +126,7 @@ fun LoginForm(
                 trailingIcon = {
                     TrailingIcon(expanded = expanded)
                 },
+                keyboardOptions = KeyboardOptions(autoCorrect = false, keyboardType = KeyboardType.Uri)
```
7heo, line 129: Possibly here too.

app/src/main/java/com/jerboa/ui/components/login/Login.kt
```diff
@@ -158,6 +161,7 @@ fun LoginForm(
         Button(
             enabled = isValid && !loading,
             onClick = { onClickLogin(form, instance) },
+            modifier = Modifier.padding(top = 10.dp)
```
7heo, line 164: And here.

7heo:
Could you please re-trigger the build? We can now see the logs.

pipe01:
Not sure how to do that, sorry.

7heo:
Just force push after having made change to the commit (with `git commit --amend --date="$(date)"` for example, or maybe `git commit --amend --reset-author` but I'm not 100% certain)

7heo:
Well, now you get a proper error output to work with :P Good luck!

pipe01:
I don't see any errors though, only warnings... It seems like gradle crashed?

7heo:
From the logs:
> * What went wrong:
> Gradle build daemon disappeared unexpectedly (it may have been killed or may have crashed)

Yeah you're right. I suppose the build is to heavy/long for the current infra and it was either OOM killed or terminated because of execution time.

7heo:
Also, the infra used by the CI is an OVH VPS, so maybe go easy on it? :D By that I mean, I don't think re-triggering the build will do any good, as it will just consume resources for 20+ minutes to likely fail the same way. But maybe I'm wrong and it will go through?

pipe01:
Oops lol, I'm just not sure what the issue is. It builds just fine on my machine.

7heo:
I deleted my previous comment because you re-triggering the build actually disclosed what the problem is: `No space left on device`. I guess the VPS build drive is full. Not your fault, it was bound to happen one way or another, I think.

7heo:
@twizmwazin Hey, do you think you could re-trigger this job on the CI please? :)

pipe01:
There we go, now it's working

dessalines:
Sorry about the CI issues all, our runners have been pretty overloaded.

dessalines (approved):
Looks good to me, thanks!

7heo:
Yaay! :)

Issue #371, opened 2023-06-02
Flicking between tabs (at least what I think) is causing the app to think the entire screen is the same as a post, and therefore flashbangs/creates a rather white light upon tab switching only.


ColdyCold:
https://youtube.com/shorts/mNxzDxvASrI?feature=share

(I don't know why but it's not letting me posta video here, but it's essentially doing this.)

dessalines:
Not sure what could be causing that. I don't do any special / custom transitions between screens.

twizmwazin:
Fixed in #428

PR #428, opened 2023-06-05
Fix the flashing between transitions (#371)

AntmanLFEz:
During transitions the window background (default white) is displaying. When using a dark theme this looks like a flash.

Making the window background transparent uses the themes color instead.

### DIFF
--- a/app/src/main/java/com/jerboa/MainActivity.kt
+++ b/app/src/main/java/com/jerboa/MainActivity.kt
@@ -2,6 +2,8 @@ package com.jerboa
 
 import android.app.Application
 import android.content.Intent
+import android.graphics.Color
+import android.graphics.drawable.ColorDrawable
 import android.net.Uri
 import android.os.Build
 import android.os.Bundle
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/jerboa/MainActivity.kt
```diff
@@ -98,6 +100,8 @@ class MainActivity : ComponentActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
+        window.setBackgroundDrawable(ColorDrawable(Color.TRANSPARENT))
```
dessalines, line 103: Where are you getting this from? Does this affect every single jetpack compose app that doesn't have this hack?

AntmanLFEz: As far as I could tell this is the only workaround. Transitions in jetpack compose are still considered experimental.

dessalines: Okay, thanks.

dessalines (approved):

Issue #373, opened 2023-06-02
Allow setting a different font size for comments


Elbullazul:
The default 16px font size is OK for most of the UI, but comments have almost double the font size. Would it be possible to change font size only for the comments?

Elbullazul:
Here's an example:
![Screenshot_20230602-161146](https://github.com/dessalines/jerboa/assets/15310985/7c8efc62-6c9e-4659-a8b5-a9692a425a15)

The community font size is pretty small, and even the post title is smaller than the comment font size

Nisker:
I have the same problem. 
I think this is a bug we are experiencing.
From what i can tell, it's due to the comments not "reacting" to the android font setting, while the rest of the UI does.
I use the smallest font size on Android, which makes this a lot worse.
@Elbullazul what font size you use in your Android settings?

Tavius02:
I have the same issue with app, and I use the second smallest font size.

dessalines:
This has to do with the markdown renderer jerboa is using, not really working great with jetpack-compose sizing, but also being the only one that handles images properly. 

I *had* a PR out to replace it with a better one, but images don't work well on that one. 

Rather than having a  custom font size just to support that older library, I'd rather just get the new one working properly.

jplsek:
> Hrm. I noticed the comments are a _bit_ small right now, but I suppose with the new markwon, its using `bodyLarge`. Perhaps all we need to change is that. We also need to verify that its actually using that: there's a chance its getting ignored.
> 
> I'd definitely prefer just a single font size, so that all titles, bodies, etc can scale accordingly. Otherwise it could get complicated very quickly.

I might be missing some context, or maybe there is a bug somewhere, but for me when testing on `main`, the comments look larger to me, not smaller.

Before (comparing against the home activity):

| | |
|---|---|
| ![Screenshot_20230610_151546](https://github.com/dessalines/jerboa/assets/2388657/d3958b73-a959-4c7e-98fa-0f5b420cc015) | ![Screenshot_20230610_151616](https://github.com/dessalines/jerboa/assets/2388657/adf51ece-1e8c-4b6b-b400-781e35209a4d) | 

After [(setting to Medium)](https://github.com/dessalines/jerboa/commit/23c0fee1a69c858e7c3582bc8a415e140c96ff75), such that the sizes looks consistent:

| | |
|---|---|
[...]

PR #432, opened 2023-06-05
Markdown renderer replacement

AntmanLFEz:
Images show at approx 45% size in the markdown renderer.

This allows us to remove the markdown scaling applied to the editor which forced the text content to also be increased which made it inconsistent with the rest of the app.

Removed the compose-markdown package and created our own wrapper around the markwon libraries with a custom resolver for images.

#373

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/common/InputFields.kt
+++ b/app/src/main/java/com/jerboa/ui/components/common/InputFields.kt
@@ -676,10 +676,15 @@ fun PreviewLines(
 fun MyMarkdownText(
     markdown: String,
     color: Color = MaterialTheme.colorScheme.onSurface,
-    onClick: (() -> Unit)? = null,
+    onClick: () -> Unit,
     onLongClick: (() -> Unit)? = null,
 ) {
-    MarkdownHelper.CreateMarkdownView(markdown = markdown, color = color, onClick = onClick, onLongClick = onLongClick)
+    MarkdownHelper.CreateMarkdownView(
+        markdown = markdown,
+        color = color,
+        onClick = onClick,
+        onLongClick = onLongClick,
+    )
 }
 
 fun String.insert(index: Int, string: String): String {
--- a/app/src/main/java/com/jerboa/ui/components/common/InputFields.kt
+++ b/app/src/main/java/com/jerboa/ui/components/common/InputFields.kt
@@ -11,7 +11,6 @@ import androidx.compose.foundation.horizontalScroll
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
-import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.rememberScrollState
@@ -71,11 +70,9 @@ import com.jerboa.api.uploadPictrsImage
 import com.jerboa.appendMarkdownImage
 import com.jerboa.db.Account
 import com.jerboa.imageInputStreamFromUri
-import com.jerboa.ui.theme.MARKDOWN_FONT_MULTIPLIER
 import com.jerboa.ui.theme.MEDIUM_PADDING
 import com.jerboa.ui.theme.SMALL_PADDING
 import com.jerboa.ui.theme.muted
-import dev.jeziellago.compose.markdowntext.MarkdownText
 import kotlinx.coroutines.launch
 
 @Composable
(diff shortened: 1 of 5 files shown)


twizmwazin:
I tested this out and it seems to work quite well. Text size is correct, and images render correctly and don't cause weird resizing and snapping issues I was experiencing with the previous renderer.

twizmwazin:
Some follow up notes from testing:

- If the last character of a line is italicized, it will be cut off. For example, if a comment is deleted, the d will appear as an a.
- Images are unloaded if they go off screen
- Images sharing a line with text do not have sufficient padding with that text. While this is probably correct for markdown, there should probably be some effort on the app's side to make it look better, by inserting additional newlines before and after pictures if they aren't there already. 
- Sometimes images are very small and effectively unviewable. 

Here is an excellent thread for testing: https://beehaw.org/post/433016

AntmanLFEz:
Thanks for the feedback @twizmwazin 

I found the issue / fix for the very small pictures. I will take a look at the other items soon

AntmanLFEz:
Actually the best option will be to use image-coil as the backend. Testing it now with my other fix seems to work very well. No image unloading.

Will do more testing and try and apply the same preview etc we use for post images.

dessalines:
Did you look at this PR at all? https://github.com/dessalines/jerboa/pull/274

RichText material 3 seems to be the best future-proof option for markdown rendering... I just couldn't get images working properly with it. IMO it would be better to work on that directly to add better image support, since that would positively affect all jetpack-compose projects that need markdown.

AntmanLFEz:
@dessalines  I will take a look.

So far I have updated the pull request with the coil implementation. So far it works pretty well (with limited testing)

Apologies if i butchered the PR. Its my first stab at doing it "properly"

dessalines:
Its no problem, I should be able to get to this tomorrow.

app/src/main/java/com/jerboa/ui/components/common/InputFields.kt
```diff
@@ -679,17 +676,10 @@ fun PreviewLines(
 fun MyMarkdownText(
     markdown: String,
     color: Color = MaterialTheme.colorScheme.onSurface,
-    onClick: () -> Unit,
+    onClick: (() -> Unit)? = null,
```
dessalines, line 679: What's going on here? Make sure these are non-optional like before.

app/src/main/java/com/jerboa/ui/components/common/InputFields.kt
```diff
@@ -679,17 +676,10 @@ fun PreviewLines(
...
-    MarkdownText(
-        markdown = markdown,
-        modifier = Modifier.fillMaxSize(),
-        onClick = onClick,
-        color = color,
-        fontSize = MaterialTheme.typography.bodyLarge.fontSize.times(MARKDOWN_FONT_MULTIPLIER),
-//        style = MaterialTheme.typography.titleLarge,
-//        imageLoader =  LocalImageLoader.current
-    )
+    MarkdownHelper.CreateMarkdownView(markdown = markdown, color = color, onClick = onClick, onLongClick = onLongClick)
```
dessalines, line 682: Hopefully this is failing lint. Run `./gradlew formatKotlin`.

app/src/main/java/com/jerboa/ui/components/common/MarkdownHelper.kt
```diff
@@ -0,0 +1,116 @@
...
+                    else -> View.TEXT_ALIGNMENT_TEXT_START
+                }
+            }
+
+            fontResource?.let { font ->
+                typeface = ResourcesCompat.getFont(context, font)
+            }
+        }
+    }
+}
```
dessalines, line 116: Not the hugest fan of all this custom code, rather than just using a decent jetpack-compose markdown library. But if there are no real good ones, I suppose we don't have a choice.

dessalines (commented):
In my testing, I'd say this is already better than the current renderer, so great job! Just fix that one issue above, and we can merge.

app/src/main/java/com/jerboa/ui/components/common/InputFields.kt
AntmanLFEz: This was not being caught by `formatKotlin`.  Updated now

app/src/main/java/com/jerboa/ui/components/common/MarkdownHelper.kt
AntmanLFEz: I am not aware of one. The one you worked on previously does have a PR from two weeks ago with some possible fixes however I am not sure how well maintained that project is.

app/src/main/java/com/jerboa/ui/components/common/InputFields.kt
```diff
@@ -680,15 +677,13 @@ fun MyMarkdownText(
     markdown: String,
     color: Color = MaterialTheme.colorScheme.onSurface,
     onClick: () -> Unit,
+    onLongClick: (() -> Unit)? = null,
```
dessalines, line 680: Make sure this is ` -> Unit` with no default, just like the one above it.

dessalines (commented):
Also, there's one merge conflict.

app/src/main/java/com/jerboa/ui/components/common/InputFields.kt
AntmanLFEz: onLongClick was never available before. If  this parameter becomes non optional then i must update all `CreateMarkdownView()` calls with an additional unneeded empty object.

Both parameters should probably have a default of `(() -> Unit)? = null` as click events on the ultimately rendered TextView should be optional in any case without needing redundant `{}, {}` everywhere. An example could be a markdown rendered popup.

dessalines (approved):

PR #436, opened 2023-06-06
Expand area tappable by user to collapse a comment

twizmwazin:
This allows the user to tap in the action bar, as well as disables the animation in the header bar in a comment.


dessalines (approved):
Looks good, thanks!

PR #438, opened 2023-06-06
Add option to show contents of collapsed comments

twizmwazin:
Basic functionality is working, but still a few TODOs on this:

- ~No indication if a comment is collapsed since the whole body is shown. Boost uses a little +N icon in the top bar.~
- ~Comments folding isn't animated currently~
- ~Settings migrations needs to be tested~

PR #446, opened 2023-06-07
Collapsing/expanding comments no longer collapses the comment itself, only its children

a1studmuffin:
Collapsing/expanding comments no longer collapses the comment itself, only its children (the previous behaviour was disorienting and could result in accidentally collapsing a comment when trying to click on a link in the comment body). In addition, the ShowMoreChildren component is reused to indicate how many comments were collapsed. Screenshot:

collapsed_comment

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/comment/CommentNode.kt
+++ b/app/src/main/java/com/jerboa/ui/components/comment/CommentNode.kt
@@ -1,9 +1,6 @@
 package com.jerboa.ui.components.comment
 
 import android.widget.Toast
-import androidx.compose.animation.AnimatedVisibility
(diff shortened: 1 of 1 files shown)


a1studmuffin:
I realise this one probably conflicts with https://github.com/dessalines/jerboa/pull/438, I don't mind which way we go but thought I'd put forward what I feel is the "right" fix for my previous comment collapsing PR. This feels more natural to me.

If users would prefer to go with a setting to control it, I'm all for an alternative fix, in which case feel free to close out.

dessalines:
Closed in favor of the other one. As mentioned, we can do a poll or something later to see if users prefer the other way.

PR #447, opened 2023-06-07
started localization for the app and added first german local

kuroi-usagi:
I started the effort to localize the app. For whatever reason i don't understand the app wanted to have a DB version bump.
I tested it on emulator and a device.
There are a whole lot of duplicate words in the first step but i thought it would be easier to keep a naming scheme that is easy to understand then to remove duplicates and make it more complicated for everyone in the first place.
Getting rid of duplicates would be another step in this feature i think.


app/src/main/java/com/jerboa/db/AppDB.kt
```diff
@@ -302,8 +302,15 @@ val MIGRATION_9_10 = object : Migration(9, 10) {
     }
 }
 
+val MIGRATION_10_11 = object : Migration(10, 11) {
```
dessalines, line 305: What's going on here? You should probably merge from main.

dessalines (commented):
This is a great first step! I've internationalized a few other android apps, but was really daunted by this one. Laying the groundwork like you've done here is the difficult part :), thank you!

Run `./gradlew formatKotlin` to pass lint, then once I get a minute I'll also grep through the code to make sure nothing's missing.

Once this passes

app/src/main/java/com/jerboa/db/AppDB.kt
kuroi-usagi: I've had the issue that after an update the app crashed and told me i need a migration to work. Tried it that way. Maybe its not correct. Sorry i am new to this 😓 
Will try to revert this.

dessalines: Haha no problem. Just merge from main and you should be good.

dessalines:
Hrm... formatting still had issues. https://woodpecker.join-lemmy.org/dessalines/jerboa/pipeline/128/3

kuroi-usagi:
Wasn't the formatting, it was a merge error, i fixed it right now.

dessalines:
> For now a few strings in the Toasts are missing because of the missing context in those cases but i think its a start.

Could you add context to those? 

It'd be good to get all the strings now, because once people start adding languages, it'll be tougher later on.

If not I can probably help out in a day or two.

kuroi-usagi:
And ran into another merge conflict, you guys are working to fast 😂

kuroi-usagi:
> > For now a few strings in the Toasts are missing because of the missing context in those cases but i think its a start.
> 
> Could you add context to those?
> 
> It'd be good to get all the strings now, because once people start adding languages, it'll be tougher later on.
> 
> If not I can probably help out in a day or two.

Honestly i can look into it tomorrow. Its 11:30pm here and i had a few Whiskeys already. I am well beyond the Ballmer Peak 😅

kuroi-usagi:
Give me a minute, i've found the context solution. Maybe the Ballmer Peak is not a lie...

kuroi-usagi:
I have added the three toasts i have found and it should work now. Please check it if you can.

dessalines:
Mmmk I'll check now.

kuroi-usagi:
Searched for Text( and text = to be sure to find everything. I think i have it covered now.

dessalines (changes requested):
Do the search again, I found a lot that were still missing:

CommentNode L419, CommentNode L620 , etc.

kuroi-usagi:
I did again a search for `text =`, then for `(text =` which resulted in other entries, then i opted to search via regex for `/.+([A-z0-9])/`. I hope i have found everything.

app/src/main/res/values-de/strings.xml
```diff
@@ -0,0 +1,198 @@
+<resources>
+    <string name="app_name">Jerboa f[r Lemmy</string>
```
MV-GH, line 2: I think this should be "für" instead of "f[r"

twizmwazin (approved):

app/src/main/res/values-de/strings.xml
kuroi-usagi: You are totally right. Fixed it just now.

kuroi-usagi:
I don't see why the build is broken. Local build runs fine. Is it possible to restart the build?

MV-GH:
Seems the  `Gradle build daemon disappeared unexpectedly (it may have been killed or may have crashed)` is due to (lack of) memory issues. A potential solution I got from [here](https://stackoverflow.com/questions/37171043/gradle-build-daemon-disappeared-unexpectedly-it-may-have-been-killed-or-may-hav) is to add this to the woodpecker CI config
```
environment:
    -  GRADLE_OPTS=-Dorg.gradle.jvmargs="-XX:MaxMetaspaceSize=1g"
```
Not sure how effective this would be though.

dessalines (approved):
Okay I think that's all of em, tested and it works great!

Thank you so much for doing this!

I'll handle the merge conflicts.

PR #451, opened 2023-06-07
Enable autofill on login form

### LINKED ISSUE #439: Login form is not treated as a login form by autofill services
Expected behaviour is that the login form is treated as such by autofill services like password managers.

pipe01:
Fixes #439

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/login/Login.kt
+++ b/app/src/main/java/com/jerboa/ui/components/login/Login.kt
@@ -2,6 +2,7 @@
 
 package com.jerboa.ui.components.login
 
+import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.fillMaxSize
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/jerboa/ui/components/login/Login.kt
```diff
@@ -150,11 +161,23 @@ fun LoginForm(
             }
         }
         MyTextField(
+            modifier = Modifier
+                .background(if (wasAutofilled) Color(0xFFFFFDE7) else Color.Transparent)
```
dessalines, line 165: No hardcoding colors like that. Use material 3 colors.

pipe01: I couldn't find any color for this so I'm not sure what the proper way to this would be. Maybe adding an extension to `ColorScheme`?

dessalines: I'm sure one of these would work, no need to add extensions.

https://m3.material.io/styles/color/the-color-system/color-roles

Try different ones out and and see what looks good.

`MaterialTheme.colorScheme...`

pipe01:
Should be good now

dessalines (approved):

PR #453, opened 2023-06-07
Comment action bar improvements

twizmwazin:
This PR currently is built on top of #432. It adds an option to show action bars by default for comments. Currently this is built into database migration 11, but if there is a release before this is merged I will make it a different migration version. It also adds the ability to long-press on a comment to toggle the action bar state.

I can rebase this to work without #432, but the current markdown renderer does not expose an `onLongClick` option, so the long press would only work in the top and bottom bars.


dessalines:
> It also adds the ability to long-press on a comment to toggle the action bar state.

I'll make an issue to add a user guide to the readme, and link it from somewhere within the app.

app/src/main/java/com/jerboa/ui/components/person/PersonProfileActivity.kt
```diff
@@ -518,6 +519,7 @@ fun UserTabs(
                             account = account,
                             moderators = listOf(),
                             isCollapsedByParent = false,
+                            showActionBarByDefault = appSettingsViewModel.appSettings.value?.showCommentActionBarByDefault ?: true,
```
dessalines, line 522: 👍

dessalines (approved):
Works great, thank you!

TrollBlox:
Could we add an option to long press to collapse and tap to expand the action bar?

dessalines:
No because tap collapses / uncollapses child comments.

PR #469, opened 2023-06-09
Fixed up ThemeMode enum to use strings in UI

a1studmuffin:
Fixed up ThemeMode enum to use strings in "Look and Feel" settings UI (adds space to "System Black" theme mode)


twizmwazin:
I think this should wait on #447 so these strings can be included.

dessalines (approved):

app/src/main/java/com/jerboa/Utils.kt
```diff
@@ -724,12 +724,12 @@ fun Context.findActivity(): Activity? = when (this) {
...
 
-enum class ThemeMode {
-    System,
-    SystemBlack,
-    Light,
-    Dark,
-    Black,
+enum class ThemeMode(val mode: String) {
+    System("System"),
+    SystemBlack("System Black"),
```
dessalines, line 729: Oops yeah, add these strings now to `strings.xml`

dessalines (changes requested):
Oops yeah, add these strings now to strings.xml

twizmwazin (approved):

dessalines (approved):

PR #472, opened 2023-06-09
Add voting arrows and score to posts in "List" view mode

a1studmuffin:
Add voting arrows and score to posts in "List" view mode. Looks like this, basically what Boost and BaconReader does:

voting_arrows_list


a1studmuffin:
image

twizmwazin (changes requested):
Looks good, can we have a settings toggle for this?

twizmwazin (approved):
Tested and seems to work as intended. May want to consider waiting for localization to merge.

app/src/main/java/com/jerboa/db/AppDB.kt
```diff
@@ -325,8 +330,17 @@ val MIGRATION_10_11 = object : Migration(10, 11) {
     }
 }
 
+val MIGRATION_11_12 = object : Migration(11, 12) {
```
twizmwazin, line 333: Actually, can you make this part of the 10_11 migration? Let's keep all the settings migrations there until 0.0.33 is released.

twizmwazin (changes requested):
Sorry about that, one minor tweak

app/src/main/java/com/jerboa/db/AppDB.kt
a1studmuffin: Good point! This confused me as I must have had V11 in my local build, then made changes, and got a schema exception on startup. But I think I just needed to clear my app data... should be fixed now, I've moved it into V11 with the other settings.

twizmwazin (approved):

app/src/main/java/com/jerboa/db/AppDB.kt
```diff
@@ -322,6 +327,9 @@ val MIGRATION_10_11 = object : Migration(10, 11) {
         database.execSQL(
             "ALTER TABLE AppSettings add column show_comment_action_bar_by_default INTEGER NOT NULL default 1",
         )
+        database.execSQL(
+            "ALTER TABLE AppSettings add column show_voting_arrows_in_list_view INTEGER NOT NULL default 1",
```
dessalines, line 331: @twizmwazin I'm ambivalent, but do you think these really needs a setting? I'm leaning towards no, that it should just always show it.

If it does, then it needs a separate migration. The only reason to mess with old migrations, is if something in them was broken AND we haven't done a release.

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
```diff
@@ -857,13 +865,74 @@ fun PostListing(
...
+            account = account,
+        )
+        Text(
+            text = instantScores.score.toString(),
+            style = MaterialTheme.typography.bodyMedium,
+            color = scoreColor(myVote = instantScores.myVote),
+        )
+        // invisible Text below aligns width of PostVotingTiles
+        Text(
+            text = "00000",
```
dessalines, line 924: Probably a better way to do this, but I'm fine with it if it works.

app/src/main/java/com/jerboa/db/AppDB.kt
twizmwazin: I personally want a setting to disable it, if it wasn't added in this PR I would have done it myself soon. Boost also has a setting to control this.

I was under the impression for migrations that we would use a single migration between each released version. Otherwise, every single checkbox we add to the settings is going to result in a very long list of migrations very quickly. We haven't discussed this explicitly though, perhaps we should make a contributing guide that clearly establishes these guidelines.

dessalines: mmmk we can leave.

I don't mind tons of migrations, we're going to end up with a lot no matter what. The general rule is that a migration should be attached to a feature.

Its also **much** safer to have new migrations, than to edit previously committed ones. I've run into this in the past, where people carelessly edited existing deployed migrations, and broke production. 

If we need to later, we can organize them into their own file(s) if it gets out of hand. Diesel for rust does this automatically, and I'm surprised Room doesn't encourage you to do this already.

dessalines: @a1studmuffin So in that case break this out into its own migration if you would.

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
a1studmuffin: Agree it feels hacky, but the good thing about this solution is it'll scale properly for the user's font size settings, so won't take up more horizontal room than needed.

a1studmuffin:
All done, let me know if you need anything else. 👍

twizmwazin (approved):

app/src/main/java/com/jerboa/db/AppDB.kt
```diff
@@ -361,6 +375,7 @@ abstract class AppDB : RoomDatabase() {
                         MIGRATION_8_9,
                         MIGRATION_9_10,
                         MIGRATION_10_11,
+                        MIGRATION_11_12,
```
dessalines, line 378: 👍

dessalines (approved):
Looks good, thx!

PR #474, opened 2023-06-09
Add custom tab support

### LINKED ISSUE #396: Support opening links in Custom Tab / Better link handling
This would allow users to see links in posts without leaving the app, while still utilising their favourite browser.

twizmwazin:
Fixes #396 

Consider blocking on #447 for strings

### DIFF
--- a/app/src/main/java/com/jerboa/Utils.kt
+++ b/app/src/main/java/com/jerboa/Utils.kt
@@ -16,6 +16,7 @@ import android.provider.MediaStore
 import android.util.Log
 import android.util.Patterns
 import android.widget.Toast
+import androidx.browser.customtabs.CustomTabsIntent
 import androidx.compose.foundation.lazy.LazyListState
 import androidx.compose.material3.DrawerState
 import androidx.compose.material3.ExperimentalMaterial3Api
@@ -367,9 +368,15 @@ fun LazyListState.isScrolledToEnd(): Boolean {
     return out
 }
 
-fun openLink(url: String, ctx: Context) {
-    val intent = Intent(Intent.ACTION_VIEW, Uri.parse(url))
-    ctx.startActivity(intent)
+fun openLink(url: String, ctx: Context, useCustomTab: Boolean) {
+    if (useCustomTab) {
+        val intent = CustomTabsIntent.Builder()
+            .build()
+        intent.launchUrl(ctx, Uri.parse(url))
+    } else {
+        val intent = Intent(Intent.ACTION_VIEW, Uri.parse(url))
+        ctx.startActivity(intent)
+    }
 }
 
 fun prettyTimeShortener(timeString: String): String {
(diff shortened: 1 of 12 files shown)


app/src/main/java/com/jerboa/db/AppDB.kt
```diff
@@ -322,6 +327,9 @@ val MIGRATION_10_11 = object : Migration(10, 11) {
         database.execSQL(
             "ALTER TABLE AppSettings add column show_comment_action_bar_by_default INTEGER NOT NULL default 1",
         )
+        database.execSQL(
+            "ALTER TABLE AppSettings add column use_custom_tabs INTEGER NOT NULL default 1",
```
dessalines, line 331: I know we haven't released, but its a pain to break dev setups. Add this as a separate migration.

app/src/main/java/com/jerboa/ui/components/settings/lookandfeel/LookAndFeelActivity.kt
```diff
@@ -165,6 +167,13 @@ fun LookAndFeelActivity(
                     },
                     onCheckedChange = { updateAppSettings() },
                 )
+                SettingsCheckbox(
+                    state = useCustomTabsState,
+                    title = {
+                        Text(text = "Use custom tabs")
```
dessalines, line 173: Merge from main and add this as a string.

app/src/main/java/com/jerboa/Utils.kt
```diff
@@ -366,9 +367,15 @@ fun LazyListState.isScrolledToEnd(): Boolean {
...
 }
 
-fun openLink(url: String, ctx: Context) {
-    val intent = Intent(Intent.ACTION_VIEW, Uri.parse(url))
-    ctx.startActivity(intent)
+fun openLink(url: String, ctx: Context, useCustomTab: Boolean) {
+    if (useCustomTab) {
+        val intent = CustomTabsIntent.Builder()
+            .build()
+        intent.launchUrl(ctx, Uri.parse(url))
```
dessalines, line 374: These both did the exact same thing in my testing, just loaded up my external browser.

Seems more complicated than this. Check here:

https://stackoverflow.com/a/70514350/1655478

twizmwazin: I got some weird results in my testing, but it does appear to work in both the emulator and on my Pixel 5, using chrome and firefox as providers, respectively. The first couple of times I tried using it, it would open in the default browser, which is a documented fallback mode if no installed browser provides custom tabs. But it is behaving correctly now. I wish I had a better understanding of this behavior.

dessalines: Mmmk. Maybe its just a result of my default browser not supporting custom tabs.

dessalines (approved):
LGTM. Feel free to merge whenever you get the conflicts resolved.

PR #475, opened 2023-06-09
Fix accompanist deprecations

twizmwazin:
~I have not fully tested this yet~

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -84,7 +84,6 @@ dependencies {
     // Accompanist
     implementation 'com.google.accompanist:accompanist-pager:0.30.1'
     implementation 'com.google.accompanist:accompanist-pager-indicators:0.30.1'
-    implementation 'com.google.accompanist:accompanist-swiperefresh:0.30.1'
     implementation 'com.google.accompanist:accompanist-flowlayout:0.30.1'
 
     // LiveData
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/jerboa/ui/components/inbox/InboxActivity.kt
```diff
@@ -143,7 +144,7 @@ enum class InboxTab {
     Messages,
 }
 
-@OptIn(ExperimentalPagerApi::class)
+@OptIn(ExperimentalFoundationApi::class, ExperimentalMaterialApi::class)
```
dessalines, line 147: Double :) . Android studio does this to me all the time too.

dessalines (commented):
I'll review once you're satisfied w/ it. I'll mark as draft until then.

app/src/main/java/com/jerboa/ui/components/inbox/InboxActivity.kt
twizmwazin: These are separate, removing either causes a build failure. Is there an opt-in that is a superset of these two?

dessalines: Ohhh my bad, I thought they were the same string at first glance.

app/build.gradle
```diff
@@ -75,7 +75,6 @@ dependencies {
     // Accompanist
     implementation 'com.google.accompanist:accompanist-pager:0.30.1'
     implementation 'com.google.accompanist:accompanist-pager-indicators:0.30.1'
-    implementation 'com.google.accompanist:accompanist-swiperefresh:0.30.1'
```
dessalines, line 78: yaaay no more warnings from gradle builds!

dessalines (commented):
One thing I noticed that's missing, was those handy pull refresh loading indicators. This issue shows how to add them: https://stackoverflow.com/a/73953806/1655478

My `autogen_types` branch reworks the viewmodels and API to show API loading states easier, but that's a bit away. For now just use the old isRefreshing logic (its very hacky and imperfect, but it works).

twizmwazin:
Alright indicators are back!

dessalines (approved):
LGTM. Glad we can finally get rid of those warnings.

Issue #411, opened 2023-06-04
Allow paging instead of infinite scroll in feeds


oscarnylander:
Infinite scroll can unintentionally promote unhealthy consumption patterns. Paging can help create natural 'exit points', which in turn improves user agency over their consumption patterns.

dessalines:
I agree with this fully. If anyone wants to work on that, I'd be glad to merge it.

Die4Ever:
This would also be a good workaround for #1003

PR #483, opened 2023-06-09
Reduce stuttering in post listings

### LINKED ISSUE #445: Scrolling seems to sutter/lag on latest version (0.0.32)
This suttering/lag/jerkyness shows itself on both posts and comments, this is running on a Samsung a32

I checked logcat but unfortunately I am seeing no errors about jerboa, I do see an related scroll warning come up which I will link below.

image

I am also willing to put a $50 usd (monero) bug bounty on this for whoever can fix this issue.

pipe01:
I've managed to alleviate the constant stuttering every time a new post listing enters the screen by grouping a few posts into groups. This way the LazyColumn only has to create and destroy components much less often, reducing the feeling of stuttering.

This of course means that, whenever a new group must be created or removed, the stutter is more noticeable. However, this stutter doesn't seem to scale linearly with the amount of items per chunk and it's only slightly more noticeable than the stutter caused by every single item, so I think the trade is worth it.
If we decide that this is a good way to do it, it should be possible to make the same improvements to the comment nodes.

The current chunk size is set to 10 as a completely arbitrary number, we could play around with it to see what works best or even set it to the amount of items returned in a single post fetch call.

When/if #411 is implemented we could probably get away with grouping an entire page into a single chunk, making the experience completely stutter free.

I have also disabled the home top bar coming in and out as you scroll because it was a huge cause of lag, however this was only for testing purposes and I imagine we might wanna re-enable it later.

Fixes #445 somewhat.

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/home/HomeActivity.kt
+++ b/app/src/main/java/com/jerboa/ui/components/home/HomeActivity.kt
@@ -64,7 +64,7 @@ fun HomeActivity(
     val postListState = rememberLazyListState()
     val snackbarHostState = remember { SnackbarHostState() }
     val drawerState = rememberDrawerState(DrawerValue.Closed)
-    val scrollBehavior = TopAppBarDefaults.enterAlwaysScrollBehavior(rememberTopAppBarState())
+    val scrollBehavior = TopAppBarDefaults.pinnedScrollBehavior(rememberTopAppBarState())
     val ctx = LocalContext.current
     val account = getCurrentAccount(accountViewModel)
 
--- a/app/src/main/java/com/jerboa/ui/components/post/PostListings.kt
+++ b/app/src/main/java/com/jerboa/ui/components/post/PostListings.kt
@@ -1,10 +1,10 @@
 package com.jerboa.ui.components.post
 
+import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.LazyListState
-import androidx.compose.foundation.lazy.items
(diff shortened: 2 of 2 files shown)


dessalines:
We desperately need an android / jetpack-compose performance expert on this. I'm 99% sure we're missing something trivial in either the code or our release compilation, that would make these hacks unecessary.

dessalines:
Yep, its very possible that the `PostListing` component is the issue, although comment scrolling *feels* similarly stuttered. 

I don't see why it would be getting recomposed tho. When I did try to optimize things months ago, adding keys and such to lazylist, it didn't really improve anything, and I wasn't smart enough to be able to use the info from that android studio recomposition count view.

pipe01:
I think it does recompose since it needs to know what has changed, which is only possible by calling the composable function again. It's only after calling this method that the `LazyColumn` is able to reuse some other composable's result, at least that's how I see it. If this is right, the issue would be that the `PostListing` function call takes too long on the main thread, causing the stutter.

pipe01:
It does seem like when a deadline is missed, the main time hog is the `PostListing` composable:

image

dessalines:
What's that tool you're using? If I could figure out how to read that I could help a bit. Looks like the CircularIcon might be the issue?

Footsiefat:
@dessalines I may have found the tool he was using.

Go to https://medium.com/icerock/optimize-or-die-profiling-and-optimization-in-jetpack-compose-a165c8897b3f

and cntrl + f for `In Threads, you can click a method to search for it across the entire recorded interval and see how many times the thread was called and the average time spent for each call.`

pipe01:
I followed [this guide](https://medium.com/androiddevelopers/jetpack-compose-composition-tracing-9ec2b3aea535).

> Looks like the CircularIcon might be the issue?

It does look like that but the graph doesn't make much sense. For example, it's implying that `CircularIcon` is calling `PostFooterLine`, so I think something is not quite right.

dessalines:

PR #484, opened 2023-06-09
Remember listing type

### LINKED ISSUE #449: App doesn't remember if you were on Local, Subscribed, or All after restart
it always defaults to local.

taitsmith:
adds shared preferences to remember last selected listing type view, allowing the app to default to that view when reopening. uses shared preferences in place of account update so unregistered users get the same effect.

fixes #449


dessalines (commented):
I'm not interested in this, as it adds yet another data storage location, furthering complication with getting all 3 now synced up. 

There should only be two: the server, and the database. All that needs to happen to fix current issues, is to update and save your listingtype and sort settings in the database, after calling `GetSite`

twizmwazin:
I think a better option here is to instead have an option to select a default listing type in the settings. This would probably meet most users needs without adding much storage complexity.

dessalines:
That's in your profile settings. Someone did correctly add updating those to the accountSettingsViewModel (the function they added was `maybeUpdateAccountSettings`, but there's a few places where that's not syncing up. Once I get some time I'll be able to fix it.

taitsmith:
i'm happy to redo this, i just want to clarify preferred behavior / source of truth:

are we meant to solely rely on user settings (ie only update when a user clicks 'save settings') and ignore temporary changes made through filtering in the menu bar? if so it seems like that's already functioning as intended when i update my settings in the app, close && reopen and check on the web.

dessalines:
The source of truth in this case could be either the DB, or a fetch of `GetSite`, which has the user settings (`MyUserInfo`)

To allow faster startup times, we save the `default_listing_type` and `default_sort_type` in the `Account` table, so that the front page of posts can be fetched immediately, without fetching from getsite first, to get that setting. But what if the user changed that default setting in a different UI (So jerboa has no idea about it)? Then we need to set it after that initial `GetSite`, and save it in our local DB. That is the only thing that's currently missing.

> are we meant to solely rely on user settings (ie only update when a user clicks 'save settings') and ignore temporary changes made through filtering in the menu bar?

Filter changes are temporary. Only going into your account settings should make that permanent in both your local DB, and pushed to the lemmy backend.

twizmwazin:
I think we're not going to merge this since changing listing type is just temporary and there is a setting for default view.

PR #487, opened 2023-06-10
Add linkify plugin to markwon

twizmwazin:
This adds the linkify plugin to markwon. This will allow links to be automatically detected and made clickable, even if that link isn't formatted as a markdown link. I also added a second plugin to configure links rendered by markwon to be opened using our `openLink` function, intended for future use like custom tabs, special handling of detected fediverse links, etc.


dessalines (approved):
Nice, thanks! I noticed that URLs weren't getting linkified by markwon.

PR #489, opened 2023-06-10
Simple filter to display matching instances in the dropdown.

onowrouzi:
[login_instace_filter.webm](https://github.com/dessalines/jerboa/assets/15151458/b8fd3e4f-810d-4b9a-acc2-cd5e5ca5bff8)


twizmwazin (approved):

app/src/main/java/com/jerboa/ui/components/login/Login.kt
```diff
@@ -147,7 +150,7 @@ fun LoginForm(
                     expanded = false
                 },
             ) {
-                instanceOptions.forEach { selectionOption ->
+                instanceOptions.filter { instance.isEmpty() || it.contains(instance) }.forEach { selectionOption ->
```
dessalines, line 153: I noticed a lot of slowness and keyboard issues here, and it might be because the filtering is being done inside the ExposedDropdown.

https://alexzh.com/jetpack-compose-dropdownmenu/

Check the bottom of that page, it has an example of filtering. Very close to yours, but try it out.

dessalines (approved):
Works well, thank you!

dessalines:
Reverting this because it causes a lot of performance and weird issues with keyboards.

PR #492, opened 2023-06-10
Option to blur NSFW

TrollBlox:
Please give feedback on this, I know I did some stuff wrong because I have little experience with Kotlin and Jetpack Compose.

### DIFF
--- a/app/src/main/java/com/jerboa/Utils.kt
+++ b/app/src/main/java/com/jerboa/Utils.kt
@@ -382,8 +382,8 @@ fun prettyTimeShortener(timeString: String): String {
                 .replace(Regex("hours?"), "h")
                 .replace(Regex("days?"), "d")
                 .replace(Regex("weeks?"), "w")
-                .replace(Regex("months?"), "M")
-                .replace(Regex("years?"), "Y")
+                .replace(Regex("months?"), "mo")
+                .replace(Regex("years?"), "y")
                 .filter { !it.isWhitespace() }
         }
     } else {
--- a/app/src/main/java/com/jerboa/Utils.kt
+++ b/app/src/main/java/com/jerboa/Utils.kt
@@ -382,8 +382,8 @@ fun prettyTimeShortener(timeString: String): String {
                 .replace(Regex("hours?"), "h")
                 .replace(Regex("days?"), "d")
                 .replace(Regex("weeks?"), "w")
-                .replace(Regex("months?"), "mo")
-                .replace(Regex("years?"), "y")
+                .replace(Regex("months?"), "M")
+                .replace(Regex("years?"), "Y")
                 .filter { !it.isWhitespace() }
         }
     } else {
--- a/app/src/main/java/com/jerboa/ui/components/common/ImageViewerDialog.kt
+++ b/app/src/main/java/com/jerboa/ui/components/common/ImageViewerDialog.kt
@@ -69,7 +69,7 @@ fun ImageViewerDialog(url: String, onBackRequest: () -> Unit) {
     }
 
     val backColor = MaterialTheme.colorScheme.scrim
-    val backColorTranslucent = MaterialTheme.colorScheme.scrim.copy(alpha = 0.4f)
+    val backColorTranslucent = MaterialTheme.colorScheme.scrim.copy(alpha = 1f)
 
     var showTopBar by remember { mutableStateOf(true) }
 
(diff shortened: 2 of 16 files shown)


TrollBlox:
It works as intended though

app/src/main/java/com/jerboa/Utils.kt
```diff
@@ -382,8 +382,8 @@ fun prettyTimeShortener(timeString: String): String {
                 .replace(Regex("hours?"), "h")
                 .replace(Regex("days?"), "d")
                 .replace(Regex("weeks?"), "w")
-                .replace(Regex("months?"), "M")
```
twizmwazin, line 385: This change is unrelated

app/src/main/java/com/jerboa/ui/components/common/ImageViewerDialog.kt
```diff
@@ -69,7 +69,7 @@ fun ImageViewerDialog(url: String, onBackRequest: () -> Unit) {
     }
 
     val backColor = MaterialTheme.colorScheme.scrim
-    val backColorTranslucent = MaterialTheme.colorScheme.scrim.copy(alpha = 0.4f)
+    val backColorTranslucent = MaterialTheme.colorScheme.scrim.copy(alpha = 1f)
```
twizmwazin, line 72: Is this related to the main change?

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
```diff
@@ -319,7 +324,7 @@ fun PostTitleAndImageLink(
         .clickable { showImageDialog = true }
     PictrsUrlImage(
         url = url,
-        nsfw = nsfwCheck(postView),
+        blur = nsfwCheck(postView) && appSettings?.blurNSFW!!,
```
twizmwazin, line 327: I think it'd be a better idea to use `?: true` instead of `!!`

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
```diff
@@ -419,17 +429,17 @@ fun PostBody(
...
-fun PreviewStoryTitleAndMetadata() {
-    PostBody(
-        postView = samplePostView,
-        onPostLinkClick = {},
-        fullBody = false,
-        expandedImage = false,
-        account = null,
-    )
-}
+//@Preview
```
twizmwazin, line 432: Please restore this and the other previews

app/src/main/java/com/jerboa/ui/components/settings/SettingsActivity.kt
```diff
@@ -41,16 +41,6 @@ fun SettingsActivity(
         },
         content = { padding ->
             Column(modifier = Modifier.padding(padding)) {
-                SettingsMenuLink(
```
twizmwazin, line 44: This change seems unrelated

twizmwazin (changes requested):
So far in jerboa we've chose to pass specific flags into the composables rather than the whole AppSetting, so it'd be preferable to continue that. Also make sure to run `./gradlew formatKotlin`.

app/src/main/java/com/jerboa/Utils.kt
TrollBlox: Sorry, I have things changed for myself and they must have gotten in here somehow. I will change it on the next commit.

TrollBlox:
> So far in jerboa we've chose to pass specific flags into the composables rather than the whole AppSetting, so it'd be preferable to continue that.

Could you explain further?

twizmwazin:
You've added parameters to many functions for the AppSettings object. Instead of doing this, add a parameter for blur_nsfw instead. Avoid passing the appSettings object directly.

TrollBlox:
> You've added parameters to many functions for the AppSettings object. Instead of doing this, add a parameter for blur_nsfw instead. Avoid passing the appSettings object directly.

Oh I see what you mean now. I was already working to do that because I realized how dumb it was.

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
TrollBlox: all of the previews should be restored.

app/src/main/java/com/jerboa/ui/components/settings/SettingsActivity.kt
TrollBlox: i moved this back to its original location

TrollBlox:
oops

PR #500, opened 2023-06-10
Fixing markdown line height issue.

dessalines:
The markdown textview wasn't using the correct units, and it was also missing line height.


app/src/main/java/com/jerboa/Utils.kt
```diff
@@ -893,3 +894,7 @@ fun Modifier.onAutofill(vararg autofillType: AutofillType, onFill: (String) -> U
             }
         }
 }
+
+fun spToPx(sp: TextUnit, context: Context): Int {
```
twizmwazin, line 898: I think we should use a longer name than `spToPx` to make it more clear what the function does.

MV-GH: Or you could add a KDoc comment

for example
```kt
/**
 * Converts a scaleable pixel (sp) to an actual pixel (px)
 */
 fun spToPx(sp: TextUnit, context: Context): Int {
  ```

twizmwazin (approved):

PR #501, opened 2023-06-10
QoL: Allow debug version installed next to release version

MV-GH:
The purpose of this is to add a little QoL change for the devs. Now we can have both the debug version and the release version installed next to each other. Makes it also more clear which version that you are currently using.


dessalines (approved):
Nice, had no idea about this. Thx!

app/src/debug/res/values/strings.xml
```diff
@@ -0,0 +1,3 @@
+<resources>
+   <string name="app_name">Jerboa (DEBUG)</string>
+</resources>
```
twizmwazin, line 3: Silly thing, can you add a trailing newline here? Otherwise LGTM!

twizmwazin (approved):

PR #502, opened 2023-06-10
Hide downvote button on comments and posts when disabled

### LINKED ISSUE #390: Hide downvote button when downvoting is disabled
As of 0.0.31, the downvote button is clickable but gives a toast notification indicating that downvoting is disabled. Ideally the button should be removed when it is not enabled.

lsim:
Please have a look at this - it's my first 'proper' PR. I may have violated any number of code style rules and what not. 

I'm threading a `siteView` into the two composables where a check is needed in order to not render the downvote button.

Githubs differ got a bit confused due to the indentation change. There shouldn't be any surprises in those parts, though.

Fixes #390


lsim:
Having a look at the linter problems

lsim:
The only way I could figure out to test it was to hardcode the new conditionals to false and see that the UI was as expected. So if the LocalSite.downvotes_enabled gives the right value it should work.

Happy to do some more testing if you can teach me a better way to set that flag (I don't have my own lemmy test instance - yet).

app/src/main/java/com/jerboa/ui/components/comment/CommentNode.kt
```diff
@@ -492,6 +498,7 @@ fun CommentFooterLine(
     account: Account?,
 ) {
     var showMoreOptions by remember { mutableStateOf(false) }
+    var localSite = siteView.local_site
```
dessalines, line 501: Use val for constants.

app/src/main/java/com/jerboa/ui/components/comment/CommentNode.kt
```diff
@@ -204,6 +206,7 @@ fun LazyListScope.commentNodeItem(
     account: Account?,
     isCollapsedByParent: Boolean,
     showActionBar: (commentId: Int) -> Boolean,
+    siteView: SiteView,
```
dessalines, line 209: @twizmwazin I'm leaning towards only passing *just* the `enable_downvotes` setting, rather than the entire `SiteView`, to potentially prevent rerenders. What do you think?

app/src/main/java/com/jerboa/ui/components/community/CommunityActivity.kt
```diff
@@ -106,127 +108,134 @@ fun CommunityActivity(
...
-                onPostClick = { postView ->
-                    navController.navigate(route = "post/${postView.post.id}")
-                },
-                onPostLinkClick = { url ->
-                    openLink(url, ctx, appSettingsViewModel.appSettings.value?.useCustomTabs ?: true)
-                },
-                onSaveClick = { postView ->
-                    account?.also { acct ->
-                        communityViewModel.savePost(
+            siteViewModel.siteRes?.site_view?.also { siteView ->
```
dessalines, line 111: This is another reason to only pass the `enable_downvotes` setting... not having to wait for the entire siteview to get back. 

Then you can just do `siteViewModel.siteRes?.site_view?.local_site?.enable_downvotes ?: true`

app/src/main/java/com/jerboa/ui/components/comment/CommentNode.kt
twizmwazin: Just passing the `enable_downvotes` setting is consistent with the current approach. That said, this is the fifth setting being passed here, so I want to continue exploring alternatives.

lsim: Thanks. I'll get on it

lsim:
Hm - weird. CI is failing on something I already pushed a fix for (unused import). I think github has gotten confused. In `Files changed`, it still shows the unused import. And in `Commits` it shows the commit that removed the import :)

twizmwazin (approved):

PR #508, opened 2023-06-11
Generate default instance list on build

### LINKED ISSUE #6: Convert Gradle buildscripts to Kotlin script
Suggestion/discussion ticket for converting the app's buildscript files over to KTS.
This helps writing and updating the buildscript since it can be fully validated without invoking gradle (which tends to be faster), and you also get autocompletion which is pretty handy.

pipe01:
Added a script that will download a list of instances from https://the-federation.info/platform/73 using the GraphQL API, then write it to a file containing a constant.

Currently it'll get the top 10 instances with the most monthly users, but it can easily be changed to e.g. all instances with more than X monthly users instead. The script is called on every build but maybe it'd be better to only call it on release builds, not sure how to detect that though.

The script is written in Bash and thus will only work on Linux systems (it also requires `jq` and `curl`), so when building on other platforms a dummy instance list will be emitted.

Also migrated all Gradle build scripts to Kotlin, so fixes #6.


twizmwazin:
I really like this idea, but I would prefer it be done in a cross-platform manner rather than requiring shell scripts whose dependencies aren't explicitly documented.

kinther:
Jerboa shouldn't be dependent on running a script to pull a list from a website it doesn't control.

twizmwazin:
On second thought, this would violate reproducible builds by default. Maybe instead, this could be a Gradle task that is run that updates the file,  and then that file is committed to the repo. It would then be done before performing a release.

pipe01:
Sure, it can just make it a separate task instead. I could write it in a Kotlin script, but AFAIK it would mean that the `build.gradle` file must be converted to Kotlin, and the code would get pretty ugly.

twizmwazin:
A python script has the same problem as bash, unless gradle supports grabbing a python interpreter. There is another issue that suggests migrating the gradle build scripts to kotlin, I'd be in favor of that.

pipe01:
Yeah that's true, I'll write it in Kotlin then.

pipe01:
I've migrated the `build.gradle` file to Kotlin and also translated the fetch script into Kotlin. The script is in a separate file which makes it neater, but apparently you can't import libraries from other `.kts` files so I had to use the built-in APIs. This was a bit of a pain to figure out, but the result doesn't look too bad.

To run the task, just do `./gradlew fetchInstances`

app/build.gradle.kts
```diff
@@ -0,0 +1,128 @@
...
+    implementation("androidx.compose.ui:ui-tooling-preview:1.4.3")
+    implementation("androidx.activity:activity-compose:1.7.2")
+    testImplementation("junit:junit:4.13.2")
+    androidTestImplementation("androidx.test.ext:junit:1.1.5")
+    androidTestImplementation("androidx.test.espresso:espresso-core:3.5.1")
+    androidTestImplementation("androidx.compose.ui:ui-test-junit4:1.4.3")
+    debugImplementation("androidx.compose.ui:ui-tooling:1.4.3")
+    debugImplementation("androidx.compose.ui:ui-test-manifest:1.4.3")
+    implementation("net.engawapg.lib:zoomable:1.4.3")
+    implementation("androidx.browser:browser:1.5.0")
```
dessalines, line 127: Thanks for converting this file!

app/get_instances.gradle.kts
```diff
@@ -0,0 +1,103 @@
...
+import java.net.HttpURLConnection
+
+// We can't import libraries here for some reason, so we must use what is provided
+// by gradle, which isn't much. The groovy JSON library is meant for use by groovy code,
+// so we need some creativity to use it in Kotlin.
+import org.apache.groovy.json.internal.LazyMap
+import groovy.json.JsonOutput
+import groovy.json.JsonSlurper
+
+val endpointUrl = "https://the-federation.info/v1/graphql"
```
dessalines, line 12: The-federation.info isn't that well maintained, and contains a lot of old and generally incorrect info, as evidenced by the errors in the list it generated below. 

Two other options:
- Use https://lemmy.fediverse.observer/ , which has an API. Sort by Active users in the last month, not total users.
- Use the https://github.com/LemmyNet/lemmy-stats-crawler/ , which generates a usable json file.

app/get_instances.gradle.kts
```diff
@@ -0,0 +1,103 @@
+import java.io.OutputStreamWriter
```
dessalines, line 1: For generating a static list like this, you can use whatever you like, rust, kotlin, bash, node, anything. All that matters is that it replaces those lines in the default_instances, so its *results* are compiled.

I'll only run it before releases anyway, and double check its results. Just make sure you give instructions on how to run it in the README.

app/src/main/java/com/jerboa/DefaultInstances.kt
```diff
@@ -0,0 +1,14 @@
+package com.jerboa
+
+val DEFAULT_LEMMY_INSTANCES = arrayOf(
+    "dev.lemmy.ml", // 588 monthly users
```
dessalines, line 4: See above.

app/get_instances.gradle.kts
pipe01: I think the first option would be the most convenient, it would be pretty easy to modify the code to use that API instead.

app/get_instances.gradle.kts
pipe01: Since it's already in Kotlin I might as well leave as is, it's nice that it's integrated into gradle. I did notice that the AndroidManifest needs the references to the domains too, should I update that too?

dessalines: I wouldn't worry about it for this script. I'm going to remove all deeplinking at some point as its extremely complicated for federated apps.

pipe01: Ah fair enough

pipe01:
Where should I add the info on how to fetch the instances to the README? Maybe we should add a CONTRIBUTING file.

dessalines:
You're right, this would make more sense in a `CONTRIBUTING.md`. 

Could you create that, and add just that info? I can rework it and add a few things about how to debug and such once that's done.

ZJouba:
> You're right, this would make more sense in a `CONTRIBUTING.md`.
> 
> Could you create that, and add just that info? I can rework it and add a few things about how to debug and such once that's done.

Don't forget to add this info now that `CONTRIBUTING.md` has been merged

twizmwazin:
Obsoleted by #884

PR #519, opened 2023-06-11
Add a user agent

camporter:
Adds a user agent for debugging.


twizmwazin (approved):
Looks good to me, though I think there would be a portion of users who might prefer to disable this for privacy reasons. We might want to add that option in the future.

PR #534, opened 2023-06-11
Fix big font size cutting off settings options

calincara:
Capture1

Font slider still needs to be fixed, but I'm newbie when it comes to Jetpack compose; will keep trying to find a solution for that as well.


twizmwazin (approved):
Make sure to run `./gradlew formatKotlin` to fix the lint issues.

PR #549, opened 2023-06-11
Added default community icon where appropriate

a1studmuffin:
Added a default community icon wherever we show lists of communities. Prevents user confusion, as some users thought a community without an icon was a sub-category since the community name was on the left. Reused the same icon for communities as the "Go to <community>" submenu option.

Examples marked with red arrow:

image


a1studmuffin:
Discussion thread here: https://lemmy.ml/post/1202394

twizmwazin (approved):

PR #505, opened 2023-06-10
Expanded default lemmy instances list to active users/mo > 100

shombando:
Added more Lemmy instances to the DEFAULT_LEMMY_INSTANCES list and alphabetized the list so future additions can be easier. 

Added all lemmy instances which currently show greater than 100 users/month according to: https://the-federation.info/platform/73

The following new instances were added:
lemmy.ca
lemmy.one
lemmy.world
midwest.social
sh.itjust.works

PR #552, opened 2023-06-12
Update AndroidManifest.xml to match instances added in #505

shombando:
Alphabetized the list of instances and added the following new instances to intent list, to match PR #505: 
lemmy.ca
lemmy.one
lemmy.world
midwest.social
sh.itjust.works

I can investigate creating automation around updating DEFAULT_LEMMY_INSTANCES and AndroidManifest.xml based on defined criteria and information from https://the-federation.info/platform/73, if the maintainers would be interested in adopting that approach.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -42,54 +42,90 @@
             </intent-filter>
 
             <!-- The main instances-->
-            <intent-filter>
-                <action android:name="android.intent.action.VIEW" />
-                <category android:name="android.intent.category.DEFAULT" />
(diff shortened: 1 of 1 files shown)


twizmwazin (changes requested):
It looks like you've changed the indentation from spaces to a mix of tabs and spaces. Please change that back and then we can merge this.

shombando:
@twizmwazin sorry about that, editor config issue and I didn't catch it in the diff. I've reverted to spaces to be consistent with the previous state. Thanks for the kind feedback.

twizmwazin (approved):

PR #560, opened 2023-06-12
Bash isn't portable, use POSIX sh instead

7heo:
Also, hardcoding bash to /bin/bash is even less portable.

In addition, I commented an unused variable (name not present anywhere in the code base, not exported).


7heo:
> Why comment the unused variable? If truly unused removal should be okay.

I am not the original author of this script, it's not for me to decide. Maybe they will remember what they wanted to do with it and finish it. If I remove it, they might forget entirely.

7heo:
I mean, anyway, @dessalines wrote that line according to the git blame, so he will say something if it's meant to stay.

PR #563, opened 2023-06-12
Localization of user tabs

kuroi-usagi:
I've seen that there were a few locales missing for the UserTabs and the SortTypes in a few places, so i added some Util Functions to help with that and added all translations as good as possible.


app/src/main/res/values-nl/strings.xml
```diff
@@ -208,4 +208,18 @@
...
+    <string name="person_profile_activity_posts">Posts</string>
+    <string name="person_profile_activity_comments">Comments</string>
+    <string name="sorttype_active">Actief</string>
+    <string name="sorttype_hot">Populair</string>
+    <string name="sorttype_new">Nieuw</string>
+    <string name="sorttype_old">Oud</string>
+    <string name="sorttype_topday">Top van de dag</string>
+    <string name="sorttype_topweek">Top van de week</string>
+    <string name="sorttype_topmonth">Top van de maand</string>
+    <string name="sorttype_topyear">Top van het jaar</string>
```
MV-GH, line 221: In the english translation you keep them short, for parity maybe its better too?
```xml
<string name="sorttype_topday">TopDag</string>
<string name="sorttype_topweek">TopWeek</string>
<string name="sorttype_topmonth">TopMaand</string>
<string name="sorttype_topyear">TopJaar</string>
<string name="sorttype_topall">TopAlles</string>
```

kuroi-usagi: Agreed. Fixed it as well.

Issue #571, opened 2023-06-12
Increase the blurriness of nsfw posts


Monoxide8969:
- [ ✅] Did you check to see if this issue already exists?
- [ ✅] Is this only a single feature request? Do not put multiple feature requests in one issue.
- [ ✅] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/jerboa

Blurred NSFW posts as of now don't leave much room for imagination, it's not very effective and I think the blurriness should be increased so that nobody's able to tell what's going on before opening the photo.

Should be a good beginner issue too.

PR #576, opened 2023-06-13
Increased NSFW blur

XanderV2001:
I've increased the blur from 20 to 100.

This way it is harder to see what is behind the blur and thus hides the NSFW posts more effectively.

Fixes this issue: https://github.com/dessalines/jerboa/issues/571


twizmwazin (approved):

Monoxide8969:
Thank you

Issue #565, opened 2023-06-12
Gifs are static


ngosang:
** Jerboa Version **
0.33

**Describe the bug**
This post has an animated git. It's not moving in Jerboa app.
https://lemmy.fmhy.ml/post/25075

PR #580, opened 2023-06-13
Implement animated gifs when clicking on an image

beatgammit:
I only did it when clicking on an image for simplicity.

I thought about animating it on page load and maybe as a thumbnail, but I figured it would probably need a setting configuration and I didn't want to overcomplicate the PR.

Implements #565


dessalines (approved):
LGTM, thanks.

Issue #556, opened 2023-06-12
Crash on opening specific link


helpimnotdrowning:
** Jerboa Version **
0.0.33

**Describe the bug**
In this post to c/Anime@lemmy.ml : https://feddit.de/post/761086 , clicking on the markdown-styled link in the first paragraph to c/japanesemusic@feddit.de crashes the app.

I haven't seen any other links that also do this, but I will update if I find anything.

**To Reproduce**

1. Go to https://feddit.de/post/761086
2. Click on the link in the 1st paragraph 
3. App crashes

Using:
* LG v60 (root)
* Android 11
* Jerboa 0.0.33 via Play Store
* My user host: lemmy.sdf.org

awdsns:
Woohoo I broke the app!

So this link is a relative one (no protocol or host part, only `/c/japanesemusic@feddit.de`) which AFAIK is currently the only way to do sort-of portable links to communities. At least this works with the web UI.

FWIW for me Jerboa opens Chrome with the link, which then obviously doesn't work because of the missing host.

In general it would be nice if Jerboa (and other frontends) could just natively convert `!community@instance` into links by itself, then such hacks wouldn't be necessary.

awdsns:
Related: #454

awdsns:
Nevermind, for me it crashes too with 0.0.33 from Play Store. I think it was with a previous version that those links opened in Chrome.

@Anna-log7 might this be related to #413?

awdsns:
But would that community name `name@instance@instance` cause the app to crash?

I can't install Android Studio to debug it right now and zero experience with Jetpack Compose, but if you have a link target of *only* `/c/somecommunity@someinstance`, could that still match on the `uriPattern = "{instance}/c/{name}"` and cause the null check on `instance` to blow up?

https://github.com/dessalines/jerboa/blob/e4923c9308ba618ef80a0aa31c5fb8c411d9987d/app/src/main/java/com/jerboa/MainActivity.kt#L207

beatgammit:
Ok, this should behave a lot better now, please retest.

twizmwazin:
Given the lack of updated info and the number of changes that have happened surrounding link handling, I'm going to mark this one as complete.

Issue #454, opened 2023-06-07
Rewrite links in /c/community@server or /u/user@server in appropriate ways


Barbariandude:
In lemmy-ui, a link formatted in the ways described in the title are automatically rewritten to the appropriate URL by prepending the URL of the instance the account is from. This functionality is missing from Jerboa. Not the most urgent thing in the world as users can use the website instead, but definitely something that should be looked at at some point.

dessalines:
Is this referring to the copy permalink button? Or content that's in the markdown fields?

If its the permalink, we should probably use the original links, and if its the latter, jerboa doesn't even have a smart auto-filler for users and communities yet like lemmy-ui does.

And when it does, we'll need to match lemmy-ui, and add both links (a local and a fedi-link) : https://github.com/LemmyNet/lemmy-ui/issues/1048

Barbariandude:
If I format a link like [this]\(/c/lemmy@lemmy.ml) in the web UI, then a user sees the link as https://my.server/c/lemmy@lemmy.ml. This is very useful funcionality once people get used to it. It would be nice if the functionality matched in Jerboa.

beatgammit:
This has been merged and I believe resolved.

PR #583, opened 2023-06-13
Handle /c/community@host URLs

beatgammit:
This is a partial fix to #556, but it at least prevents crashing when an unknown URL type is encountered. There is still some work to do, since the markdown plugin interprets the following as mailto: urls:

- @user@instance
- !community@instance

I'm not sure what the supported URL types are intended to be, so I didn't bother implementing a plugin.

### DIFF
--- a/app/src/main/java/com/jerboa/Utils.kt
+++ b/app/src/main/java/com/jerboa/Utils.kt
@@ -383,13 +383,14 @@ fun LazyListState.isScrolledToEnd(): Boolean {
  * Parses a "url" and returns a spec-compliant Url:
  *
  * - https://host/path - leave as-is
+ * - http://host/path - leave as-is
  * - /c/community -> https://currentInstance/c/community
  * - /c/community@instance -> https://instance/c/community
  * - !community@instance -> https://instance/c/community
  * - @user@instance -> https://instance/u/user
  */
 fun parseUrl(url: String): String? {
-    if (url.startsWith("https://")) {
+    if (url.startsWith("https://") || url.startsWith("http://")) {
         return url
     } else if (url.startsWith("/c/")) {
         if (url.count({ c -> c == '@' }) == 1) {
--- a/app/src/main/java/com/jerboa/Utils.kt
+++ b/app/src/main/java/com/jerboa/Utils.kt
@@ -379,7 +379,49 @@ fun LazyListState.isScrolledToEnd(): Boolean {
     return out
 }
 
+/*
+ * Parses a "url" and returns a spec-compliant Url:
+ *
+ * - https://host/path - leave as-is
+ * - /c/community -> https://currentInstance/c/community
+ * - /c/community@instance -> https://instance/c/community
+ * - !community@instance -> https://instance/c/community
+ * - @user@instance -> https://instance/u/user
+ */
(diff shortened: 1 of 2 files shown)


beatgammit:
Also partial fix for #454

app/src/main/java/com/jerboa/Utils.kt
```diff
@@ -379,6 +379,16 @@ fun openLink(url: String, ctx: Context, useCustomTab: Boolean) {
     if (useCustomTab) {
         val intent = CustomTabsIntent.Builder()
             .build()
+
```
dessalines, line 382: Why does this only apply to custom tabs?

app/src/main/java/com/jerboa/Utils.kt
```diff
@@ -379,6 +379,16 @@ fun openLink(url: String, ctx: Context, useCustomTab: Boolean) {
...
+        val url = when {
+            url.startsWith("https://") -> url
+            url.startsWith("/c/") && url.count({ c -> c == '@' }) == 1 -> {
+                val (community, host) = url.split("@", limit = 2)
+                "https://$host$community"
+            }
+            else -> return
```
dessalines, line 389: Externalize this to a function, and don't use when. Use if else, or ?? url

beatgammit: Since this list will probably grow (I'm guessing `!community@instance` and `/u/user@instance`/`@user@instance`), is there a way you'd prefer to handle this? I could switch it to regex instead if that's preferable.

app/src/main/java/com/jerboa/Utils.kt
beatgammit: Fixed

app/src/main/java/com/jerboa/Utils.kt
beatgammit: Ok, fixed. The new code is pretty repetitive, but hopefully it's easy to follow. Maybe I'll get around to cleaning it up a bit if I can think of something clever.

beatgammit:
I redid the PR and added some tests. The code is kind of repetitive, but hopefully easy to follow.

I'm interested in upgrading to JUnit 5 and adding a bunch of tests, so let me know if there's a good reason not to do that before I get too deep into it.

twizmwazin (approved):

app/src/main/java/com/jerboa/Utils.kt
```diff
@@ -379,7 +379,49 @@ fun LazyListState.isScrolledToEnd(): Boolean {
...
+ * Parses a "url" and returns a spec-compliant Url:
+ *
+ * - https://host/path - leave as-is
+ * - /c/community -> https://currentInstance/c/community
+ * - /c/community@instance -> https://instance/c/community
+ * - !community@instance -> https://instance/c/community
+ * - @user@instance -> https://instance/u/user
+ */
+fun parseUrl(url: String): String? {
+    if (url.startsWith("https://")) {
```
twizmwazin, line 392: Can we also handle http (not https) links?

beatgammit: Done. :)

PR #593, opened 2023-06-13
Comment action bar fixes

a1studmuffin:
The "Show action bar by default for comments" setting now only affects comment trees in post listing views, not single comments in the user's inbox (replies/mentions), saved comments, or profile comments. This gives them convenient access to the action bar for single comments they're likely to want to interact with.

Also allow user to toggle action bar visibility for inbox replies and mentions by long tapping on the comment body (not just the comment header), which makes this behaviour consistent with long tapping on comments in normal post listings.


dessalines (approved):
Works for me, thx!

Issue #421, opened 2023-06-05
Creating a new post with an image and cancelling the image selection crashes Jerboa


7heo:
Pretty much what the title says.

Steps to reproduce:

1. Go create a new post
2. Tap on "upload image"
3. Exit the image selection program without selecting one
4. Profit.

7heo:
This issue needs to be re-open.

noloman:
I've got the fix. I'll create a PR later today.

PR #600, opened 2023-06-13
Remove unnecessary check

7heo:
According to flutter's app:compileDebugKotlin job, in the context of CreatePostActivity.kt, 'uri != null' is always 'true'.


dessalines (approved):
There are a lot of these that need fixing.

7heo:
> There are a lot of these that need fixing.

I will try and help, once I fixed #421 for good.

PR #609, opened 2023-06-13
Support android 13's per-app language selection, fixes #603

### LINKED ISSUE #603: App language settings

It'd be great if jerboa would let us choose which language to display the app in, instead of auto localising everything.

Alternatively, if you don't want to create an in-app language picker, it would be cool if you could at least enable default app language in the android app settings via android 13's per-app language preferences.

twizmwazin:
Also fixing deprecated permission.

### DIFF
--- a/app/src/main/res/xml/locales_config.xml
+++ b/app/src/main/res/xml/locales_config.xml
@@ -5,6 +5,6 @@
     <locale android:name="it"/>
     <locale android:name="ko"/>
     <locale android:name="nl"/>
-    <locale android:name="pt-rBR"/>
+    <locale android:name="pt-BR"/>
     <locale android:name="se"/>
 </locale-config>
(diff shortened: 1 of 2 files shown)


app/src/main/res/xml/locales_config.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<locale-config xmlns:android="http://schemas.android.com/apk/res/android">
+    <locale android:name="en"/>
+    <locale android:name="de"/>
+    <locale android:name="it"/>
+    <locale android:name="ko"/>
+    <locale android:name="nl"/>
+    <locale android:name="pt-rBR"/>
```
dessalines, line 8: pt-rBR correct?

twizmwazin: Android studio shows these abbreviations in the android project sidebar, which is where I got that from
image

You're right though, its supposed to just be `pt-BR`.

dessalines (approved):

PR #614, opened 2023-06-13
Use image previewer when clicking on images in small cards and list view

twizmwazin:
Fixes #513

PR #625, opened 2023-06-14
don't have translucent background on image post thumbnail click

vijaykramesh:
right now there's this backColorTranslucent applied when you click on the image link preview for a post:

this looks kind of funky as the preview is then centered over the original post and you get doubling of the image with the translucent on in the background, until you click and then it black out the background. 

this proposed change would just always have the blacked out background and not do the translucent thing until you click:


app/src/main/java/com/jerboa/ui/components/common/ImageViewerDialog.kt
```diff
@@ -73,7 +73,6 @@ fun ImageViewerDialog(url: String, onBackRequest: () -> Unit) {
     }
 
     val backColor = MaterialTheme.colorScheme.scrim
-    val backColorTranslucent = MaterialTheme.colorScheme.scrim.copy(alpha = 0.4f)
```
dessalines, line 76: I actually do really like the background translucence, but I agree 40% in way too little... it should be like 10-20% . Could you try that out and see how it looks?

vijaykramesh:
OK updated this, the alpha is for the black overlay not for the underlying background, so 0.8 seems to look the nicest (the underlying background is still visible but doesn't interfere with the image you have in the `ImageViewerDialog`  

it seems like when I screenshot it on my phone it doesn't quite capture the effect but let me know if you think this is a reasonable change (at least with 0.8 it seems way less distracting to me than 0.4 is)

vijaykramesh:
also just super impressed with how fast people are fixing bugs here, in this same MR i went to fix the small card click for images and then upon rebasing saw https://github.com/dessalines/jerboa/pull/614 already fixed it!

PR #651, opened 2023-06-14
Show link dialog on long press

pipe01:
When long pressing a post thumbnail or image a dialog will show containing actions related to the post's link, similar to Boost. If the link is an image, an option to download it will be added.

image

### DIFF
--- a/app/src/main/java/com/jerboa/Utils.kt
+++ b/app/src/main/java/com/jerboa/Utils.kt
@@ -15,6 +15,7 @@ import android.os.Environment
 import android.provider.MediaStore
 import android.util.Log
 import android.util.Patterns
+import android.webkit.MimeTypeMap
 import android.widget.Toast
 import androidx.browser.customtabs.CustomTabsIntent
 import androidx.compose.foundation.ExperimentalFoundationApi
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/jerboa/Utils.kt
```diff
@@ -881,6 +884,23 @@ fun saveBitmap(
...
+    val extension = MimeTypeMap.getFileExtensionFromUrl(url)
+    val mimeType = MimeTypeMap.getSingleton().getMimeTypeFromExtension(extension)
+
+    withContext(Dispatchers.IO) {
+        URL(url).openStream().use {
+            saveBitmap(context, it, mimeType, fileName)
+        }
+    }
+
+    Toast.makeText(context, "Saved image", Toast.LENGTH_SHORT).show()
```
MV-GH, line 901: "Saved image" should be a stringResource

twizmwazin (changes requested):
Please use string resources over literals, then it'll lgtm.

app/src/main/java/com/jerboa/Utils.kt
```diff
@@ -881,6 +884,23 @@ fun saveBitmap(
     }
 }
 
+suspend fun saveImage(url: String, context: Context) {
+    Toast.makeText(context, "Saving image...", Toast.LENGTH_SHORT).show()
```
dessalines, line 888: stringResource.

MV-GH:
Obsolete by #1189

Issue #662, opened 2023-06-15
comment upvote/downvote count doesn't consistently display


vijaykramesh:
** Jerboa Version **
0.0.34 installed from apk from github

**Describe the bug**
See https://lemmy.ml/comment/641080

Comment upvote/downvote counts don't show unless you downvote. If you upvote they still don't show

**To Reproduce**
See comment no count
Downvote count
Upvote no count

![Screenshot_20230614_220330_Jerboa for Lemmy](https://github.com/dessalines/jerboa/assets/556288/7c0ff80e-8ce8-419b-bf9d-94a18f0eedb9)

![Screenshot_20230614_220321_Jerboa for Lemmy](https://github.com/dessalines/jerboa/assets/556288/8ddbd3f4-ce1a-4a52-a3d0-8098ad86d956)

vijaykramesh:
This is also oddly inconsistent in lemmy ui - when I downvote I see both counts but when I upvote I don't

![Screenshot_20230615_070153_Firefox](https://github.com/dessalines/jerboa/assets/556288/5317a505-1f01-4344-bed8-651812335465)
![Screenshot_20230615_070135_Firefox](https://github.com/dessalines/jerboa/assets/556288/014295ef-d987-4d49-b274-c3397e8d2574)

vijaykramesh:
![Uploading Screenshot_20230615_070141_Firefox.jpg…]()

vijaykramesh:
Validates in #667 the behavior is correct closing this and that

PR #667, opened 2023-06-15
Make comment up/down vote consistent after you vote

vijaykramesh:
this fixes https://github.com/dessalines/jerboa/issues/662

it makes it so for a comment you don't see the upvote/downvote specific count unless you vote - and then if you upvote or downvote you'll see both specific counts.

currently as #662 describes you only see this when you downvote, if you upvote you don't see it.

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/comment/CommentNode.kt
+++ b/app/src/main/java/com/jerboa/ui/components/comment/CommentNode.kt
@@ -565,7 +565,7 @@ fun CommentFooterLine(
                 item = commentView,
                 type = VoteType.Upvote,
                 onVoteClick = onUpvoteClick,
-                showNumber = (instantScores.downvotes != 0),
+                showNumber = (instantScores.myVote?.let { it != 0 } ?: let { false }),
                 account = account,
             )
             if (enableDownVotes) {
@@ -575,6 +575,7 @@ fun CommentFooterLine(
                     item = commentView,
                     type = VoteType.Downvote,
                     onVoteClick = onDownvoteClick,
+                    showNumber = (instantScores.myVote?.let { it != 0 } ?: let { false }),
                     account = account,
                 )
             }
--- a/app/src/main/java/com/jerboa/ui/components/common/VoteHelpers.kt
+++ b/app/src/main/java/com/jerboa/ui/components/common/VoteHelpers.kt
@@ -27,11 +27,7 @@ fun <T> VoteGeneric(
     }
 
     val votesStr = if (showNumber) {
-        if (type == VoteType.Downvote && votes == 0) {
-            null
-        } else {
-            votes.toString()
-        }
+        votes.toString()
     } else {
         null
     }


app/src/main/java/com/jerboa/ui/components/comment/CommentNode.kt
```diff
@@ -565,7 +565,7 @@ fun CommentFooterLine(
                 item = commentView,
                 type = VoteType.Upvote,
                 onVoteClick = onUpvoteClick,
-                showNumber = (instantScores.downvotes != 0),
+                showNumber = (instantScores.myVote?.let { it != 0 } ?: let { false }),
```
dessalines, line 568: Might have been a bug someone slipped in, but really it should always show the upvotes, no matter what. So this should just be `showNumber = true`

The downvotes below, should only be shown, if there are downvotes, so `instantScores.downvotes != 0`

It doesn't have anything to do with whether *you* voted or not.

vijaykramesh: That is how the number on the right works but this is for the numbers next to the arrows for the comments, currently it is only showing up if you downvote, if you upvote they don't show. But neither show until you vote 

You are saying always show the upvote number (even before voting) but then show downvote if they click down?

vijaykramesh: It looks like lemmy ui does the same. Always show the up-down number on the right, only show the numbers by the arrow if you downvote, and never show the individual numbers if you upvote. My proposal would be always show the specific up and down numbers once you vote (this is how it works on the top level post up/down buttons)

vijaykramesh: Assuming you agree I can take a stab at changing this in lemmy-ui to match too

dessalines: Ohhh wait this is for the footer line.

Okay this is hard to articulate, but if there are no downvotes, it should only show the score, on the top right. It would be pointless to show `+5` on the top right, and `+5` on the footer bar.

If there are downvotes, this changes.

The score would then be `+4` on the top right, but the comment footer should now show `+5, -1` . I'm pretty sure this is how it already works in both lemmy-ui and jerboa.

vijaykramesh: Ahhh ok I follow now. I don't think it is actually working like that but let me test and see

vijaykramesh: I finally had time to see exactly what you mean and can confirm correct behavior in lemmy ui. (Sorry posting from my phone I'll go on my computer later to resize this)

Tldr if there are prexisting downvotes you see the detailed up/down count and total on the right. For that case if you up/down you keep seeing all 3 numbers 

If there are no down you only ever see right total. And if you upvote in this case still no up/down breakout needed or shown.

I'll double check jerboa later and I think will close this and the issue that came out of a !jerboa@lemmy.ml thread. Thanks for helping validate

PR #671, opened 2023-06-15
Fix image upload cancel crash

### LINKED ISSUE #421: Creating a new post with an image and cancelling the image selection crashes Jerboa

noloman:
Closes #421 

[Screen_recording_20230615_172521.webm](https://github.com/dessalines/jerboa/assets/183264/6ee27971-da77-4c9a-ad21-80fb0ecf42d2)


7heo:
@noloman Great, thanks, TIL about `let`. Quick question, when I did my initial patch (I essentially had the fix :tada:, just the wrong call...), flutter later reported, during build, that url != null would always evaluate to 'true'. Why does that work with `let`? Thanks!

PR #672, opened 2023-06-15
Fixed image not loading & added coil logger for easy debugging

subashz:
Previous commit broke the image (thumbnail) loading


dessalines:
Don't open up multiple PRs with the same commits please.

PR #674, opened 2023-06-15
Small Accesibility Tweaks

### LINKED ISSUE #616: Small Accessibility Tweaks
** Jerboa Version ** 0.0.34

**Describe the bug**
Some small accessibility tweaks are still required. Screenshots and suggestions below.

- [ ] in screenshot 1 talkback reads "Todo DeGoogle Yourself at lemmy ml". It should not read " todo" which i am assuming is likely coming from the generic forum icon to the left as other subs that have icons do not do this. 
Jerboa 1

- [ ] in screenshot 2 the "select listing filter" button is highlighted. To make it clearer i suggest changing the string to "select feed" 
Jerboa 2_1

- [ ] in screenshot 3 the "select sort option" button is highlighted. I suggest replacing the string with "sort" or "sort by" 
Jerboa 3_1

- [ ] in screenshot 4 the "show more options" button is highlighted. I suggest changing the string to "more options" 
Jerboa 4_1

- [ ] in screenshot 5 the "toggle vote" button is highlighted for upvoting a post (and i assume comments use it too and this applies. This button should read as "upvote" when in this state and "upvoted" when it has been pressed. 
Jerboa 5_1

- [ ] in screenshot 6 the "toggle vote" button is highlighted for downvoting a post (and i assume comments use it too and this applies. This button should read as "downvote" when in this state and "downvoted" when it has been pressed. 
Jerboa 6_1

Edit 1:
- [ ] the "select inbox filter" should change to "filter"
- [ ] the "select sort option" in the bookmarks tab should be changed to "sort" or "sort by". See screenshot 7. 
Screenshot_20230613-173008_Jerboa_for_Lemmy
- [ ] on the profile page the " select sort option" button should be changed to "sort" or "sort by". 
Screenshot_20230613-173247_Jerboa_for_Lemmy

scottmmjackson:
- Dropped "todo" from the community icon description
- Changed "Select listing filter" to "select feed"
- Changed "Select sort option" to "sort by"- this affects multiple components
- Changed "Show more options" to "more options"
- Changed the way vote component descriptions are written. The type and vote are used to select among "upvote", "upvoted", "downvote", and "downvoted"
- Changed "Select inbox filter" to "filter"

Fixes #616

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -119,7 +119,7 @@
     <string name="create_report_report">Report</string>
     <string name="create_report_type_your_reason">Type your reason</string>
     <string name="create_report_view_model_report_created">Report Created</string>
-    <string name="dialog_moreOptions">Show more options</string>
+    <string name="dialog_moreOptions">More options</string>
     <string name="dialogs_active">Active</string>
     <string name="dialogs_all">All</string>
     <string name="dialogs_done">Done</string>
@@ -139,7 +139,7 @@
     <string name="dialogs_unread">Unread</string>
     <string name="floating_createPost">Create post</string>
     <string name="form_submit">Submit</string>
-    <string name="homeHeader_filter">Select listing filter</string>
+    <string name="homeHeader_filter">Select feed</string>
     <string name="home_add_account">Add Account</string>
     <string name="home_all">All</string>
     <string name="home_communities">Communities</string>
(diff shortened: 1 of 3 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -289,4 +289,8 @@
     <string name="topAppBar_back">Back</string>
     <string name="utils_login_first">Login first</string>
     <string name="voteToggle">Toggle vote</string>
```
scottmmjackson, line 291: Didn't drop this because I'm not sure of the implications since it's been translated a bunch

twizmwazin (approved):

app/src/main/res/values/strings.xml
MV-GH: If they aren't used, they should be deleted including the translations imo. Would need a code owner to make the final verdict.

twizmwazin: If its unused and not expected to be used again we should remove it, but it doesn't have to be in this PR necessarily.

scottmmjackson: - [ ] Drop `voteToggle` unused string resource

PR #676, opened 2023-06-15
Dev image

subashz:
Show circular loading bar when viewing image in full screen preview.


twizmwazin (approved):

twizmwazin:
Actually, it seems like This includes changes from #672. Is this PR strictly dependent on that one?

dessalines (approved):
Looks good, was wondering why images were broken. Thx!

dessalines:
In that case I'll close your other one.

dessalines (commented):
Needs conflicts resolved.

PR #680, opened 2023-06-15
Clicking on link opens a webview

subashz:
Right now, if we click on thumbnail of the post - type link, it just redirects to the ImageViewerDialog with no image url to load. This commit fix the issue by opening the url post in the new tab.


twizmwazin:
Yeah it should work better in 0.0.34. This was caused by [my commit](https://github.com/dessalines/jerboa/commit/25d3f61c6be805ba34ef0914bde6b8e060de994d), which tried to copy the behavior of boost here. What we need to do is, like boost, be able to determine if the content is an image before choosing to open it in the image viewer. If you can rebase we can merge this and then work on fixing it better later.

twizmwazin:
Please format with `./gradlew formatKotlin`

dessalines:
Also needs conflicts resolved.

Issue #526, opened 2023-06-11
Host translations in Weblate


ngosang:
- [x] Did you check to see if this issue already exists?
- [x] Is this only a single feature request? Do not put multiple feature requests in one issue.
- [x] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/jerboa

**Describe the feature request below**

I see you already have internationalization / multi-language support. I will like to help with translation but I don't want to keep track of the changes in GitHub.
Could you add some kind of automation that syncs the `.xml` files with the translations with [Weblate](https://hosted.weblate.org/)? It's easy to translate in their Web UI and I will receive notifications when new translations are required.

https://hosted.weblate.org/

ngosang:
Lemmy is already doing that => https://weblate.join-lemmy.org/projects/lemmy/

dessalines:
Closing as not really needed, ppl have been doing PRs to strings.xml.

PR #683, opened 2023-06-15
Implement !community@instance and @user@instance links

### LINKED ISSUE #556: Crash on opening specific link

beatgammit:
They merely pop up a toast saying it's unimplemented, but the links are being created and handled.

Fixes #556.

Partial fix for #526 since short forms of lemmy links can be handled separately from other links.

Related to #454.

### DIFF
--- a/app/src/main/java/com/jerboa/Utils.kt
+++ b/app/src/main/java/com/jerboa/Utils.kt
@@ -423,6 +423,14 @@ fun parseUrl(url: String): String? {
 fun openLink(url: String, ctx: Context, useCustomTab: Boolean, usePrivateTab: Boolean) {
     val url = parseUrl(url) ?: return
 
+    if (url.startsWith("@")) {
+        Toast.makeText(ctx, "User info links not yet supported", Toast.LENGTH_SHORT).show()
+        return
+    } else if (url.startsWith("!")) {
+        Toast.makeText(ctx, "Community links not yet supported", Toast.LENGTH_SHORT).show()
+        return
+    }
+
     if (useCustomTab) {
         val intent = CustomTabsIntent.Builder()
             .build().apply {
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -105,6 +105,7 @@ dependencies {
 
     // optional - Test helpers
     testImplementation "androidx.room:room-testing:2.5.1"
+    testImplementation  "pl.pragmatists:JUnitParams:1.1.1"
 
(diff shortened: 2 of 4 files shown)


twizmwazin:
This looks really great! Look forward to when its ready.

beatgammit:
I'm happy to merge as-is, or I can take a couple days and figure out how to open community and user pages.

Preference?

app/src/main/java/com/jerboa/Utils.kt
```diff
@@ -423,6 +423,14 @@ fun parseUrl(url: String): String? {
 fun openLink(url: String, ctx: Context, useCustomTab: Boolean, usePrivateTab: Boolean) {
     val url = parseUrl(url) ?: return
 
+    if (url.startsWith("@")) {
```
twizmwazin, line 426: Following #583, I don't think this section triggers. `parseUrl` will return a standard https link. This should probably be rewritten to check if a link appears to be a user or community link, and then handle it.

twizmwazin: I got excited and ended up doing that: https://github.com/twizmwazin/jerboa/tree/feat/implement-lemmy-links

beatgammit: I looked through your changes, and you may want to consider setting `navController` outside of the `MarkdownHelper.init()` so it can easily be used in previews. What I did was make it nullable and made a setter for it, and then check it before passing it on to openLinks.

Just an idea, I'm certainly no expert here. 😄

twizmwazin:
I merged this as part of #692, thank you!

app/src/main/java/com/jerboa/Utils.kt
twizmwazin: Yeah I think you're right, would welcome a PR for it.

PR #686, opened 2023-06-15
Added Japanese Translation Strings

ShinyLuxray:
Hello. I added a base set of translation strings for Jerboa in Japanese. It might not be perfect but I think it fits most of the things.  - Message Rentlar@lemmy.ca if you have questions.


ShinyLuxray:
I've added additional edits to this pull request, allowing for translated Home Menu, Inbox, Comment and Community headers to show up in the app.

I still haven't found where the Inbox subheaders ("Replies", "Mentions", "Messages") or the word "Saved" is in the code in order to be localized.

app/src/main/java/com/jerboa/Utils.kt
```diff
@@ -991,3 +994,39 @@ fun getLocalizedStringForUserTab(ctx: Context, tab: UserTab): String {
     }
     return returnString
 }
+
+/**
+ * Returns localized Strings for ListingType Enum
+ */
+fun getLocalizedListingTypeName(context: Context, listingType: ListingType): String {
```
dessalines, line 1001: Follow the convention above: `ctx`

app/src/main/java/com/jerboa/Utils.kt
```diff
@@ -991,3 +994,39 @@ fun getLocalizedStringForUserTab(ctx: Context, tab: UserTab): String {
...
+        ListingType.Local -> context.getString(R.string.home_local)
+        ListingType.Subscribed -> context.getString(R.string.home_subscribed)
+    }
+    return returnString
+}
+
+/**
+ * Returns localized Strings for CommentSortType Enum
+ */
+fun getLocalizedCommentSortTypeName(context: Context, commentSortType: CommentSortType): String {
```
dessalines, line 1013: Same.

app/src/main/java/com/jerboa/Utils.kt
```diff
@@ -991,3 +994,39 @@ fun getLocalizedStringForUserTab(ctx: Context, tab: UserTab): String {
...
+        CommentSortType.Old -> context.getString(R.string.sorttype_old)
+        CommentSortType.Top -> context.getString(R.string.dialogs_top)
+    }
+    return returnString
+}
+
+/**
+ * Returns localized Strings for UnreadOrAll Enum
+ */
+fun getLocalizedUnreadOrAllName(context: Context, unreadOrAll: UnreadOrAll): String {
```
dessalines, line 1026: Same.

dessalines (commented):
Thanks for catching these!

In the future, it'd be better to split up these two PRs: one for the translations, and another for the localized strings.

app/src/main/java/com/jerboa/Utils.kt
ShinyLuxray: OK thanks for the heads up. I was looking for a convention to follow for naming, I found in Utils.kt context: Context instead of ctx was being used in convertSpToPx and getLocalizedSortinTypeName methods so I was following that. I'll make the fixes now and try to push them to this branch.

ShinyLuxray:
This morning I should have changed the context variable/value names to ctx as per the desired convention. Feel free to let me know if there are any other changes that should be made. Sorry that I'm not the best at Git.

dessalines (approved):
LGTM, thanks! I'll handle the merge conflicts.

Issue #539, opened 2023-06-11
Post age has no unit


axelrechenberg:
** Jerboa Version **
0.0.33

**Describe the bug**
Post age has no unit - shows just a number without an indication whether that's seconds, minutes, hours etc.

**To Reproduce**
Happens every single time if my phone's language is set to Polish.

![Screenshot_2023-06-11-21-48-12-32_a10f969cf1c4adf2c503042b459b8476](https://github.com/dessalines/jerboa/assets/20260799/a359a26f-a6ef-4b01-86c6-e8d27ccc516d)

gressen:
![Screenshot_20230611_220109_Jerboa for Lemmy](https://github.com/dessalines/jerboa/assets/8057552/a6f555d4-bbfb-4026-8b2b-a92a39e5724f)
I'm experiencing the same bug, phone set to Polish as well.

sensiblepuffin:
I have to go to sleep, but I figured I'd share what I found -

I tried 5 languages (English, Spanish, Polish, Korean, and Danish) and Polish was the only one that triggered this issue. 

What's interesting is that, of these languages, Polish is the only locale for which `override` is defined in org.ocpsoft.prettytime.impl.ResourcesTimeFormat:L112:

`return override == null ? super.format(duration) : override.format(duration);`

For Polish, this override takes us to org.ocpsoft.prettytime.i18n.Resources_pl:L48, where only the number is returned:

```      
public String format(Duration duration)
{
  long quantity = duration.getQuantityRounded(tolerance);
  return String.valueOf(quantity);
}
```

as opposed to org.ocpsoft.prettytime.format.SimpleTimeFormat:L95, which is where we go when there is no override above.

```
private String format(final Duration duration, final boolean round)
{
  String sign = getSign(duration);
  String unit = getGramaticallyCorrectName(duration, round);
  long quantity = getQuantity(duration, round);
  
  return applyPattern(sign, unit, quantity);
}
```

Update: Okay, I got curious and tried Russian. It, too, has an override there, and it too has missing time units. So `format()` behaves differently for languages with override ResourceBundles, and from what I can see it will never return the time unit.

lbenedetto:
I've filed a bug report with PrettyTime
https://github.com/ocpsoft/prettytime/issues/259

dawid2193487:
This might be happening because Polish has different plurals for different numbers.

1 sekund*a*, 2..4 sekund*y*, 5.. sekund

[...]

PR #687, opened 2023-06-15
#539 PrettyTime Missing Unit Workaround

lbenedetto:
This is a workaround for #539 

A bug in PrettyTime causes four languages to not get any time units when formatted. 

We can work around this by detecting an invalid prettyDate (rather than hardcoding a list of broken languages) and switching to English formatting instead.

https://github.com/ocpsoft/prettytime/issues/259


app/src/main/java/com/jerboa/Utils.kt
```diff
@@ -467,6 +465,28 @@ fun openLink(url: String, navController: NavController, useCustomTab: Boolean, u
...
+val invalidPrettyDateRegex = "^[0123456789 ]+$".toRegex()
+fun formatDuration(date: Date, longTimeFormat: Boolean = false): String {
+    if (prettyTime.locale != Locale.getDefault()) {
+        prettyTime = PrettyTime(Locale.getDefault())
+    }
+
+    var prettyDate = prettyTime.formatDuration(date)
+
+    // A bug in PrettyTime means that some languages (pl, ru, uk, kk) will not include any time unit
+    if (prettyDate.matches(invalidPrettyDateRegex)) {
```
dessalines, line 479: Where's the bug in prettytime? It seems pretty active. Definitely open up a PR over there, and reference it here, so that we don't always have to use this hack.

lbenedetto: I linked the pretty time bug report in the PR description. 

But really, it's not so much a bug as a lack of support for these languages. For other languages, they have maps of words in those languages and logic to choose those words with the correct plurality and such.

However, for these four languages, the implementation does not contain any words or logic and just returns back whatever number was passed in.

I wrote this workaround here so that if/when the issue is fixed, it won't continue defaulting to English, it'll just become irrelevant.

PR #691, opened 2023-06-16
#602 Change post button to be more obvious

### LINKED ISSUE #602: Post Button
- [ ] Did you check to see if this issue already exists?
- [ ] Is this only a single feature request? Do not put multiple feature requests in one issue.

I'd like to see the Post Button, which is currently a "+” sign, be changed to the word Post, or some other blatantly obvious button, so the user doesn't think it's an " additional options" thing.  Even here on Github it's a blue arrow that stands out from the rest of the screen.

Chris-Kropp:
resolves #602 This changes the submit icon from `add` to `submit` and adds a preview for the post header.
Icon_only

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/post/create/CreatePost.kt
+++ b/app/src/main/java/com/jerboa/ui/components/post/create/CreatePost.kt
@@ -76,17 +76,14 @@ fun CreatePostHeader(
                     )
                 } else {
                     // Todo add are you sure cancel dialog+
-                    if (formValid && !loading) {
-                        Icon(
-                            Icons.Filled.Send,
-                            contentDescription = stringResource(R.string.form_submit),
-                        )
-                    } else {
-                        Icon(
-                            Icons.Outlined.Send,
-                            contentDescription = stringResource(R.string.form_submit),
-                        )
-                    }
+                    Icon(
+                        if (formValid) {
+                            Icons.Filled.Send
+                        } else {
+                            Icons.Outlined.Send
+                        },
+                        contentDescription = stringResource(R.string.form_submit),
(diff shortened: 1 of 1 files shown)


twizmwazin (changes requested):
Your change looks great. Can you fix the commit history?

twizmwazin:
Run `./gradlew formatKotlin` to fix the formatting

app/src/main/java/com/jerboa/ui/components/post/create/CreatePost.kt
```diff
@@ -74,11 +75,18 @@ fun CreatePostHeader(
                         color = MaterialTheme.colorScheme.onSurface,
                     )
                 } else {
-                    // Todo add are you sure cancel dialog
-                    Icon(
-                        Icons.Outlined.Add,
-                        contentDescription = stringResource(R.string.form_submit),
-                    )
+                    // Todo add are you sure cancel dialog+
+                    if (formValid && !loading) {
```
dessalines, line 79: Loading was already checked above.

app/src/main/java/com/jerboa/ui/components/post/create/CreatePost.kt
```diff
@@ -74,11 +75,18 @@ fun CreatePostHeader(
...
+                        Icon(
+                            Icons.Filled.Send,
```
dessalines, line 81: Would be much easier to do `icon = if (formValid) { ...`

Chris-Kropp: Yep, that's much cleaner

Issue #627, opened 2023-06-14
Markdown spoilers and block quotes not rendered


trubeck:
- [x] Did you check to see if this issue already exists?
- [x] Is this only a single feature request? Do not put multiple feature requests in one issue.
- [ ] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/jerboa

As you can see in the screenshot, Markdown is not rendered in the overview. This poses a problem for me because I want to run a community for a book series and I want to hide spoilers, especially in the overview, so that new readers don’t miss out on the fun.

![tempFileForShare_20230614-085442](https://github.com/dessalines/jerboa/assets/13149855/f57175f1-ba08-460b-a9b2-4bd43722fa35)

twizmwazin:
I believe that markdown is rendered, but spoilers are not currently rendered.

trubeck:
Yes, spoilers and strike-through text are is not working. I created a separate issue for this: #628 

But in the overview, no Markdown is rendered at all.

beatgammit:
Strike through and block quotes seem to work on the latest in main, spoilers are a WIP.

ZJouba:
@twizmwazin Should we support `:::` or `<details>`? Also, do we want to support nested spoilers?

MV-GH:
This should be supported now.

PR #695, opened 2023-06-16
Add initial spoiler tag support

beatgammit:
Initial support for #627.

This renders this:

image

As this:

image

I'm not sure how to do the styling or behavioral bits, and I'll probably need to do something other than html -> Android elements, but this works for now as a POC since it parses the spoiler format and tries to render it.


dessalines (commented):
This would definitely be nice to have.. Are you ready for me to test it?

MV-GH:
Obsolete by  #990

PR #696, opened 2023-06-16
Fixes image saving crashing on Android 9 and below

### LINKED ISSUE #688: Crash when trying to save an image
** Jerboa Version **
0.0.34

**Describe the bug**
The application crashes when trying to save an image

**To Reproduce**

(0. Grant the Storage permission to the app)
1. Go to any timeline
2. Click on any image
3. Click the Download button on the top right
4. The application suddenly closes without saving the image

Using /e/OS Android 9 on Fairphone 3. Had it installed previously via F-Droid, issue was still there, this is from IzzyOnDroid.

MV-GH:
The mediastore api doesn't seem to be working on android 9 and below.
This line`resolver.insert(MediaStore.Images.Media.EXTERNAL_CONTENT_URI, values)` is always null
So i used a different method for those.

Also included the permissions android 9 and below needed for that method.
I could use a different approach that doesn't require the WRITE_EXTERNAL_STORAGE permission but then it would delete the images when the app is uninstalled. The location should be the same as before.

SaveImage should handle IO errors better since it doesn't handle it at all. For example when the device is out of space and your trying to save an image it would crash the app. But that's for another PR.

Also fixed missing string resources.

Tested it on android 8, 9 , 10 ,11  and 13

Fixes #688


dessalines (approved):
LGTM, thanks!

PR #708, opened 2023-06-16
Post link fix

beatgammit:
Redid PR #680 my pulling in code from main and fixing compile errors.

I'm still working on testing, so this will be in draft until at least later tonight when I get to it.


dessalines:
Now that auto_gen types is merged, this needs a merge from main.

beatgammit:
This has way more conflicts than is worth to fix, closing. We'll just have to implement it properly later.

PR #710, opened 2023-06-17
QOL Remove unnecessary clickable areas from PostListingList

### LINKED ISSUE #488: Easy to accidentally click a user profile or community link instead of post link in the main view
** Jerboa Version **
0.32

**Describe the bug**
In the Card/Small Card/List post views, it's easy to accidentally open the wrong link (user profile or community) instead of the comments.

image

image

**To Reproduce**

1. Click in the wrong spot :)

**Proposed fix**
Either change default behaviour or add a "fat finger/simple links" setting, so the highlighted elements open the post comments (like most other areas on the card) instead of the user profile/community links. The action bar for posts could be extended to add "Go to User Profile" and "Go to Community" accordingly, so these links are still reachable.

I'm happy to put together a PR for this if there's agreement.

lbenedetto:
The PostListingList is too small to have multiple clickable areas, leading many users to accidentally click into the community or the users profile from the front page instead of opening the post like they intended.

https://lemmy.ml/post/1303517

Fixes #488


app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
```diff
@@ -1017,15 +1013,17 @@ fun PostListingList(
...
+                            clickable = false,
                             showDefaultIcon = false,
                         )
                         DotSpacer(0.dp)
                     }
                     PersonProfileLink(
                         person = postView.creator,
                         isModerator = isModerator,
-                        onClick = onPersonClick,
+                        onClick = {},
```
dessalines, line 1025: I don't think this is a good idea. Now users won't be able to go to a person's profile, or a community, from the post listing, at all. 

@twizmwazin thoughts?

ZJouba:
I think an increase in the Heading font size would accomplish this without changing anything else?
Boost VS Jerboa
<p>
  
   
</p>

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
twizmwazin: There was na issue opened about this, I'll link it here if I can find it. Users (including myself) accidentally fat-finger the small touch targets frequently and it is a bit of an annoyance. It is simple enough to open the post and then click the appropriate link from in there. An option wouldn't be a bad idea, which is what I think Boost does.

twizmwazin: Here's the issue: https://github.com/dessalines/jerboa/issues/488

twizmwazin:
@ZJouba You're right, but there are many users who prefer a more compact layout, myself included, who would prefer this approach. Settings in the future could be a nice way to accommodate both preferences.

twizmwazin (approved):
LGTM once rebased

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
dessalines: mmmk, I spose its fine, especially since this just affects the list version.

dessalines (approved):

a1studmuffin:
With these changes, does clicking over a username or community name go to the underlying post instead, or does it just ignore the click and do nothing? I think the former would be a better experience for users.

lbenedetto:
> With these changes, does clicking over a username or community name go to the underlying post instead, or does it just ignore the click and do nothing? I think the former would be a better experience for users.

Clicking a username opens the post

PR #711, opened 2023-06-17
Fix crash when marking comment as read

beatgammit:
Small fix for marking inbox replies as read.


beatgammit:
Related to #657

dessalines (approved):
Good catch, thanks!

PR #715, opened 2023-06-17
Fix issue of an empty screen in inbox

beatgammit:
The issue was caused by the activity checking for a completed request in the PostViewModel. However, since this activity was triggered by the inbox, that request was never made.

The only data it needed was whether the user was a moderator, so I pass that information in with the deep link. In all cases we had enough information in the navigating activity to determine this, with the exception of the inbox, but we don't mark moderators in the inbox (at least lemmy-ui doesn't).

This is a little bit of a hack, but it's a relatively clean hack.

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/comment/reply/CommentReplyActivity.kt
+++ b/app/src/main/java/com/jerboa/ui/components/comment/reply/CommentReplyActivity.kt
@@ -30,6 +30,7 @@ fun CommentReplyActivity(
     postViewModel: PostViewModel,
     siteViewModel: SiteViewModel,
     navController: NavController,
+    isModerator: Boolean,
 ) {
     Log.d("jerboa", "got to comment reply activity")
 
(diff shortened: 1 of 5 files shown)


beatgammit:
Related to #657

app/src/main/java/com/jerboa/ui/components/comment/reply/CommentReplyActivity.kt
```diff
@@ -70,82 +71,68 @@ fun CommentReplyActivity(
             if (loading) {
                 LoadingBar(padding)
             } else {
-                when (val postRes = postViewModel.postRes) {
-                    is ApiState.Success -> {
-                        val moderators = postRes.data.moderators
+                commentReplyViewModel.replyItem?.let { replyItem ->
```
dessalines, line 74: I understand this fix, but not the other. What does being a moderator have to do with an inbox?

dessalines:
Nevermind, I see why now. Thx!

PR #718, opened 2023-06-17
- add translations to the top bar

perepepepa:
I added the missing code to use the translations for the top bar.


dessalines:
This has been added in a different PR now. But thanks for catching!

PR #720, opened 2023-06-17
Fix URL in pictrsImageThumbnail

APraxx:
Added Ampersand to return value of pictrsImageThumbnail which was missing

seems @beatgammit missed this

as trivial to fix, this is just a courtesy pr


camporter:
Fixes thumbnails here 👍 . formatting is upset with the braces around `$path` though.

dessalines:
I merged this as its a good short term fix, until @beatgammit's PR is ready to go.

PR #724, opened 2023-06-17
Add save post button in List view mode

### LINKED ISSUE #712: Add option to save post from list view

**Describe the feature request below**
Please add an option to save posts from list view.

Edit: Maybe a "swipe to save" would be applicable here so the list view doesn't get cluttered.

This has been requested in https://lemmy.ca/post/702033

noloman:
Closes #712 

Screenshot_20230617_170439


jacobbates:
Just adding my 2c that I think this suggested implementation overcrowds the UI for list view and makes interacting with the voting buttons harder.

How about having a long-hold menu option on the list item, or have long-hold reveal an action bar below the post in list view with more reasonably sized buttons? It would be nice to keep it consistent with the card views in that way, since currently list lacks the additional context menu for block, report, permalink. Save post could be grouped with these.

dessalines:
I agree with @jacobbates , this doesn't look too good as is. Adding a long-press expand action bar, like exists for comments, would be a much better solution. Or if not, just clicking into it is better.

The whole point of the list view mode is that it should be very compact.

dessalines (commented):
See comments above.

noloman:
> I agree with @jacobbates , this doesn't look too good as is. Adding a long-press expand action bar, like exists for comments, would be a much better solution. Or if not, just clicking into it is better.
> 
> The whole point of the list view mode is that it should be very compact.

what do you mean with `clicking into it`?

dessalines:
The list view mode doesn't show on the `PostActivity` page, which shows the full card. That card has options to bookmark.

noloman:
> The list view mode doesn't show on the `PostActivity` page, which shows the full card. That card has options to bookmark.

What about putting the save button to the right of the comments icon?

dessalines:
I'm not really a fan of that either. Long-hold to show the action bar is a better solution that keeps it compact.

Issue #553, opened 2023-06-12
Comment replies icon appears in profile


jordan1776:
** Jerboa Version **
0.0.33

**Describe the bug**
Comment replies icon (+N) appears when tapping user comments in profile.

**To Reproduce**

1. Go to 'profile'
2. Tap 'comments'
3. Tap a comment which has replies.

priyatanu:
Hi ,
I would like to work on this issue. Could you please describe more. as I new to the project. maybe with screenshot?

dessalines:
This is fixed.

PR #533, opened 2023-06-11
Added Italian localization

andreaippo:
Added Italian localization

PR #726, opened 2023-06-17
fix: Comment replies icon appears in profile #533

LufyCZ:
#553


dessalines (commented):
This needs a merge from main, but also I don't understand the point of this. We recently allowed replying to your own comment, so that should be okay.

dessalines:
Ah I see what you're saying, gotcha. The click in that case should really show the context (same for inbox)

app/src/main/java/com/jerboa/ui/components/comment/CommentNode.kt
```diff
@@ -273,10 +276,10 @@ fun LazyListScope.commentNodeItem(
                             myVote = instantScores.value.myVote,
                             isModerator = isModerator(commentView.creator, moderators),
                             onClick = {
-                                toggleExpanded(commentId)
+                                onHeaderClick(commentView)
```
dessalines, line 279: Thanks, these are more clear.

PR #727, opened 2023-06-17
concised crypto addresses

Jai707:


dessalines:
Not really necessary IMO.

PR #703, opened 2023-06-16
Render markdown in post comment preview

noloman:
Before:

Screenshot_20230615_123255_Jerboa (DEBUG)

After:

Screenshot_20230615_123203_Jerboa (DEBUG)

PR #734, opened 2023-06-18
markdown parse the 5 lines truncated body preview in post listing

vijaykramesh:
right now it doesn't markdown parse this 5 line body preview so links and such don't work properly:

this updates it to do a 5 line body preview (still with the bodyMedium text size) but markdown parse it validated looks correct now in both card & small card mode (doesn't impact list mode)


dessalines (commented):
Looks good, but one more thing:

Is it possible to have this not render images, by not adding coil, or removing it somehow for this view? Images in the post body, as well as people throwing markdown header / titles, are one reason I didn't render this before.

vijaykramesh:
Sure let me take a stab at stripping them out of the 5 line body preview before rendering it as md.

vijaykramesh:
ok updated @dessalines now the `CreateMarkdownView` has a `removeImages` mode that strips out any 

```
![optional_alt](https://image.com/photo.jpg)
```

sections and replaces them with ` img ` or ` img:<alt_text> ` if alt text is provided 

Should I also strip out `` style images as well? i know they won't get there by default but it might be possible to manually create one, not sure what lemmy does there (I know github markdown supports it)

ZJouba:
Possible duplicate of https://github.com/dessalines/jerboa/pull/703

app/src/main/java/com/jerboa/ui/components/common/MarkdownHelper.kt
```diff
@@ -165,25 +165,41 @@ object MarkdownHelper {
...
+        style: TextStyle? = null,
+        removeImages: Boolean = false,
     ) {
-        val style = MaterialTheme.typography.bodyLarge
         val defaultColor: Color = LocalContentColor.current.copy(alpha = LocalContentAlpha.current)
-
+        val typographyStyle = style ?: MaterialTheme.typography.bodyLarge
+        val preppedMarkdown = if (removeImages) {
+            // use regexp to parse out any markdown image tags
+            val regexp = Regex("!\\[(.*?)]\\(.*?\\)")
```
dessalines, line 176: This seems a bit hacky. I was thinking more along the lines of having a second Markdown renderer (mostly copied from the first), that doesn't load coil / use images.

Thoughts @twizmwazin ?

app/src/main/java/com/jerboa/ui/components/common/MarkdownHelper.kt
```diff
@@ -150,13 +151,33 @@ object MarkdownHelper {
...
+        noImageMarkwon = Markwon.builder(context)
+            // email urls interfere with lemmy links
+            .usePlugin(LinkifyPlugin.create(Linkify.WEB_URLS))
+            .usePlugin(LemmyLinkPlugin())
+            .usePlugin(StrikethroughPlugin.create())
+            .usePlugin(TablePlugin.create(context))
+            // use TableAwareLinkMovementMethod to handle clicks inside tables,
+            // wraps LinkMovementMethod internally
+            .usePlugin(MovementMethodPlugin.create(TableAwareMovementMethod.create()))
+            .usePlugin(HtmlPlugin.create()) // TODO - does this support <img ...> ? if so, we need to remove it
```
vijaykramesh, line 164: should this also be removed from the no image markwon ?

twizmwazin: Markwon's html plugin does support images, but rather than disable it entirely, we can disable just the image tag support. Have a look here: https://noties.io/Markwon/docs/v4/html/#predefined-taghandlers

app/src/main/java/com/jerboa/ui/components/common/MarkdownHelper.kt
```diff
@@ -165,28 +186,32 @@ object MarkdownHelper {
         color: Color = Color.Unspecified,
         onClick: (() -> Unit)? = null,
         onLongClick: (() -> Unit)? = null,
+        maxLines: Int? = null,
+        style: TextStyle? = null,
+        removeImages: Boolean = false,
```
twizmwazin, line 191: As a convention, I think having a negative parameter like this is a bit confusing, I'd rather do `enableImages`

app/src/main/java/com/jerboa/ui/components/common/MarkdownHelper.kt
twizmwazin: Yeah I agree, thanks for addressing it @vijaykramesh

app/src/main/java/com/jerboa/ui/components/common/MarkdownHelper.kt
vijaykramesh: updated

app/src/main/java/com/jerboa/ui/components/common/MarkdownHelper.kt
vijaykramesh: updated 

```kotlin
            .usePlugin(HtmlPlugin.create { plugin -> plugin.addHandler(TagHandlerNoOp.create("img")) })
```

app/src/main/java/com/jerboa/ui/components/common/MarkdownHelper.kt
```diff
@@ -150,13 +152,33 @@ object MarkdownHelper {
                 }
             })
             .build()
+
+        noImageMarkwon = Markwon.builder(context)
```
dessalines, line 156: These are mostly identical, so you could probably use a common builder. Then at the very end, create the two versions:

```
imageVersion = builder.usePlugin(coil...).build()
nonImageVersion = builder.build()
```

app/src/main/java/com/jerboa/ui/components/common/MarkdownHelper.kt
```diff
@@ -165,28 +187,32 @@ object MarkdownHelper {
...
         onClick: (() -> Unit)? = null,
         onLongClick: (() -> Unit)? = null,
+        maxLines: Int? = null,
+        style: TextStyle? = null,
+        includeImages: Boolean = true,
     ) {
-        val style = MaterialTheme.typography.bodyLarge
         val defaultColor: Color = LocalContentColor.current.copy(alpha = LocalContentAlpha.current)
-
+        val typographyStyle = style ?: MaterialTheme.typography.bodyLarge
```
dessalines, line 195: Considering this is a default, move this into the params instead.

vijaykramesh: updated

app/src/main/java/com/jerboa/ui/components/common/MarkdownHelper.kt
```diff
@@ -165,28 +176,31 @@ object MarkdownHelper {
...
-
         AndroidView(
             factory = { ctx ->
                 createTextView(
                     context = ctx,
                     color = color,
                     defaultColor = defaultColor,
                     fontSize = TextUnit.Unspecified,
-                    style = style,
+                    style = style!!,
```
dessalines, line 191: These are always dangerous, remove this one.

app/src/main/java/com/jerboa/ui/components/common/MarkdownHelper.kt
```diff
@@ -165,28 +176,31 @@ object MarkdownHelper {
         color: Color = Color.Unspecified,
         onClick: (() -> Unit)? = null,
         onLongClick: (() -> Unit)? = null,
+        maxLines: Int? = null,
+        style: TextStyle? = MaterialTheme.typography.bodyLarge,
```
dessalines, line 180: Since it has a non-null default, the optional isn't necessary.

dessalines (changes requested):

dessalines (approved):
LGTM, I'll let @twizmwazin take a look too.

app/src/main/java/com/jerboa/ui/components/common/MarkdownHelper.kt
```diff
@@ -141,14 +142,23 @@ object MarkdownHelper {
             // use TableAwareLinkMovementMethod to handle clicks inside tables,
```
MV-GH, line 142: This should only be part of the main markdown. I assume this would make it possible if a table with links is in the preview and you click on it it wont open the post but this link directly. Same goes for normal links. It should not allow you to open them.

twizmwazin (commented):
I think I agree with @MV-GH about link handling. Going with the idea in #710, we want to avoid too many small touch targets. I suppose this doesn't apply in the list mode because there is no preview there, but I still think it would be better UX to have the user always open posts instead of opening links.

dessalines:
@vijaykramesh any updates? Those changes should be quick to implement. If you're swamped, one of us could probably finish this up.

dessalines:
This has been merged now in #1002

vijaykramesh:
Ah sorry all I was out of town but see this got finished and merged, thanks!

PR #735, opened 2023-06-18
Fix bug with pictrs url query param rewriting

beatgammit:
Apparently I broke stuff by trying to be fancy. I'm not really a fan of parsing the URL like this though, but at least there are test cases for it now.


sockenklaus (approved):
This works for urls with query parameters and without query parameters.

dessalines (approved):

Issue #673, opened 2023-06-15
Add version incompability checks between Jerboa and Lemmy


MV-GH:
Currently Jerboa supports lemmy v17.3, soon lemmy will upgrade to v18 which contains breaking changes. If Jerboa tries to connect to server with a higher major version it should show a popup in the manner of "This version X.X is not supported yet". The connecting can either be when already logged in, when you try to login for an instance or the anon feed (which is currently lemmy.ml). It be great if you still could connect to versions which are supported.

For versions lower than 17.3 it should display that these are not supported. 

But I don't know the procedure for when Jerboa upgrades to v18. Will it still support previous versions (v17.3)? If so seemingly in the app or will the user have to download an older version of Jerboa?

PR #737, opened 2023-06-18
Check server version against app minimum api version

### LINKED ISSUE #701: Do a version check on login, to say whether the current version of jerboa supports your server.
IE have a popup error that says ("Version of lemmy unsupported by X instance. Ask your server administrators to upgrade lemmy.")

This is necessary, since logging into pre `0.18` versions will fail, due to breaking API changes.

beatgammit:
This will pop up the version error dialog every time the user launches the app.

Fixes #701 and #673.


app/src/main/res/values/strings.xml
```diff
@@ -138,6 +138,7 @@
     <string name="dialogs_top_year">Top Year</string>
     <string name="dialogs_old">Old</string>
     <string name="dialogs_unread">Unread</string>
+    <string name="dialogs_server_version_outdated">Server version (%1$s) is under the minimum (%2$s).\n\nPlease inform your administrator and login to another instance, or sign out and use the default instance.</string>
```
MV-GH, line 141: I would make a small change here:

"is under the *supported* minimum"

beatgammit:
This also implements logging in with older, "unsupported" servers. I didn't fully test everything, but the basics seem to work (logging in, making posts, responding to posts, etc).

app/src/main/res/values/strings.xml
beatgammit: So, apparently supporting older versions isn't that painful. So maybe we should fix the wording even more to let users know that most features should work, but that they should ask the admin of their instance to upgrade.

MV-GH: Wonderful, in that case it might be better to separate the supported versions (0.18), temporary supported versions (0.17.X) and not supported at all (0.16 and below) That would cover all  #673

beatgammit: I think it increases testing load too much per release. I don't think anyone is particularly interested in testing on older server versions, so merely stating that it's unsupported and some features may not work should be sufficient.

That said, if someone builds a mock "server" or something, we could at least write some unit tests. I might get around to that eventually since I'm interested in testing, but there's a lot of other missing functionality I'd also like to work on.

If we decide to officially "partially" support the previous release, we'll need to store the dismissal in the database. I intentionally didn't do that here to be just that much more annoying to get users to pressure their admins to upgrade.

I'm happy either way, I just don't want to set expectations we can't really meet.

MV-GH: I didn't mean to imply we should fully support 0.17.X or any future outdated version, Just that 16.x is not supported at all  thus should not be reported at all in the issues.

beatgammit: Ok, fair. I don't know what changed between 0.16 to 0.17, so I guess we could handle that as a special case.

I'm more interested in what we do going forward.

app/src/main/java/com/jerboa/MainActivity.kt
```diff
@@ -135,6 +145,27 @@ class MainActivity : ComponentActivity() {
 
                 ShowChangelog(appSettingsViewModel = appSettingsViewModel)
 
+                when (val siteRes = siteViewModel.siteRes) {
```
dessalines, line 148: This needs to be externalized, and also added to the login page.

app/src/main/java/com/jerboa/datatypes/types/Others.kt
```diff
@@ -20,64 +20,64 @@ enum class SortType {
     /**
      * Posts sorted by the most recent comment.
      */
-    @SerializedName("Active")
+    @SerializedName("Active", alternate = ["0"])
```
dessalines, line 23: Don't include this, we don't want to start doing the work of supporting multiple versions by hacks in a single codebase.

app/src/test/java/com/jerboa/UtilsKtTest.kt
```diff
@@ -147,4 +147,14 @@ class UtilsKtTest {
...
+
+    @Test
+    fun compareVersions() {
+        assertEquals(-1, compareVersions("0.0.1", "0.0.2"))
+        assertEquals(1, compareVersions("0.0.10", "0.0.2"))
+        assertEquals(1, compareVersions("0.1.10", "0.1.2"))
+        assertEquals(0, compareVersions("0.1.2", "0.1.2"))
+        assertEquals(-1, compareVersions("0.1.2-alpha1", "0.1.2-beta1"))
+        assertEquals(1, compareVersions("0.1.2-beta1", "0.1.2-alpha2"))
+    }
```
dessalines, line 159: Thx, nice job!

app/src/test/java/com/jerboa/datatypes/types/OthersTest.kt
```diff
@@ -0,0 +1,31 @@
...
+import com.jerboa.datatypes.types.SortType
+import com.jerboa.gson
+import junitparams.JUnitParamsRunner
+import junitparams.Parameters
+import org.junit.Assert.*
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(JUnitParamsRunner::class)
+class OthersTest {
```
dessalines, line 11: Remove this.

app/src/main/java/com/jerboa/MainActivity.kt
beatgammit: This will show two alerts when first launching the app, this one just before the change log.

Would you prefer the change log first? Or have it be one or the other?

app/src/main/java/com/jerboa/datatypes/types/Others.kt
dessalines: I know, but these are going to get out of hand *very quickly* with breaking changes, especially with lemmy API not even having a 1.0 release yet. So for now I'd rather not do hacks to support beta versions.

app/src/main/java/com/jerboa/ui/components/login/LoginViewModel.kt
```diff
@@ -93,32 +95,45 @@ class LoginViewModel : ViewModel() {
...
-                        name = luv.person.name,
-                        current = true,
-                        instance = instance,
-                        jwt = jwt,
-                        defaultListingType = luv.local_user.default_listing_type.ordinal,
-                        defaultSortType = luv.local_user.default_sort_type.ordinal,
-                    )
+                    val siteVersion = siteRes.data.version
+                    if (compareVersions(siteVersion, MINIMUM_API_VERSION) < 0) {
+                        Toast.makeText(ctx, "Server version too low.", Toast.LENGTH_SHORT).show()
```
dessalines, line 100: Needs a translation

dessalines (commented):
I tested, and it works great, thanks a ton for this. Just needs a translation string.

app/src/main/java/com/jerboa/ui/components/login/LoginViewModel.kt
beatgammit: Not sure how to do that since stringResource needs a Composable block. That's why I didn't use the AlertDialog here.

I can look into it again though, there ought to be a solution

dessalines: I think `ctx.getString(R.string...` is how you do it. 

https://stackoverflow.com/a/7454485

beatgammit: Thanks. It's fixed now.

dessalines (approved):
LGTM, thanks. We're now all set for `0.18.0` !

I'll do a jerboa release the same day as we do the lemmy release, so it should be on izzyondroid that day, and github releases immediately.

MV-GH:
@dessalines Is it possible to do a earlier release but not make available. So that it gets already reviewed. Atm it takes about a week to be actually available on the play store. And in that period users wont be able to use the app.

Issue #742, opened 2023-06-18
Render Markdown Tables


ElenaRgC:
- [x] Did you check to see if this issue already exists?
- [x] Is this only a single feature request? Do not put multiple feature requests in one issue.
- [ ] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/jerboa

** Describe the feature request below **

Jerboa can't render Markdown tables, here is a comparison with a web broser on the same phone.

![Screenshot_2023-06-18-16-16-41-637_com.jerboa.jpg](https://github.com/dessalines/jerboa/assets/95099947/e8066a94-1b55-4bec-91dc-a910e93b7dd8)

![Screenshot_2023-06-18-16-16-35-896_com.brave.browser.jpg](https://github.com/dessalines/jerboa/assets/95099947/bacc0934-7c0b-4205-bb06-1870de3b12e4)

As you can see, the titles and links are rendered properly but tables aren't.

MV-GH:
https://github.com/dessalines/jerboa/assets/67873169/dd0ec28c-137b-44ae-813d-63e74816fd13

That feature is already in main but the links didn't work in tables. So I'm making a pr that fixes those

PR #744, opened 2023-06-18
Fix links not working in markdown tables

MV-GH:
Fix links not working in markdown tables

https://github.com/dessalines/jerboa/assets/67873169/741b8137-9d2d-4b30-b9e8-328348d4bd83

see #742


dessalines (approved):
So many weird edge cases with markdown... oh well, at least markwon has the plugins to support it.

PR #746, opened 2023-06-18
Added blur for Android version older than 12

### LINKED ISSUE #733: Blur Effect doesn't work on old Android version
** Jerboa Version **
0.0.34

**Describe the bug**
In Android version lower than 12, the blur nsfw picture doesn't work at all. 

**To Reproduce**
0. Use Jerboa on Android version 9
1. Go to instance that has nsfw post
2. Scroll down to any post with nsfw picture 
3. See that it shows plain image instead of blurred image

Note that it doesn't happen on latest version of Android. Most likely caused by using blur from jetpack compose in here https://github.com/dessalines/jerboa/blob/145d04aaab7af3ef8a11528b4e261efdd8951f38/app/src/main/java/com/jerboa/ui/components/common/PictrsImage.kt#L103
This function only works on Android 12 and above based on the documentation https://developer.android.com/reference/kotlin/androidx/compose/ui/Modifier#(androidx.compose.ui.Modifier).blur(androidx.compose.ui.unit.Dp,androidx.compose.ui.unit.Dp,androidx.compose.ui.draw.BlurredEdgeTreatment)

sockenklaus:
Closes #733 

I added a check for the Android version in `app/src/main/java/com/jerboa/ui/components/common/PictrsImage.kt`

Android version 12 or newer still uses the Jetpack Compose function `.blur()`.
For older versions I added a BlurTransformation provided by https://github.com/T8RIN/BlurTransformation that is applied to the Coil ImageRequest.

Because I added a new utility class to the project i created a new package `com.jerboa.util` and moved the existing class `Util.kt` to this package. This lead to a lot of refactoring to meet the lint requirements. Hence the many changes.

The real changes are in:
`app/src/main/java/com/jerboa/ui/components/common/PictrsImage.kt`

and

`app/src/main/java/com/jerboa/util/BlurTransformation.kt`


dessalines (commented):
I'm guessing you accidentally added a lot more than was necessary for some reason, or github messed up somehow, due to all the conflicts. You might need to start from scratch from main, and copy your new file over.

sockenklaus:
> I'm guessing you accidentally added a lot more than was necessary for some reason, or github messed up somehow, due to all the conflicts. You might need to start from scratch from main, and copy your new file over.

Ugh honestly I'm 99 % sure i made a stupid decision moving and renaming your `Util.kt` class grouping it with my newly added `BlurTransformation.kt` in the package `com.jerboa.util`... I thought grouping together utility classes would clean up the project structure but it got ktlint all upset because the moving and renaming messed up the lexicographical import-order....

I will revert this refactoring and update the PR accordingly this evening.

dessalines:
Sounds good, thx! There's no rush on it, I usually do frequent releases, so this will get in at some point.

Issue #542, opened 2023-06-11
Quick comment navigation


XLRUUU:
- [x] Did you check to see if this issue already exists?
- [x] Is this only a single feature request? Do not put multiple feature requests in one issue.
- [ ] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/jerboa

** Describe the feature request below **

It would be nice to have two buttons to quickly navigate between next and previous top level comment like it is possible on Boost.

DanielVolz:
Navigation through top comments with volume buttons would be great.

V13Axel:
Came here to suggest the same thing; I keep hitting the volume keys expecting it to jump between top level comments like Sync did.

Daniel15:
Relay for Reddit also used the volume buttons for this purpose. It's a great feature.

ToNIX-:
Please, implement this. Virtual buttons to navigate through the top comments is crucial to have. Thank you!

AaronMaria:
Joey has this with an additional button inbetween to select what the buttons should move to. So for example you can choose "new" instead of "top-level comments" to jump between new comments. I would love this kind of functionality in Jerboa.

ToNIX-:
Thank you for implementing this quickly! Another great addition would be to move to the top (the post itself) when you long press the "up" button.

tomewarren:
Can we also have the volume button set to scroll to the next card? On relay we could configure the button actions

dessalines:
Open these as separate individual issues.

PR #749, opened 2023-06-19
Quick comment navigation

yate:
This adds two buttons to the bottom of the Post activity for quick parent comment navigation (#542), as well as a Look and Feel setting to toggle them (enabled by default).

It also allows for parent comment navigation with volume buttons. This setting is disabled by default.

[device-2023-06-18-230927.webm](https://github.com/dessalines/jerboa/assets/726758/0ed61da6-c799-44e7-8940-e8b2505dbc98)


app/src/main/java/com/jerboa/ui/components/post/PostActivity.kt
```diff
@@ -135,8 +161,55 @@ fun PostActivity(
...
+                    containerColor = MaterialTheme.colorScheme.background.copy(alpha = .75f),
+                    modifier = Modifier.height(50.dp),
+                    content = {
+                        IconButton(modifier = Modifier.weight(.5f), onClick = {
+                            scrollToPreviousParentComment(scope, parentListStateIndexes, listState)
+                        }) {
+                            Icon(
+                                modifier = Modifier.scale(1.5f),
+                                imageVector = Icons.Filled.KeyboardArrowUp,
+                                contentDescription = "Up",
```
dessalines, line 197: Use i18n strings, lots of examples in this file.

app/src/main/java/com/jerboa/ui/components/post/PostActivity.kt
```diff
@@ -113,9 +135,13 @@ fun PostActivity(
     val unExpandedComments = remember { mutableStateListOf<Int>() }
     val commentsWithToggledActionBar = remember { mutableStateListOf<Int>() }
     var showSortOptions by remember { mutableStateOf(false) }
+    val focusRequester = remember { FocusRequester() }
 
     val listState = rememberLazyListState()
+    var lazyListIndexTracker: Int
```
dessalines, line 141: This should probably be in the remembered state as well.

dessalines (commented):
Nice job, everything works well.

My gut feeling was initially that maybe the index tracker should exist on `CommentNodes` rather than `Post`, but considering it should only ever be seen from the post page, and not on other comment lists, its probably better this way.

twizmwazin:
> Solve conflicts, then I'll check this out.
> 
> Random thought, but do you think in addition to volume button, a floating action button would work well too?
> 
> cc @twizmwazin

A floating action button would be more material-y, but I don't think that's necessarily what users prefer. This feature is an example of porting over a feature users have grown used to using in other apps, and I think in those cases it's important to replicate the UI that they want above all else. If someone wants it we can always add it in the future as an extra option.

app/src/main/java/com/jerboa/ui/components/post/PostActivity.kt
yate: The problem I had with this is that this counter [needed to be reset](https://github.com/dessalines/jerboa/pull/749/files/573a154d6cb52b624803dc5530a91501d168576a#diff-9958f04ae114c8444c9e781405889c7fab0b5465ca616011fcfb047a243c356bR246) every time the comments were composed, and the comments are getting recomposed all the time when scrolling. Without resetting it there, the counter would just keep rising. I actually needed to clear `parentListStateIndexes` for the same reason, it would just keep adding the parent indexes forever while scrolling.

So if I define it like `var lazyListIndexTracker = remember { 2 }`, I get a warning `Variable 'lazyListIndexTracker' initializer is redundant`

I'm very new to Kotlin and Jetpack Compose, so I might be missing a better way to do this

app/src/main/java/com/jerboa/ui/components/post/PostActivity.kt
```diff
@@ -135,8 +161,55 @@ fun PostActivity(
...
+                        scrollToNextParentComment(scope, parentListStateIndexes, listState)
+                    }
+                    true
+                } else {
+                    false
+                }
+            },
+        bottomBar = {
+            if (showParentCommentNavigationButtons) {
+                BottomAppBar(
```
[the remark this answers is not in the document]
dessalines: Sure, up to you.

app/src/main/java/com/jerboa/ui/components/post/PostActivity.kt
dessalines: Ah, if it needs to be reset every time then no need.

yate:
I'm noticing a bug where enabling volume button navigation breaks swiping to go back to the previous page, so I need to look into that

yate:
@dessalines Ready for another look, thank you

dessalines (approved):
Looks good, thanks!

PR #750, opened 2023-06-19
More Themes (Dark + Light)

bynatejones:
**Created a "Grey" option for Theme Color. Modified colors for both Dark and Light Theme.** 

<hr/>

<hr/>

**Created a "Purple" option for Theme Color. Modified colors for both Dark and Light Theme.** 

<hr/>

<hr/>

**Created a "Woodland" option for Theme Color. Modified colors for both Dark and Light Theme.** 

<hr/>

<hr/>

**Created a "Beach" option for Theme Color. Modified colors for both Dark and Light Theme.** 

<hr/>

<hr/>

**Created a "Crimson" option for Theme Color. Modified colors for both Dark and Light Theme.** 

<hr/>

<hr/>

The one thing that probably still needs a change is the fact that the Dialogs and the NavBar both use the .surface color, so in order to get the NavBar to be the same color as the background, the Dialogs must also be this color and can't be given a slight accent to separate it from the background. Example:

<hr/>

**Two solutions I can think of:**

1. Separate the NavBar and Dialogs into different color values.
2. Add a dark/light translucent background behind the dialogs to separate them from the background no matter what color they are. See Boost for this same behavior.

I think it might also be smart to separate the .primary, .secondary, and .tertiary color as their own user settings, so that themes don't have to be made for each possible primary color that a user might want to use. For instance, for the Grey (Dark) theme, I've made the .primary color just ever so slightly Blue to separate it more from the other UI on the home feed (since most of the other text is already grey). However, it would be nice if a user could change that to be green, pink, purple, etc etc (custom color picker, not hard coded colors) while still being able to use my general color scheme.

**I can try my hand at any of the above solutions, just let me know what direction if any you'd like to go with it.**

<hr/>

Also, question. I wasn't able to see where the URL color in comments is found in the code, since I'd like to be able to theme that too (and ultimately should probably have its own user setting override as well). Right now it seems to be hardcoded as this green color:

<hr/>

Let me know if you'd like me to make any changes. I plan on making more themes.


dessalines:
> The one thing that probably still needs a change is the fact that the Dialogs and the NavBar both use the .surface color, so in order to get the NavBar to be the same color as the background, the Dialogs must also be this color and can't be given a slight accent to separate it from the background. Example:

In that case, we should just find the correct Material3 theme variable that works best, for both. I'd rather not add custom theme hacks, or additional settings on top of M3, but just use the existing variables, to look good with all themes.

app/src/main/java/com/jerboa/Utils.kt
```diff
@@ -832,9 +832,14 @@ enum class ThemeMode(val mode: Int) {
 
 enum class ThemeColor {
     Dynamic,
+    Beach,
+    Blue,
```
dessalines, line 836: Needs a merge from main, and you'll see that this enum has i18n strings now.

dessalines (commented):
Thanks for these, they all look really good.

app/src/main/java/com/jerboa/Utils.kt
dessalines: Could you add i18n strings to this, just liken `ThemeMode` has above it?

bynatejones: Okay, I've modified them to use i18n strings and added the English translations in the strings.xml file

dessalines (approved):
LGTM, thanks!

PR #751, opened 2023-06-19
fix(thumbnails): prevent incorrect formatting on URLs without query parameters

SphericalKat:
This fixes an issue where if the URL did not previously contain query parameters, it would incorrectly format the URL without a trailing `?`.
Before:
image

After:
image


dessalines:
Has already been fixed in main as far as I can tell. @beatgammit is also already on this one.

PR #763, opened 2023-06-19
Add CI dependency caching in between steps

dessalines:


MV-GH:
build_project should be before run_tests

dessalines:
Thx, I'll give that a shot.

PR #767, opened 2023-06-20
Use overloaded AndroidView in MarkdownHelper

### LINKED ISSUE #730: Scrolling up past images in the comments jerks you back down
**Jerboa Version**
0.0.34

**Describe the bug**
When trying to scroll up past images in the comments, your scrolling position is reset to below the image, making the scrolling feel very jerky and sometimes impossible to scroll up past the image again.

**To Reproduce**
1. Open a post's comments
2. Scroll down past an image
3. Try to scroll back up past the image
4. When you get to the image, your scrolling position is reset to some point below the image

https://github.com/dessalines/jerboa/assets/71697204/10d146e5-e0e0-4b32-91e9-8dea47f30cab

yate:
From what I've tested this mostly fixes #730. There's still a bit of a jerk upwards as the images loads in so it's not perfect, but it prevents the endless scroll downwards when hitting an image as shown in the issue.

I'm not entirely sure if anything needs to go in the `onReset` method, but it's noted in the Jetpack Compose documentation to use [this overload with LazyColumns ](https://developer.android.com/jetpack/compose/migrate/interoperability-apis/views-in-compose#androidview_in_lazy_lists) because the view can be reused during recomposition, so it may help with performance as well.

[device-2023-06-19-225611.webm](https://github.com/dessalines/jerboa/assets/726758/5b60d79b-e2da-4845-8bc3-654cff5bc6b8)


dessalines:
The example you linked shows that reset should include `view.clear()`

yate:
I looked more into that example, and that `clear()` [doesn't seem to do anything either](https://github.com/android/snippets/blob/1d381ff086395f4c3bd9bb8b5085ab67bb3e9a0f/compose/snippets/src/main/java/com/example/compose/snippets/interop/InteroperabilityAPIsSnippets.kt#L233)

The documentation says the callback "may be used to reset any transient View state like animations or user input.". I don't think that applies here?

ZJouba:
Okay, after taking a look in LayoutInspector, I believe this is caused by the View being reused and the image being discarded.

When the comment is inside the viewport everything looks fine, we can see the markdown helper in the AndroidView
Screenshot 2023-06-20 180827

However, when we scroll the image out of the viewport we see the AndroidView is empty
Screenshot 2023-06-20 181110

Then, I believe when it is scrolled back into view, the image gets reloaded and something forces the scroll to focus on the image. What's weird to me is that the height of the AndroidView remains constant at 504dp so why does the viewport suddenly jump to include the whole image? It's not a matter of layout shift.

dessalines:
I'm sure its lazylist weirdness, might be worth trying to mess around with the markwon pictrs coil image sizes.

ZJouba:
Tsk, this is what we need but it's not available for LazyColumn yet https://issuetracker.google.com/issues/172029355

dessalines (approved):
Not sure that this onReset will do anything, but I resolved the conflicts.

Issue #775, opened 2023-06-20
Quotation mark is not rendered in communities' names


Snow4DV:
**Jerboa Version**
0.0.33/0.0.34
**Describe the bug**
Quotation mark is not rendered in communities' names but does render on communities' page 
**To Reproduce**

1. Open app
2. Go to the page of community with quotation mark in its name  
3. Open any post
![Screenshot_2023-06-20-19-09-45-47_a10f969cf1c4adf2c503042b459b8476](https://github.com/dessalines/jerboa/assets/32912696/93dd9dfe-c335-4338-b7b2-85f6d7b889e9)
![Screenshot_2023-06-20-19-07-56-67_a10f969cf1c4adf2c503042b459b8476](https://github.com/dessalines/jerboa/assets/32912696/38014989-a145-46f1-b146-65bfc68fe0f9)

MV-GH:
```kt
    Text(
        text = communityNameShown(community),
        style = style,
        color = color,
        modifier = modifier,
        overflow = TextOverflow.Clip,
        maxLines = 1,
    )
    ``` It gets clipped
    
    
![image](https://github.com/dessalines/jerboa/assets/67873169/a98c96aa-f9b7-45aa-84c0-ed89f23ebc76)
Here shown working
![image](https://github.com/dessalines/jerboa/assets/67873169/ae4fb799-d014-471a-a8ec-8815ca3dc4f6)

PR #778, opened 2023-06-20
Add zero-width space to person/community's shown name

Snow4DV:
Addresses issue #775 - last word of community's name is cut if instance's URL is too long. I added zero-width space so it would cut name of instance only if it doesn't fit.


MV-GH:
image
This works

MV-GH (approved):

dessalines (commented):
Not a fan of this hack. Seems like something should be done for the post listing title, not here.

MV-GH:
Well there you could change the Overflow to show the dots instead of cutting it off if that change in behavior is wanted. But doing the above change there is much harder.

Snow4DV:
I think you're right - it's better just to change overflow in CommunityName to "overflow = TextOverflow.Ellipsis". In that case it will be eaiser to distinguish local communities from remote ones if such thing happens

PR #794, opened 2023-06-21
Russian translation & Fixes of non-localized strings

Snow4DV:
Translated app to russian and fixed some unlocalized string here and there


app/src/main/java/com/jerboa/ui/components/person/PersonProfileActivity.kt
```diff
@@ -103,7 +103,7 @@ fun PersonProfileActivity(
                     PersonProfileHeader(
                         scrollBehavior = scrollBehavior,
                         personName = if (savedMode) {
-                            "Saved"
+                            ctx.getString(R.string.bookmarks_activity_saved)
```
dessalines, line 106: Good catch, thanks.

app/src/main/res/values/strings.xml
```diff
@@ -310,4 +310,8 @@
     <string name="saved_image">Saved image</string>
     <string name="saving_image">Saving image…</string>
     <string name="permission_denied">Permission denied</string>
+    <string name="inbox_activity_replies">Replies</string>
+    <string name="inbox_activity_mentions">Mentions</string>
+    <string name="inbox_activity_messages">Messages</string>
+    <string name="bookmarks_activity_saved">Saved</string>
```
MV-GH, line 316: There is already two Saved string resources, maybe use an existing? Or is it not applicable in this context?

dessalines (commented):
Thanks, I'm sure there's more that we're missing too, but we can find them later.

Run `./gradlew formatKotlin` to pass lint.

app/src/main/res/values/strings.xml
Snow4DV: Probably. It is applicable in this context but there already were duplicating string resources previously for different contexts so i thought it would be better to separate them to leave some flexibility.

dessalines (approved):

PR #800, opened 2023-06-21
Add dynamic splash screen colors

ddmetz:
I updated android:windowBackground to use the system theme colors on the splash screen. Devices below Android 12 will use `#FFFFFF` and `#121212` for light and dark theme respectively.

Should prevent dark mode users from getting hit with a flashbang when opening the app. :)

dark-small light-small


dessalines (approved):
Sweet, thanks!

PR #802, opened 2023-06-21
Fix links not working when exclamation mark before it

### LINKED ISSUE #795: !community@instance is not clickable if has a exclamation mark before it.
**Jerboa Version**
latest from main

**Describe the bug**
A community@instance is not clickable if it has a exclamation mark before it. Instead it will just collapse the comment.

**To Reproduce**
Try to click on it

go to this post https://lemmy.dbzer0.com/post/81613
try to click on !piracy... link

It does get properly turned into a visible @ but it is not clickable while lemmy-ui doesnt even turn it into a @.

How it looks on lemmy-ui
image

How it looks on jerboa
image

This post https://lemmy.dbzer0.com/post/81613

Demo

https://github.com/dessalines/jerboa/assets/67873169/1eb15add-9d39-46d7-bf64-2a388d9407aa

MV-GH:
It didnt work in some cases because it included the space, in the current link matcher.
image

`(?:^|\\s)!($communityPatternFragment)(?:@($instancePatternFragment))?\\b` 
This pattern used a non capturing group, but this is still included in the matched string thus it added the space, instead I used a lookbehind since these aren't included in the matched string. Also I updated the tests to include this edge case.

Fixes #795


dessalines (approved):
LGTM, thanks!

PR #803, opened 2023-06-21
Better workaround for incorrectly implemented languages in prettytime

Snow4DV:
In the issue #539 it is mentioned that Polish, Russian, Ukrainian and probably some other languages don't have "formatDuration" implemented correctly.

After digging into it and looking how it is used in other applications i found that these languages' implementation works correctly if you use 'format" method that displays "8 days ago" instead of "8 days" so it's probably better workaround at the moment until it is fixed.


Snow4DV:
Screenshot_2023-06-21-23-40-03-24_c9298fda48802cf3e2d17cb10a48759f
Tested also with polish and Ukrainian - works flawlessly 


Fix shouldn't be that hard but it wil require fixing for each language separately so can't be done quick

MV-GH:
Does that mean it will say now `5 hours ago` instead of `5h`, Using more screen real estate? or will it only do that for those languages? It doesn't increase UX.

Snow4DV:
> Does that mean it will say now `5 hours ago` instead of `5h`, Using more screen real estate for those languages? It doesn't increase UX. Since i don't know Russian would you say this is an improvement over the shorthand English notation as fallback?

You're right - probably it won't. In most languages (at least in Ukrainian, Russian and polish) just leaving only first letter of first word is fine (cutting 4 дней назад to 4д) but looks too hacky to me - better leaving it as is

PR #805, opened 2023-06-21
Fix: Add Account fails when full server uri is used

### LINKED ISSUE #739: Add Account fails when full URI of server instance used
**Jerboa Version**
0.0.33

**Describe the bug**
When joining Lemmy for the first time, it is natural to:
1. sign up for an instance using a web browser
2. copy/paste the domain from browser into the app

But if you have `https://` at the beginning of the instance name, Jerboa errors that it is not a Lemmy instance.

**To Reproduce**
- Sign up in the browser to a community (eg. `https://lemmy.world`)
- copy/paste the server domain into Jerboa (eg. `https://lemmy.world`)
- Fails with `https://lemmy.world is not a Lemmy instance`
- remove the `https://`, and it logs in successfully

**Suggestions**
1. Warn the user if a protocol is specified in the instance name `http://`, `https://`, etc
2. Strip the protocol, and attempt to connect

sockenklaus:
Implemented a function that extracts the host part from a URL,  removes the protocol and anything that may come after a :, a / or a  ?.

Closes #739


app/src/main/java/com/jerboa/Utils.kt
```diff
@@ -1231,6 +1231,18 @@ fun scrollToPreviousParentComment(
     }
 }
 
+/**
```
dessalines, line 1234: Someone should really make a library for this, if its functionality isn't already included in URL.

dessalines (approved):
LGTM, although I really wish we didn't have to use regex for something like this, and could use a library.

app/src/main/java/com/jerboa/Utils.kt
MV-GH: But this is the java ecosystem, of course something like this exists already.

```kt
Patterns.WEB_URL.matcher(url).matches() // To check if a valid URL
println(URL("https://lemmy.ml").host) // lemmy.ml
```

twizmwazin (changes requested):
Let's use @MV-GH's solution

sockenklaus:
> Let's use @MV-GH's solution

But regex is fun! 😢

No, seriously: @MV-GH's solution is clearly more elegant. I will adjust the PR later the day.

twizmwazin:
In some cases regex is being used because the links are metaverse links that aren't actually valid URLs, or at least I don't think they are valid. `@user@instance`, `!community@instance` mainly. If it isn't one of those I agree using regex is sub-optimal and fixing those would be welcome.

app/src/main/java/com/jerboa/Utils.kt
dessalines: Perfect

sockenklaus:
I reworked the function but didn't ended up using `Patterns.WEB_URL` to match against because for some reason it doesn't handle hosts without domain-part (i.e. `localhost`). I think local hosts are useful for local testing.

Also i added [Roboelectric](https://robolectric.org/) to the project because i didn't see another way to test my function without mocking. But I'm fairly new to using and writing unit tests to correct me if i was wrong using that package.

MV-GH:
Could you not use `@RunWith(JUnitParamsRunner::class)` ?
see `com/jerboa/ui/components/common/LemmyLinkPluginTest.kt`

MV-GH (approved):

dessalines (approved):

twizmwazin (approved):

PR #816, opened 2023-06-22
Fix thumbnail links being treated as images. Fixes #815

### LINKED ISSUE #815: Links are being treated as images
**Jerboa Version**

Main branch

**Describe the bug**

Clicking on a link tries to load an image, rather than going to the link.

**To Reproduce**

See above

**In the case of a crash**

dessalines:
Somewhere along the line, clicking on thumbnail links got removed, and every thumbnail was treated as an image, even if it was a link post.

This fixes that by altering the thumbnail tile to check to see if the url is an image.

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
+++ b/app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
@@ -77,6 +77,7 @@ import com.jerboa.hostName
 import com.jerboa.isImage
 import com.jerboa.isSameInstance
 import com.jerboa.nsfwCheck
+import com.jerboa.openLink
 import com.jerboa.ui.components.common.ActionBarButton
 import com.jerboa.ui.components.common.CircularIcon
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
```diff
@@ -1095,9 +1141,23 @@ private fun ThumbnailTile(
...
+            .clickable {
+                if (isImage(url)) {
+                    showImageDialog = true
+                } else {
+                    openLink(
+                        url = url,
+                        navController = navController,
+                        useCustomTab = useCustomTabs,
+                        usePrivateTab = usePrivateTabs,
+                    )
```
dessalines, line 1158: Important code here.

beatgammit (approved):
Looks good to me.

I wonder if we can turn on a lint for variable mutability.

PR #785, opened 2023-06-21
Enable swiping on comments to go back to posts

CharlieGitDB:
The Reddit clients I have always used on Android were Slide, Sync, Infinity, all of these by default support swiping left to right on a Post to go backwards.  I have made code changes to make this possible on the comments to go back to the post view.

Closes #431

(Sorry for the laggy recording)
> [swipebacklemmypr.webm](https://github.com/dessalines/jerboa/assets/17328118/13ea83d0-c6c8-4d49-a336-63c06d64b472)

PR #817, opened 2023-06-22
Nav scoped view models

### LINKED ISSUE #402: Profile-Screen shows stale data after re-entry
Expected behaviour is to not see any data from previous Profile Screen-opens when opening the screen for a different user.

### LINKED ISSUE #704: Comments from the previously viewed post linger until comments from the current post load
** Jerboa Version **
0.0.33

**Describe the bug**
While the comments on a post are still loading, Jerboa shows the comments from the last-viewed post as if they were on the current post. Additionally, if the comments from the current post fail to load, Jerboa seems to happily display the wrong comment section without showing an error.

This doesn't necessarily make the app less usable, but it's really confusing for new users and might lead to people with slow or patchy internet commenting on the wrong post, so adding some sort of "Comments Loading" bit should be a somewhat high priority.

**To Reproduce**

1. Open a post and wait for the comments to load
2. Close the post, open another one, and quickly scroll down to the comments
3. While the loading bar is still present (or if the new comments fail to load), Jerboa displays the comments from the previous post

### LINKED ISSUE #774: Navigating back to a single-community's post list from comments scrolls back to the first page of posts and unloads other pages.
**Jerboa Version**
0.0.34

**Describe the bug**
If you view the comments on a post beyond the first page of a single community, the post list will scroll back to the first page and unload all other pages.

**To Reproduce**

1. Go to a community (this behavior does not seem to happen on the the local/all/subscribed pages)
2. Scroll down to a post on the second page (or beyond)
3. Open the comments
4. Navigate back to the list of posts.
5. The post list will automatically refresh and scroll back to the top, unloading all but the first page

Didn't seem like anyone reported this yet, but sorry if my searching skills were off. Let me know if I can be of any assistance/provide more details!

### LINKED ISSUE #835: Lazy list state is not stored somewhere if composable is destroyed
**Jerboa Version**
Latest main

**Describe the bug**
Feed that was previously scrolled for a long time ends up returning to the beginning if composable is destroyed (example: if you switch to your inbox or go to someone's profile)

**To Reproduce**

1. Open app
2. Switch to inbox
3. Switch back to home

LazyListState should be stored somewhere to outlast the composable if it gets destroyed

nahwneeth:
Since PR #765 was too big, splitting it up into multiple PR to make it easier to review.

# Scoped view models
- All the view models that were declared in the activity are now scoped to the routes in which they are used. View models used by most of the routes (like SiteViewModel, AccountViewModel & AccountSettingsViewModel) are left untouched.
- This prevents sharing view models that lead to stale data & reloading the data when navigating back.
- Making the LaunchedEffect run only once per route. This helps ensure that the content isn't reset with new data when moving back from one route to another. Things like scroll position & filters are also retained.
- Using extension function on navController to navigate in a type safe manner.

Closes #704
Closes #402 
Closes #774
Closes #835 

---

The following is the existing design where the view models reside in the `MainActivity` and are passed to all the activities:
Untitled Diagram-Existing drawio

---

The following is the design implemented by the PR:
Untitled Diagram-PR drawio

- The channel is also a ViewModel that is scoped to the NavHost's nav graph which has `route = Route.Graph.ROOT`. This is created on the fly whenever you see something like `transferPostEditDepsViaRoot = navController.rootChannel<PostEditDeps>()`.
- Instead of initializing a global viewmodel and then navigating to it's activity. We store the dependencies in the global channel and then initialize the view model in it's respective activity.
- On success the updated data is put to the parent's back stack entry with a specific key (`PostEditReturn.POST_VIEW`) and then navigate back. On navigating back, the parent will recompose and we check if a value has been set in `PostEditReturn.POST_VEW` and update the post accordingly.

Other Notes:
- I've provided extensions to navigate to each activity. For eg. to navigate to PostEditActivity on has to call `navController.toPostEdit(Channel<PostView>, PostView)`. This way it's a compile time error to forget providing the dependencies necessary to initialize an activity.
- These deps are converted to JSON and rememberSaved in the respective route (when `navController.takeDepsFromRoot<PostView>()` is called). The updated data on the back stack entry's `SavedStateHandle` is also in JSON. Both survive process death.
- To reuse code, I've introduced abstractions like `InitializeRoute` and `ConsumeResult` which are self-explanatory.


dessalines:
It might be a while till I can go over bigger refactors like this.

twizmwazin (approved):
Would love to get this merged once rebased

nahwneeth:
Note regarding changes made while resolving merge conflicts:

- The user preferred sort & listing defaults were housed in HomeViewModel (PR #854). This has been moved to SiteViewModel. Reason: The CommunityActivity depended on HomeViewModel for the defaults which hinders scoping. Since, both already depended on SiteViewModel it made more sense to keep them there. Also, changing the filters in HomeActivity doesn't overwrite the user defaults. Would appreciate if @beatgammit reviewed these changes and pointed out if I missed out or overwrote any fixes.
- The PostCreate & PostEdit were consolidated (PR #860). I've tested the post edit and create functionality after rebasing and they seem to be working. Would appreciate if @sockenklaus reviewed these changes gave a thumbs up.

app/src/main/java/com/jerboa/MainActivity.kt
```diff
@@ -94,25 +79,11 @@ class JerboaApplication : Application() {
     val appSettingsRepository by lazy { AppSettingsRepository(database.appSettingsDao()) }
 }
 
-class MainActivity : AppCompatActivity() {
```
twizmwazin, line 97: This needs to stay as AppCompatActivity for the in-app language picker

app/src/main/java/com/jerboa/ui/components/post/create/CreatePost.kt
twizmwazin, line 1: This file is empty

app/src/main/java/com/jerboa/ui/components/home/HomeActivity.kt
```diff
@@ -46,39 +46,54 @@ import com.jerboa.datatypes.types.BlockPerson
...
-    homeViewModel: HomeViewModel,
     accountViewModel: AccountViewModel,
     siteViewModel: SiteViewModel,
-    postEditViewModel: PostEditViewModel,
     appSettingsViewModel: AppSettingsViewModel,
     showVotingArrowsInListView: Boolean,
     useCustomTabs: Boolean,
     usePrivateTabs: Boolean,
     drawerState: DrawerState,
+    homeViewModel: HomeViewModel,
```
twizmwazin, line 93: I think there's sorta an unwritten convention where view models are kept towards the top of the argument lists for composable.

app/src/main/java/com/jerboa/MainActivity.kt
```diff
@@ -443,112 +380,68 @@ class MainActivity : AppCompatActivity() {
...
+                                type = Route.PostArgs.ID_TYPE
                             },
                         ),
                     ) {
-                        LaunchedEffect(Unit) {
-                            val postId = it.arguments?.getInt("id")!!
-                            postViewModel.initialize(id = Either.Left(postId))
-                            postViewModel.getData(account)
-                        }
-                        SwipeToNavigateBack(navController = navController) {
```
twizmwazin, line 506: #785 added a gesture to swipe to close comments. It would be great if that could be maintained.

twizmwazin (changes requested):

sockenklaus:
> Would appreciate if @sockenklaus reviewed these changes gave a thumbs up.

I'll gladly do and give it a look tonight.

twizmwazin (approved):

nahwneeth:
I've opened a new issue for the filters #886. For now, I've pushed a commit to ensure that the filters (in home & community activity) are read form the account variable. Ig we can continue with merging this PR and take up the filters thing separately. Any other issues with this branch, do let me know.

app/src/main/java/com/jerboa/ui/components/post/edit/PostEditActivity.kt
```diff
@@ -164,11 +161,11 @@ fun onSubmitClick(
                 auth = acct.jwt,
                 nsfw = isNsfw,
             ),
-            navController = navController,
-            postViewModel = postViewModel,
-            personProfileViewModel = personProfileViewModel,
-            communityViewModel = communityViewModel,
-            homeViewModel = homeViewModel,
-        )
+        ) { postView ->
```
sockenklaus, line 164: Again: Maybe use named parameter for onSuccess to increase readability?

sockenklaus (commented):
I only reviewed the files regarding post creation and editing because I honestly can't wrapt around my head about that amount of abstraction you added without investing a serious amount of time which I just don't have atm. (Hence my rather small PRs)
What I reviewed looked correct though and post creation and editing acted as expected.

app/src/main/java/com/jerboa/ui/components/post/edit/PostEditActivity.kt
```diff
@@ -13,51 +14,56 @@ import androidx.compose.runtime.setValue
...
 
-    var name by rememberSaveable { mutableStateOf(pv?.post?.name.orEmpty()) }
-    var url by rememberSaveable { mutableStateOf(pv?.post?.url.orEmpty()) }
-    var isNsfw by rememberSaveable { mutableStateOf(pv?.post?.nsfw ?: false) }
+    val postEditViewModel: PostEditViewModel = viewModel()
+    InitializeRoute(postEditViewModel) {
+        postEditViewModel.initialize(postView)
+    }
+
+    var name by rememberSaveable { mutableStateOf(postView.post.name) }
```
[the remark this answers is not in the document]
nahwneeth: Now that the view models are scoped and declarative navigation is being used, the `PostEditActivity` can only be opened with a proper `PostView`. Perviously, the `pv` was taken from `PostViewModel` and no guarantees could be made that as postView was actually loaded and that's why all those null checks were required.

nahwneeth:
> What I reviewed looked correct though and post creation and editing acted as expected.
Thanks a lot for taking the time! Will make the changes you requested. 

I think that's the hard part, reviewing changes. I'll try to document these changes better.

sockenklaus:
> I think that's the hard part, reviewing changes. I'll try to document these changes better.

Thank you for further describing the changes you've made. I really appreciate your effort to make these changes as understandable as possible. 🙂

MV-GH:
I have question regarding `initializeRoute`, why did you take this approach over just having the necessary fields passed to the constructor of the ViewModel. Passing it in the constructor is much more typesafe as we now don't have to set each field as nullable that we have to check constantly

nahwneeth:
I was trying to keep the changes small in this PR since it was getting
harder to review. Other than the `initialized` field the view models don't
have any other changes.

Constructor initialization of view models is safer and we should try to do
that.

MV-GH:
Thanks for your quick response, I'll make a issue for that and work on refactoring that

MV-GH:
I do have another question, 

For communityViewmodel you did this

```kt

val communityViewModel: CommunityViewModel = viewModel(
   remember(it) { navController.getBackStackEntry(Route.Graph.COMMUNITY) },
)

```

I am not sure what this does or why it exists? 

Why can it not be embedded in CommunityAcitivity?

nahwneeth:
navController.getBackStackEntry(Route.Graph.COMMUNITY)

This scopes the view model to the route with name Route.Graph.COMMUNITY

The community activity and community sidebar need to utilise the same view
model. I've made both these activities as subroutes of
Route.Graph.COMMUNITY.

So when creating a view model we don't scope it to the route but instead to
it's parent. When you navigate to the sidebar, it ends up using the view
model scoped to the parent.

So both community activity and community sidebar activity aren't embedded.

nahwneeth:
Interesting. Give it a try. Make sure to put the addReturn lines within the
NavController.toCommunitySidebar() extension method. It should work and the
dependency is also explicit ig.

MV-GH:
Alright got it working, had to make a reverse addReturn/consumeReturn

Issue #818, opened 2023-06-22
(German Translation) Ugly line break in bottom Navbar 


sockenklaus:
**Jerboa Version**
main branch

**Describe the bug**
The words "Posteingang" and "Lesezeichen" have ugly line breaks. 

This happens on my OnePlus 5t.

* Size: 6.01 inch
* Resolution: 2160 X 1080.

See:
![Screenshot_20230622-175700__01](https://github.com/dessalines/jerboa/assets/870042/ae43bdd1-a35c-4bef-a23c-9c137846f873)

I'll try to look into this later this evening. Maybe it's enough to adjust the padding of those buttons.

Is there any lower resolution limit Jerboas GUI has to work flawlessly?

andreaippo:
A similarly ugly screenshot:

https://lemm.ee/post/238562

Aloso:
Same issue here. Please use an ellipsis ("Posteing...", "Lesezei...") when the text doesn't fit in one line.

sockenklaus:
Truncating and wrapping text in NavigationBars is discouraged by Material.io guidelines:

https://m3.material.io/components/navigation-bar/guidelines

I'd really like to try how it looks with reduced padding between icons but I feel we have to completely rewrite the component because I've found no way to inject smaller padding with modifiers.

Another idea could be to not use labels at as the icons are pretty descriptive in itself.

Aloso:
> Truncating and wrapping text in NavigationBars is discouraged by Material.io guidelines

so, what's the alternative? I don't see any _actionable_ advice, besides coming up with a shorter label, but that is difficult.

- Instead of _Posteingang_ (inbox), we could use _Postfach_ (mailbox / post office box), which is shorter but less descriptive.

- Instead of _Lesezeichen_ (bookmarks), we could use _Gemerkt_ (remembered / bookmarked), or _Favoriten_ (favourites), but then the app uses different terminology than the web app, which might confuse people.

And this solves the problem for German, but not for other languages.

Aloso:
Also, I noticed that the label font size in Google Drive is smaller than in Google Photos. So even though it's discouraged to make a label smaller to fit the text, it is apparently OK to make _all_ labels smaller by the same amount.

andreaippo:
I'm 

> Also, I noticed that the label font size in Google Drive is smaller than in Google Photos. So even though it's discouraged to make a label smaller to fit the text, it is apparently OK to make _all_ labels smaller by the same amount.

[...]

PR #826, opened 2023-06-22
Replace "Posta in arrivo" with "Inbox"

andreaippo:
This avoid issues with the label in the bottom menu bar

### DIFF
--- a/app/src/main/res/values-it/strings.xml
+++ b/app/src/main/res/values-it/strings.xml
@@ -45,10 +45,10 @@
     <string name="app_name">Jerboa per Lemmy</string>
     <string name="bottomBar_bookmarks">Vai ai miei segnalibri</string>
     <string name="bottomBar_home">Vai alla home</string>
-    <string name="bottomBar_inbox">Vai alla posta in arrivo</string>
+    <string name="bottomBar_inbox">Vai alla inbox</string>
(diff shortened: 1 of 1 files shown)


MV-GH (approved):

NicKoehler:
"Posta" would be good enought

andreaippo:
TBH I'm not proud of this change. It feels like a workaround to solve a problem that should be fixed differently (via proper hyphenation, for example).

So do with this what you will. I'm not a fan of Posta or Messaggi since one is a bit vague and the other a bit misleading. But I understand the criticism towards an English term, and in that case maybe the best option is wrapping words properly in the UI (if feasible, I'm not an Android dev).

MV-GH:
What you mean wrapping the words? Isn't doing that already, if its too long it should put the text below it see #818

andscape-dev:
If it's necessary to have a single-word replacement I think "notifiche" is the most appropriate solution

dessalines (changes requested):
`notifiche` seems like a better fit then

andreaippo:
Really sorry everyone, I just found a few minutes now to update the PR and I'm on GitHub for Android on top of that.

I hope the diff looks good.

Thanks

MV-GH (approved):

dessalines (approved):
No probs, thx!

Issue #808, opened 2023-06-22
Option to not use Chrome Custom Tabs for opening links


jgeorgeson:
- [ x] Did you check to see if this issue already exists?
- [ x] Is this only a single feature request? Do not put multiple feature requests in one issue.
- [ ] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/jerboa

** Describe the feature request below **

Using a Chrome Custom tab means every link the Jerboa opens "internally" is now part of my Chrome history. If I want to save the posts/comments/links, I can use the built-in bookmark/save feature of Lemmy. If I want to have it in my Chrome history, I can select _Open in Chrome_ from the triple-dot menu. 

Making this a toggle between using an embedded WebView, or perhaps the Chrome Custom Tab can be an incognito tab.

twizmwazin:
There is already an option to use an incognito tab if you custom tabs provider allows it. It is under the "look and feel" section, which I know isn't particularly intuitive. It is also poorly supported, chrome is the only browser that advertises any kind of support and you have to enable a flag to get it, and even then I'm not sure that it works all of the time. Maybe try it though until we have a better implementation.

jgeorgeson:
I don't seem to have that option. Is it new in v0.0.34? I'm still on v0.0.33, and only have an option for "Use custom tabs" which is enabled. Disabling it opens all links in Chrome proper rather than an embedded Custom tab. My phone is running Android 13 build TQ2A.230505.002.A1 with Chrome v114.0.5735.130.

twizmwazin:
It is on main but seems like it hasn't been released yet.

dessalines:
Done.

PR #827, opened 2023-06-22
Refactor browser setting to use an enum

twizmwazin:
I was taking a look at implementing an in-app browser for #808, and it quickly stood out to me that we should be using an enum rather than a bunch of booleans to indicate which kind of browser to use when opening links. I've refactored to do exactly that, plus I've refactored the link handling to eliminate regex in the URL parsing. There should also now be a toast if for whatever reason a URL can't be opened, which was silently failing previously.


app/src/main/java/com/jerboa/util/OpenLink.kt
```diff
@@ -0,0 +1,154 @@
...
+import com.jerboa.api.API
+
+enum class BrowserType(val id: Int, val resourceId: Int) {
+    External(0, R.string.browserType_external),
+    CustomTab(1, R.string.browserType_customTab),
+    PrivateCustomTab(2, R.string.browserType_privateCustomTab),
+    ;
+
+    companion object {
+        fun fromId(id: Int): BrowserType {
```
beatgammit, line 19: We can just do:

    BrowserType.values()[id]

app/src/main/java/com/jerboa/util/OpenLink.kt
```diff
@@ -0,0 +1,154 @@
...
+import android.content.Context
+import android.content.Intent
+import android.net.Uri
+import android.widget.Toast
+import androidx.browser.customtabs.CustomTabsIntent
+import androidx.navigation.NavController
+import com.jerboa.R
+import com.jerboa.api.API
+
+enum class BrowserType(val id: Int, val resourceId: Int) {
```
beatgammit, line 12: Looks like enums are supported by Room as of 2.3.0: https://developer.android.com/jetpack/androidx/releases/room#version_230_3

It'll save/load from strings, but we can instead use a converter to keep it as a number if we want: https://stackoverflow.com/questions/57326789/how-to-save-enum-field-in-the-database-room

So we can stick this with the rest of the db stuff instead of here.

app/src/main/java/com/jerboa/MainActivity.kt
```diff
@@ -136,8 +137,7 @@ class MainActivity : ComponentActivity() {
 
                 MarkdownHelper.init(
                     navController,
-                    appSettingsViewModel.appSettings.value?.useCustomTabs ?: true,
-                    appSettingsViewModel.appSettings.value?.usePrivateTabs ?: false,
+                    BrowserType.fromId(appSettings?.browserType ?: 1),
```
beatgammit, line 140: Ideally use enum variants here, like:

    BrowserType.values().getOrDefault(...)

dessalines:

PR #833, opened 2023-06-22
Added switch to mark new posts as NSFW / toggle tag on existing posts

### LINKED ISSUE #679: No option to mark a new post NSFW
** Jerboa Version **
0.0.34

**Describe the bug**
There seems to be no way to mark a new post as NSFW.

**To Reproduce**
Create a new post via web interface, you see a checkbox for NSFW.
Screenshot_20230615-211518

Create a post in Jerboa: There is no checkbox for NSFW.
Screenshot_20230615-211920

sockenklaus:
Closes #679 

Added a switch to `CreatePost.kt` and `PostEdit.kt` and added logic to the corresponding activities.

Screenshot_20230623-002021
Screenshot_20230623-002029

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/post/composables/PostComposables.kt
+++ b/app/src/main/java/com/jerboa/ui/components/post/composables/PostComposables.kt
@@ -1,0 +1,32 @@
+package com.jerboa.ui.components.post.composables
+
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.material3.Checkbox
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.res.stringResource
+import com.jerboa.R
+
+@Composable
+fun CheckboxIsNsfw(
+    checked: Boolean,
+    onCheckedChange: (checked: Boolean) -> Unit,
+) {
+    Row(
+        verticalAlignment = Alignment.CenterVertically,
+        horizontalArrangement = Arrangement.SpaceBetween,
+        modifier = Modifier.fillMaxWidth(),
+    ) {
+        Text(
+            text = stringResource(R.string.create_post_tag_nsfw),
+        )
+        Checkbox(
+            checked = checked,
+            onCheckedChange = onCheckedChange,
+        )
+    }
+}
--- a/app/src/main/java/com/jerboa/ui/components/post/create/CreatePost.kt
+++ b/app/src/main/java/com/jerboa/ui/components/post/create/CreatePost.kt
@@ -6,6 +6,7 @@ import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.PaddingValues
+import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.fillMaxWidth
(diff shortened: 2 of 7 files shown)


twizmwazin (changes requested):
I personally think a checkbox is more appropriate here. I checked the material guidelines and they seem to prefer a checkbox in cases where it is part of a form. Switches are for things that take immediate effect in the UI (maybe we should refactor settings...). Consider right aligning it too and see if that looks good. Additionally, I think "Mark post as NSFW" might be better than just "NSFW".

sockenklaus:
> I personally think a checkbox is more appropriate here. I checked the material guidelines and they seem to prefer a checkbox in cases where it is part of a form. Switches are for things that take immediate effect in the UI (maybe we should refactor settings...). 

Would like to show me your source? According to [Material.io](https://m3.material.io/components/switch/guidelines)  _"Switches are the preferred way to adjust settings. They're used to control binary options – think On/Off or True/False."_ furthermore _"Use switches to turn states on or off, specifically on mobile instead of a checkbox"_.

According to checkboxes, [Material.io](https://m3.material.io/components/checkbox/guidelines) says: _"Use checkboxes to [...] Select one or more options from a list [...] Turn an item on or off in a desktop environment. Checkboxes should be used instead of switches if multiple options can be selected from a list"_.

I read this in favour of Switches in this situation.

> Consider right aligning it too and see if that looks good. 

Agreed! I will try both: `Arrangement.SpaceBetween` as well as switching control and label but aligning both left. Both should be better than the current arrangement.

> Additionally, I think "Mark post as NSFW" might be better than just "NSFW".

Agreed!

twizmwazin:
> Would like to show me your source? According to [Material.io](https://m3.material.io/components/switch/guidelines) "Switches are the preferred way to adjust settings. They're used to control binary options – think On/Off or True/False." furthermore "Use switches to turn states on or off, specifically on mobile instead of a checkbox".

I think the difference in interpretation is that a form and a setting are two different things. They both have binary fields, but the difference is that toggling an option in a form isn't toggling the state. No state is actually created until the form is submitted. On https://m3.material.io/components/switch/guidelines it states:

> Use switches to:
> - Toggle a single item on or off
> - Immediately activate or deactivate something

This language leads me to feel it is geared towards settings rather than form fields. On the same page:

> Alternate selection controls
> Checkboxes and radio buttons are alternative selection controls that can be used to enter decisions or declare preferences such as settings or dialogs.

This language pushes me towards preferring a checkbox in cases where the input my be a "decision".

I see the validity in both of our interpretations and I appreciate you explaining how you came to your conclusion. If the wording and layout changes are implemented I'm fine with a toggle switch, though if anyone else has an experience or opinion to offer on the subject I think it would be welcome.

sockenklaus:
> I think the difference in interpretation is that a form and a setting are two different things. They both have binary fields, but the difference is that toggling an option in a form isn't toggling the state. No state is actually created until the form is submitted.

Oh I see that now. I did a little research on UX regarding switches and checkboxes and all articles I read clearly support your stance that switches are to be used when you want to communicate an instantaneous state change and checkboxes are preference when the decision takes effect after a further action (button press). I also get what you meant by "state change" vs. "the state is created when the form is submitted" in this context.

> I see the validity in both of our interpretations and I appreciate you explaining how you came to your conclusion. 
Thank you, that's very kind of you and I appreciate you taking the time to discuss this with me. I hope I didn't come off as fussy asking for further explanation.

I will change the PR tonight and address the layout changes.

sockenklaus:
Here the current changes. I also moved the new option as its own composable out of CreatePost and Posted it. I feel like this could be done with some more elements to reduce unnecessary redundancy. 

Maybe I find the time later the day and create its own PR.

Screenshot_20230623-155538__01

twizmwazin (approved):

Issue #356, opened 2023-04-01
Saving drafts


suoko:
It'd be nice to save drafts so you can finish writing you post afterwards.

dessalines:
Agree, but I probably won't be able to get to it anytime soon.

xxxserxxx:
Any way to prioritize this? Several times now (more than 3), I've written out long responses or posts and had a failure. When this happens, I've lost the entire post, which is especially painful when it's been typed out on a mobile phone.

The failures come in three types, not all are Jerboa, but they could all be alleviated if Jerboa autosaved drafts like Fedilab does.

1. I navigate away to look something up for the post; when I return, Jerboa shows a white screen and no widgets, and I've lost my post.
2. Crash. This is quite rare, but I've had it happen while posting
3. There's some error while posting. Maybe this is on the server side, or just a network error, or whatevet. In any case, the post is lost.

Having a draft autosave would be a godsend for these cases.

peake100:
+1 on this. With Lemmy getting so overloaded I've lost posts three times in the last hour from timeouts.

I've resorted to writing my posts in a separate note application and then copy/pasting it into jerboa so if the request times out I don't lose what I wrote.

lucaspar:
Looking forward to this, few things are more dissatisfying than having whatever you just wrote lost because of technical issues out of your control.

With Lemmy instances overloaded I'm unfortunately seeing this all the time lately.

dessalines:
I don't have time atm, but anyone else is free to work on it.

PR #844, opened 2023-06-23
Initial drafts support

beatgammit:
This is still a WIP. TODO:

- update dialog when deleting drafts
- fix drafts to load earlier
- test db upgrades (fresh install worked though)
- styling (esp. delete button)

But it does work. Here are some initial screenshots:

image

image

But it does technically work, so mostly just needs polish.


beatgammit:
Should fix #356.

dessalines:

PR #845, opened 2023-06-23
Fix for scrolling up past images in the comments jerks you back down

### LINKED ISSUE #730: Scrolling up past images in the comments jerks you back down

### LINKED ISSUE #825: Post scrolling issues
**Jerboa Version**
v0.0.35-alpha

Android: 13 (AOSP)

**Describe the bug**
Some posts with embed images it's impossible to scroll up after reaching the bottom 

**To Reproduce**

1. Go to https://lemmy.pt/post/453555
2. Scroll all to botton
3. Scroll down back up

Video:  

https://github.com/dessalines/jerboa/assets/44146/34566ac7-cf70-4248-a299-b1a68ceb9b69

nahwneeth:
Closes #730 
Closes #825 

This happens only with images that go beyond the available canvas width. The attached videos shows 2 images one after another and the jerk down only occurs for the larger image which is of res 1600x1200 while only 990px was the available width.

This seems to be a problem with the `CoilImagePlugin` (or maybe that's how it's meant to behave) which doesn't call the image resolver if the canvas width is not known. The canvas width is obtained during the first paint call.

So, the initial bounds are 1600x1200 and during the first paint call, it realizes the image won't fit and reduces the bounds based on the canvas width. My assumption is this change in bounds is picked up by the lazy column and the jerk down happens trying to maintain the scroll position.

Other ways to test this:
- Provide a custom image size resolver which always returns a fixed size and the jerk down will still happen.
- Add a Modifier.onSizeChange() to the AndroidTextView and log the markdown & size. One can see that it initially takes up a large size and immediately reduces itself.

To solve this, I tried wrapping the `AndroidTextView` with a `BoxWithConstraints` and initialized the canvas width of `AsyncDrawable` with the maxWidth available. This seems to work. After the first paint call the canvas width is picked up and there is no (or minimum) change in bounds. Changing orientation of the device also works as intended.

https://github.com/dessalines/jerboa/assets/55971921/85e7fc14-8e13-47f3-b481-4626800b6a6d

https://github.com/dessalines/jerboa/assets/55971921/fe39cbcf-de20-4cac-a7be-f4da2c344ee4

Link to the [post](https://lemmy.ml/post/1363960) in the above videos.

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/common/MarkdownHelper.kt
+++ b/app/src/main/java/com/jerboa/ui/components/common/MarkdownHelper.kt
@@ -11,6 +11,7 @@ import android.view.View
 import android.widget.TextView
 import androidx.annotation.FontRes
 import androidx.annotation.IdRes
+import androidx.compose.foundation.layout.BoxWithConstraints
 import androidx.compose.material.LocalContentAlpha
 import androidx.compose.material.LocalContentColor
 import androidx.compose.material3.MaterialTheme
@@ -18,6 +19,7 @@ import androidx.compose.runtime.Composable
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.takeOrElse
 import androidx.compose.ui.graphics.toArgb
+import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.unit.TextUnit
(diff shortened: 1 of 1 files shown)


dessalines (approved):
Looks great now, thx!

dessalines:
I'll let @twizmwazin take a look before we merge.

twizmwazin (approved):

PR #850, opened 2023-06-23
Copy post and comment information to clipboard

Noxor11:
Allows to copy the thumbnail url of a post if available, and copy a comment made to a post.


MV-GH:
The three composables are almost the change them into a generic something like MyThumbnailCopy and pass the differences as arguments

Noxor11:
Comments don't have thumbnails, but I agree the common code should be contained in a method

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
```diff
@@ -1443,6 +1445,19 @@ fun PostOptionsDialog(
                         icon = Icons.Outlined.Block,
                         onClick = onBlockCommunityClick,
                     )
+                    if (postView.post.thumbnail_url != null) {
+                        IconAndTextDrawerItem(
```
twizmwazin, line 1449: This should probably be grouped with the other copy options in this menu, also allowing the creator to copy the thumbnail link.

app/src/main/res/values/strings.xml
```diff
@@ -66,21 +66,25 @@
     <string name="comment_edit_type_your_comment">Type your comment</string>
     <string name="comment_mention_node_block">Block %1$s</string>
     <string name="comment_mention_node_copy_permalink">Copy Permalink</string>
+    <string name="comment_mention_node_copy_comment">Copy Comment</string>
```
twizmwazin, line 69: Copy comment text

app/src/main/res/values/strings.xml
```diff
@@ -66,21 +66,25 @@
...
     <string name="comment_mention_node_go_to">Go to %1$s</string>
     <string name="comment_mention_node_permalink_copied">Permalink Copied</string>
+    <string name="comment_mention_node_comment_copied">Comment Copied</string>
```
twizmwazin, line 72: Comment text copied

twizmwazin (changes requested):
One other thing, I'm noticing that the dialog is not being dismissed after selecting the added options. If you're feeling up to it, maybe add in the option to copy post text as well while we're here.

Noxor11:
I just realized there's an Utils file, so I used that to place the new function.
Now the popup gets dismissed after clicking, and added ability to copy title, name, and text from a post.

twizmwazin (approved):

PR #855, opened 2023-06-23
Bottom nav bar and screen transitions

### LINKED ISSUE #714: Change Android Back Button Behavior
- [✓] Did you check to see if this issue already exists?
- [✓] Is this only a single feature request? Do not put multiple feature requests in one issue.

The current behavior for the Android OS back button is such that Jeroba will simply cycle backwards through the pages you've visited, irrespective of what "level" they are at in the app navigation.

For example, if you open the app, and from the home screen use the app's bottom nav bar to go to your inbox, then your profile, then your bookmarks, hitting the Android back button repeatedly will get you:

Bookmarks → profile → inbox → app home → exit app

Instead, the app should cause any use of the lower app navigation bar to reset the stack for the back navigation. It should reset to:

Current screen → app home → exit app

This would help get rid of the weird behavior where you cycle through multiple screens on the same level of the app before unexpectedly exiting because you hit the OS back button too many times. It would also get rid of the weird behavior where hitting the OS back button can send you back into a comments section long after you navigated multiple screens away and have forgotten about it.

While I'm sure "fixing" this navigation issue will be a low priority, I still wanted to put in the request just so we at least have a placeholder.

### LINKED ISSUE #830: Tapping bottom navigation buttons multiple times spawns duplicate instances of the view.
**Jerboa Version**
0.0.34

**Describe the bug**
When tapping any button in the bottom navigation bar, a new instance of the view is created. For instance, pressing the Search button 5 times creates 5 separate Search views that all need to be closed individually.

If you tap a bottom nav button that's currently selected, new instances shouldn't spawn on top of it. 

**To Reproduce**

1. Tap a bottom navigation button, such as search, 3 or 4 times.
2. Tap the back arrow to close the view.
3. You'll need to close multiple views depending on how many times you tapped the button

nahwneeth:
- Added slide in and out screen transitions.
- Removed multiple `BottomAppBarAll`.
- Added `BottomNavActivity` which holds the bottom navigation bar and uses a second NavHost to shows the screens.
- Updated drawer to not open inbox, saved & profile as new screens.
- CommunityListAcitivity does not show bottom nav bar when selecting a community for creating a post.
- Tapping multiple times on a bottom nav tab doesn't open multiple views.

- The selected icon is filled and the others are outlined.
- Both the bottom nav bar and the navigation bar below it will always have the same color.
- Back button isn't shown for screen opened from the bottom nav bar.

- Fixed the bottom nav bar padding rising with the keyboard.

Closes #714
Closes #830 

PS: This isn't a huge commit. The home activity has one less indent coz drawer was moved.


twizmwazin:
This change seems to conflict with the "Show navigation bar" option in settings. If this setting is disabled, there doesn't seem to be a way to navigate from search/inbox/saved/profile back to the home page. There is no back or drawer indicator on the page, and hitting back just exits the app.

nahwneeth:
I've restored the back button if bottom nav bar isn't shown.

PR #860, opened 2023-06-24
Consolidate composables for CreatePost and PostEdit

sockenklaus:
Those two forms had lots of redundancies.

I consolidated the redundant composables to reduce unnecessary clutter.

Unfortunately I couldn't test image handling because voyager.lemmy.ml doesn't seem to allow image uploads and i don't have a local instance for testing purposes (yet).


MV-GH:
You don't need voyager, there is about 60+ instances already upgraded to 0.18

sockenklaus:
> You don't need voyager, there is about 60+ instances already upgraded to 0.18

Ah thanks for the advice, you're probably correct. I just don't want to spam a live instance with my testing.

MV-GH:
I usually use /c/jerboa for testing, I dont think they ll mind

sockenklaus:
**Todo:**
* Remove unnecessary "SuppressWarning" annotation
* Implement consistent behaviour for attached images
* Check CreateEditPostBody composable for unnecessary child composables and remove them

app/src/main/java/com/jerboa/Utils.kt
```diff
@@ -462,8 +462,12 @@ fun pictrsImageThumbnail(src: String, thumbnailSize: Int): String {
     return "$host/pictrs/image/$path?thumbnail=$thumbnailSize&format=webp"
 }
 
-fun isImage(url: String): Boolean {
-    return imageRegex.matches(url)
+fun isImage(url: String?): Boolean {
```
twizmwazin, line 465: Why was this changed?

sockenklaus: Oh thanks for pointing that out. This belongs to #877 but is also wrong: 
I thought I could get away with one less check for null statement and I wrongly assumed `someRegex.matches(null)` results in `false` but it actually leads to a NullPointerException.

Anyway, doesn't belong in this PR anyway and will be removed.

PR #868, opened 2023-06-24
Add totp field to login

### LINKED ISSUE #867: 2fa support

Lemmy UI 0.18 supports 2FA. Once I set it up, I can't login to my Lemmy account anymore from jerboa. Please add support for 2FA

MV-GH:
Adds the 2fa field for TOTP to the login form. Logging in without 2fa still works. Can't test if totp field actually works because I can't even get it work on lemmy-UI possible due to https://github.com/LemmyNet/lemmy/issues/3316 (tried with ms authenticator)
But I am positive that it should work.

There was no autofill for TOTP so i used smsOTP should be enough.
Fixes #867


KaKi87:
Hello,
Thanks for implementing this.
I'd be grateful if we could be notified when it gets released.
Thanks

twizmwazin:
You can use the "watch" feature on github to be notified about new releases on this repo, or there is a post on https://lemmy.ml/c/jerboa for new releases.

KaKi87:
Yes, but that requires reading changelogs until finding this feature in one, and unsubscribing afterwards, while a notification in here (or only closing issues when PR is released like some other apps do) allows users to be notified only for what they're waiting for.

twizmwazin:
It'll be part of 0.0.36 when it is released

PR #872, opened 2023-06-24
Add swiping between posts in PostActivity

kensand:
This was a feature that I absolutely could not live without in BaconReader. This lets you swipe left and right when in a post to get to the next or previous one. Let me know your thoughts.

Thanks!


MV-GH:
Does this conflict with  #785 ?

kensand:
Huh - To be honest, I didn't even notice that feature. I suppose this renders that non-functional - I'm using my build right now on my phone and it does not go back to the HomeActivity when swiping right.

I guess my question then is how to best bring the two features together? Personally, I can live without #785 in favor of this.

MV-GH:
The codeowners are gonna have to decide that but most likely you will have to set an option somewhere to decide between yours, the previous and none.

kensand:
Yeah, that's probably the best solution - I'll get started on adding that as an option in the meantime.

twizmwazin (commented):
Let's make this a setting, "Allow swiping between posts" or something like that. Disabled by default. When disabled, the current behaviour is maintained where swiping right on the post view will return you to the home/community view. If enabled, then you must use back or the back arrow in the top left to navigate back, and swiping left/right switches between posts.

kensand:
Cleaned everything up a bit and got it functioning as requested @twizmwazin - should be all set for your review :crossed_fingers: 

Thanks for taking the time!

app/src/main/java/com/jerboa/ui/components/common/SwipeBetweenPosts.kt
```diff
@@ -0,0 +1,84 @@
...
+
+@Composable
+fun SwipeBetweenPosts(
+    homeViewModel: HomeViewModel,
+    postViewModel: PostViewModel,
+    account: Account?,
+    content: @Composable () -> Unit,
+) {
+    val forward = SwipeAction(
+        icon = { Text("Next") },
```
twizmwazin, line 23: This and any other strings in the UI need to be localized.

app/src/main/java/com/jerboa/ui/components/common/SwipeBetweenPosts.kt
```diff
@@ -0,0 +1,84 @@
...
+                            }
+                        } else if (currIndex >= 0) {
+                            postViewModel.initialize(res.data.posts[currIndex + 1].post.id.left())
+                            postViewModel.getData(account)
+                        }
+                        Unit
+                    }
+            }
+        },
+        background = Color.Transparent,
```
twizmwazin, line 56: Use `MaterialTheme.colors.background`

app/src/main/java/com/jerboa/ui/components/common/SwipeBetweenPosts.kt
```diff
@@ -0,0 +1,84 @@
...
+    )
+
+    val backward = SwipeAction(
+        icon = { Text("Next") },
```
twizmwazin, line 60: Probably a typo

app/src/main/java/com/jerboa/ui/components/home/HomeViewModel.kt
```diff
@@ -77,7 +77,7 @@ class HomeViewModel : ViewModel() {
         }
     }
 
-    fun appendPosts(form: GetPosts) {
+    fun appendPosts(form: GetPosts) =
```
twizmwazin, line 80: Are any of the changes in this file part of the feature?

app/src/main/res/values/strings.xml
```diff
@@ -214,6 +214,7 @@
     <string name="look_and_feel_theme_color_pink">Pink</string>
     <string name="look_and_feel_theme_color_purple">Purple</string>
     <string name="look_and_feel_theme_color_woodland">Woodland</string>
+    <string name="look_and_feel_swipe_between_posts">Swipe between posts</string>
```
twizmwazin, line 217: I don't love this settings label. "Allow swiping between posts" might be a bit better while still being reasonably concise.

twizmwazin (changes requested):

app/src/main/java/com/jerboa/ui/components/home/HomeViewModel.kt
kensand: No, sorry, my ide got a bit overzealous

app/src/main/java/com/jerboa/ui/components/common/SwipeBetweenPosts.kt
kensand: This seems to cause the swipe page to be white regardless of the theme, which is causes swipes to be very flashy in dark or black mode.

I think you may have meant `MaterialTheme.colorScheme.background`? It matches the PostActivity background color regardless of theme.

app/src/main/java/com/jerboa/ui/components/home/HomeViewModel.kt
kensand: However, this one line (80) is important - it causes `appendPosts` to return a `Job` instead of `Unit`, which `SwipeBetweenPosts` then uses to `invokeOnCompletion` after the next page has been loaded. Otherwise though, yes, there are extraneous changes to the file.

nahwneeth:
What behaviour should be expected when the post is opened via deep link? The post might not be in the list of posts loaded by HomeViewModel.

kensand:
@nahwneeth If the current post is not in the post list, then swiping will simply have no effect; there is no forward/backward on a deep link, right?

app/src/main/res/values/strings.xml
```diff
@@ -214,6 +214,7 @@
     <string name="look_and_feel_theme_color_pink">Pink</string>
     <string name="look_and_feel_theme_color_purple">Purple</string>
     <string name="look_and_feel_theme_color_woodland">Woodland</string>
+    <string name="look_and_feel_allow_swipe_between_posts">Allow swipe between posts</string>
```
twizmwazin, line 217: swipe -> swiping

app/src/main/java/com/jerboa/MainActivity.kt
```diff
@@ -84,6 +84,7 @@ class JerboaApplication : Application() {
 
 class MainActivity : AppCompatActivity() {
     private val siteViewModel by viewModels<SiteViewModel>()
+    private val homeViewModel by viewModels<HomeViewModel>()
```
twizmwazin, line 87: I don't think we want to be doing this, as it is a partial reversion of #817. Is there another way we could approach this? Or failing that, we should make sure that we don't accidentally re-introduce issues that PR solved.

twizmwazin (changes requested):

app/src/main/java/com/jerboa/MainActivity.kt
nahwneeth: @kensand 

Check out the community & community sidebar routes. They both have to share the same community view model which is similar to this case.

- Create a `navigation` with a routename (say Route.Graph.Home) and put both `BottomNavActivity` & `PostActivity` inside it.
- When creating the viewmodel use the parent backstack entry (ie Route.Graph.Home, see the community route on how to do that).
- Deeplinks will not work in nested navigation, so a redirect route has to be created after the `navigation` (again, see the redirect for community acitivity deeplink).

nahwneeth: Just realized the `PostActivity` can be opened from the community page as well. Should it then show only the posts from the community when swiping left and right?

kensand: > Just realized the PostActivity can be opened from the community page as well. Should it then show only the posts from the community when swiping left and right?

Yes, it should. It looks like these navigation changes have added significantly complexity to this, especially considering that I now have to consider two different sources for posts, deeplinks, and a navigation system I have no experience with. I'm not sure how I want to proceed with this yet.

kensand: Ok, I think I got navigation worked out - had to add a separate path & `PathActivity` for posts opened from a community, but it works as intended above. 

Made the changes as requested (I hope). Thanks for looking!

MV-GH:
@twizmwazin imo there should also be an option to disable both behaviours, i frequent /c/jerboa and #979. There are users who do not want either. Simple turning it into a Enum with a additional None would suffice and it should default to none.

kensand:
> @twizmwazin imo there should also be an option to disable both behaviours, i frequent /c/jerboa and #979. There are users who do not want either. Simple turning it into a Enum with a additional None would suffice and it should default to none.

FWIW, I agree, and am happy to implement it that way :+1:

app/src/main/java/com/jerboa/model/CommunityViewModel.kt
```diff
@@ -53,6 +57,12 @@ class CommunityViewModel : ViewModel(), Initializable {
         mutableStateOf(ApiState.Empty)
     private var blockPersonRes: ApiState<BlockPersonResponse> by mutableStateOf(ApiState.Empty)
 
+    var communityId: CommunityId? by mutableStateOf(null)
+    var communityName: String? by mutableStateOf(null)
+
+    var fetchingMore by mutableStateOf(false)
+        private set
```
dessalines, line 64: You'll need to re-merge from main, because all these should be removed now.

dessalines (changes requested):

app/build.gradle.kts
```diff
@@ -93,6 +93,8 @@ android {
     }
     buildFeatures {
         compose = true
+        dataBinding = true
+        viewBinding = true
```
MV-GH, line 97: I don't think we need these

app/schemas/com.jerboa.db.AppDB/18.json
```diff
@@ -64,7 +64,7 @@
...
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`id` INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, `font_size` INTEGER NOT NULL DEFAULT 16, `theme` INTEGER NOT NULL DEFAULT 0, `theme_color` INTEGER NOT NULL DEFAULT 0, `viewed_changelog` INTEGER NOT NULL DEFAULT 0, `post_view_mode` INTEGER NOT NULL DEFAULT 0, `show_bottom_nav` INTEGER NOT NULL DEFAULT 1, `show_collapsed_comment_content` INTEGER NOT NULL DEFAULT 0, `show_comment_action_bar_by_default` INTEGER NOT NULL DEFAULT 1, `show_voting_arrows_in_list_view` INTEGER NOT NULL DEFAULT 1, `show_parent_comment_navigation_buttons` INTEGER NOT NULL DEFAULT 1, `navigate_parent_comments_with_volume_buttons` INTEGER NOT NULL DEFAULT 0, `use_custom_tabs` INTEGER NOT NULL DEFAULT 1, `use_private_tabs` INTEGER NOT NULL DEFAULT 0, `secure_window` INTEGER NOT NULL DEFAULT 0, `blur_nsfw` INTEGER NOT NULL DEFAULT 1, `save_search_history` INTEGER NOT NULL DEFAULT 1)",
```
MV-GH, line 67: Do not edit existing schemes

app/src/main/java/com/jerboa/db/AppDB.kt
```diff
@@ -42,7 +42,7 @@ val APP_SETTINGS_DEFAULT = AppSettings(
 )
 
 @Database(
-    version = 26,
+    version = 27,
```
MV-GH, line 45: I don't think this would necessitate a migration

app/src/main/java/com/jerboa/model/CommunityViewModel.kt
```diff
@@ -103,25 +108,33 @@ class CommunityViewModel(account: Account, communityArg: Either<CommunityId, Str
...
+            sort = sortType,
+            auth = jwt,
+        )
 
-            val newRes = apiWrapper(API.getInstance().getPosts(form.serializeToMap()))
+        val newRes = apiWrapper(API.getInstance().getPosts(form.serializeToMap()))
 
             postsRes = when (newRes) {
                 is ApiState.Success -> {
+                    if (newRes.data.posts.isEmpty()) { // Hit the end of the posts
```
MV-GH: I had removed this before

app/src/main/java/com/jerboa/model/CommunityViewModel.kt
```diff
@@ -277,6 +290,62 @@ class CommunityViewModel(account: Account, communityArg: Either<CommunityId, Str
...
+        val res = postsRes
+        return if (res is ApiState.Success) {
+            if (current == null) {
+                res.data.posts.firstOrNull()?.post?.id
+            } else {
+                res.data.posts
+                    .mapIndexed { index, postView -> index to postView }
+                    .firstOrNull { it.second.post.id == current }
+                    ?.first?.let { currIndex ->
+                        if (currIndex >= res.data.posts.size - 7) {
```
MV-GH, line 302: This 7 is a buffer of next posts loaded already? Ifso make it constant somewhere that explains that

app/src/main/java/com/jerboa/model/CommunityViewModel.kt
```diff
@@ -103,25 +108,33 @@ class CommunityViewModel(account: Account, communityArg: Either<CommunityId, Str
 
     fun appendPosts(id: CommunityId, jwt: String?) {
         viewModelScope.launch {
-            val oldRes = postsRes
-            postsRes = when (oldRes) {
-                is ApiState.Appending -> return@launch
-                is ApiState.Holder -> ApiState.Appending(oldRes.data)
-                else -> return@launch
-            }
+            fetchMore(id, jwt)
```
MV-GH, line 111: ViewmodelScope.launch inside a viewModelScope.launch?

app/src/main/java/com/jerboa/model/CommunityViewModel.kt
```diff
@@ -103,39 +108,46 @@ class CommunityViewModel(account: Account, communityArg: Either<CommunityId, Str
...
-                            ),
-                        ),
-                    )
-                }
+        val newRes = apiWrapper(API.getInstance().getPosts(form.serializeToMap()))
 
-                else -> {
+        postsRes = when (newRes) {
+            is ApiState.Success -> {
+                if (newRes.data.posts.isEmpty()) { // Hit the end of the posts
```
MV-GH, line 135: I had removed this

app/src/main/java/com/jerboa/ui/components/common/PostStream.kt
```diff
@@ -0,0 +1,10 @@
+package com.jerboa.ui.components.common
+
+import com.jerboa.datatypes.types.PostId
+import com.jerboa.db.entity.Account
+
+interface PostStream {
+    fun getNextPost(current: PostId?, account: Account? = null): PostId?
```
MV-GH, line 7: Account is not nullable anymore

app/src/main/java/com/jerboa/ui/components/common/Route.kt
```diff
@@ -2,10 +2,14 @@ package com.jerboa.ui.components.common
...
 import androidx.navigation.NavBackStackEntry
 import androidx.navigation.NavType
+import com.jerboa.datatypes.types.CommunityId
+import com.jerboa.datatypes.types.PostId
 
 object Route {
     object Graph {
+        const val HOME = "graph_post"
         const val ROOT = "graph_root"
+        const val COMMUNITY = "graph_community"
```
MV-GH, line 12: For what are these extra graphs neccessary?

build.gradle.kts
```diff
@@ -1,14 +1,21 @@
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
+buildscript {
```
MV-GH, line 3: Remove this

build.gradle.kts
```diff
@@ -1,14 +1,21 @@
...
+    repositories {
+        google()
+        mavenCentral()
+    }
+}// Top-level build file where you can add configuration options common to all sub-projects/modules.
+
 plugins {
-    id("com.android.application") version "8.1.1" apply false
-    id("com.android.library") version "8.1.1" apply false
+    id("com.android.application") version "8.0.2" apply false
```
MV-GH, line 11: Are these remnants?

MV-GH (changes requested):
Thanks for coming back to this. I ll take a deeper look some time later

app/src/main/java/com/jerboa/model/CommunityViewModel.kt
kensand: Yes, it's intended to ensure that there are posts loaded before the user swipes left - otherwise the swipe will have no effect, which leads to a poor experience when you hit the end of a page (you end up having to swipe twice)

Will pull out as a const.

app/src/main/java/com/jerboa/model/CommunityViewModel.kt
```diff
@@ -277,6 +279,65 @@ class CommunityViewModel(account: Account, communityArg: Either<CommunityId, Str
...
+                } else {
+                    null
+                }
+            }
+        } else {
+            null
+        }
+    }
+
+    override fun isFetchingMore(): Boolean = when (this.postsRes) {
```
MV-GH, line 336: Use the `isLoading` overload

app/src/main/java/com/jerboa/model/HomeViewModel.kt
```diff
@@ -243,6 +245,64 @@ class HomeViewModel(private val accountRepository: AccountRepository) : ViewMode
...
+                } else {
+                    null
+                }
+            }
+        } else {
+            null
+        }
+    }
+
+    override fun isFetchingMore(): Boolean = when (this.postsRes) {
```
MV-GH, line 301: same here

app/src/main/java/com/jerboa/ui/components/inbox/InboxActivity.kt
```diff
@@ -290,13 +289,8 @@ fun InboxTabs(
...
 
                     val goToComment = { crv: CommentReplyView ->
-                        // Go to the parent comment or post instead for context
-                        val parent = getCommentParentId(crv.comment)
-                        if (parent != null) {
-                            appState.toComment(id = parent)
-                        } else {
-                            appState.toPost(id = crv.post.id)
-                        }
+                        // TODO navigate to comment (using appState.toComment()) once the route is supported
```
MV-GH, line 292: What's blocking this?

app/src/main/java/com/jerboa/ui/components/post/PostActivity.kt
```diff
@@ -145,18 +147,27 @@ fun PostActivity(
...
-    appState.ConsumeReturn<CommentView>(CommentReplyReturn.COMMENT_VIEW, postViewModel::appendComment)
-    appState.ConsumeReturn<CommentView>(CommentEditReturn.COMMENT_VIEW, postViewModel::updateComment)
+    appState.ConsumeReturn<CommentView>(
+        CommentReplyReturn.COMMENT_VIEW,
+        postViewModel::appendComment,
+    )
+    appState.ConsumeReturn<CommentView>(
+        CommentEditReturn.COMMENT_VIEW,
+        postViewModel::updateComment,
+    )
```
MV-GH, line 170: Formatting changes? Set the column width to at least 120

MV-GH (changes requested):

app/src/main/java/com/jerboa/ui/components/inbox/InboxActivity.kt
kensand: In my testing, navigating to the [Comment route](https://github.com/dessalines/jerboa/blob/main/app/src/main/java/com/jerboa/ui/components/common/Route.kt#L28) caused a crash, presumably because it is not supported by any existing composable :shrug:

kensand:
I found a bug where, after having already gone into the `CommunityActivity` once, going back to the `HomeActivity` and then back to a different `CommunityActivity`, then trying to use swiping between posts no longer works, which appears to be due to the `CommunityViewModel` that is passed into the `PostSwipeWrapper` not being updated/recreated with the new `communityId` (and therefor not containing the `current` `PostId`. I'll need to have a further look at that.

app/src/main/java/com/jerboa/MainActivity.kt
```diff
@@ -446,72 +518,6 @@ class MainActivity : AppCompatActivity() {
...
-                            accountViewModel = accountViewModel,
-                            appState = appState,
-                            useCustomTabs = appSettings.useCustomTabs,
-                            usePrivateTabs = appSettings.usePrivateTabs,
-                            showCollapsedCommentContent = appSettings.showCollapsedCommentContent,
-                            showActionBarByDefault = appSettings.showCommentActionBarByDefault,
-                            showVotingArrowsInListView = appSettings.showVotingArrowsInListView,
-                            showParentCommentNavigationButtons = appSettings.showParentCommentNavigationButtons,
-                            navigateParentCommentsWithVolumeButtons = appSettings.navigateParentCommentsWithVolumeButtons,
-                            siteViewModel = siteViewModel,
```
MV-GH, line 508: the toComment crashes because you removed this composable.

This is the one that makes it possible

MV-GH (changes requested):
There also some warnings that need to fixed. I still need to take a deeper look at this

dessalines:
There are so many changes here that you might be better off starting from `main` and either copy-pasting or cherry-picking.

dessalines:

PR #873, opened 2023-06-24
Add backwards compatible language picker

### LINKED ISSUE #755: Backwards compatible locale / language selection
**Proposed problem:**
The PR #609 resolved support for locale selection, but only for android 13. I believe it's an oversight as Android 13 adoption percentage is only at ~12% as of April ([source](https://www.androidcentral.com/apps-software/android-13-distribution-report)), and many phones won't get the update (anecdotal; mine).

**Proposed solution:**
Query language option stored in profile, and select that language.

**Alternate solution:**
Adding an option in the menus to change language for phones not supporting API level 13, would be a great solution. It would also allow a redirect to the settings menu for Android 13 users looking for language change option.

**Context:**
It is of my understanding, that the current implementation just uses the main language configured in Android.

MV-GH:
Adds a language picker that is backwards compatible with android 13 per app language setting.
Tested on API 26, 31, 33.

As soon as you select the new language it is saved and everything changes to that language.

Fixes #755

image


krestenlaust:
Maybe the button should redirect to language selection screen on phones on version 13, having them use the native configuration instead.

app/src/main/res/xml/locales_config.xml
```diff
@@ -1,10 +1,18 @@
 <?xml version="1.0" encoding="utf-8"?>
+<!--
+ When updating this file also update Utils.localeOptions
+-->
```
twizmwazin, line 4: Is there not a way to retrieve this info at runtime instead of maintaining two different lists?

app/src/main/res/values/strings.xml
```diff
@@ -330,4 +330,17 @@
     <string name="inbox_activity_messages">Messages</string>
     <string name="bookmarks_activity_saved">Saved</string>
     <string name="generic_error">An error has occurred</string>
+    <string name="lang_en" translatable="false">English</string>
```
twizmwazin, line 333: In the system language picker it defines the names for us. Can we access this at runtime instead of keeping it here?

app/src/main/java/com/jerboa/ui/components/settings/lookandfeel/LookAndFeelActivity.kt
```diff
@@ -107,6 +117,27 @@ fun LookAndFeelActivity(
                     .verticalScroll(scrollState)
                     .padding(padding),
             ) {
+                SettingsListDropdown(
+                    title = {
+                        Text(text = "Select your language")
```
twizmwazin, line 122: This string should be localized, but maybe just "Language" for the English version.

twizmwazin (changes requested):

app/src/main/res/xml/locales_config.xml
MV-GH: Not that I am aware, but this also maps from locale to its translated name in that language

app/src/main/res/values/strings.xml
MV-GH: [The locales_config controls which languages are displayed in the language per app setting, so](https://stackoverflow.com/questions/36061116/get-language-name-in-that-language-from-language-code)

Not from there that I can tell but i did find another way

app/src/main/res/xml/locales_config.xml
MV-GH: https://stackoverflow.com/questions/74114067/get-list-of-locales-from-locale-config-in-android-13
There is an option here but this requires reading the xml file manually

MV-GH:
https://github.com/dessalines/jerboa/assets/67873169/54b34e5e-62b8-4f1a-880c-0670bca35823

twizmwazin (approved):

krestenlaust:
Awesome work making it synchronize @MV-GH !

PR #876, opened 2023-06-24
Added ability to switch languages

Noxor11:
Now we can change languages inside of the app. 
I didn't add the selected language to the local database and instead used SharedPreferences because the language change occurs before the database is even ready (when you first open the app). If there's a workaround or a better way to do it let me know.


Noxor11:
Didn't realize there is already a pull request on this

PR #893, opened 2023-06-27
Fix downvote on post in postview doing a upvote

### LINKED ISSUE #891: Downvoting a post actually upvotes it
**Jerboa Version**
0.0.35

**Android version + phone**
13, Xiaomi Redmi

**Describe the bug**
Downvoting a post actually upvotes it

**To Reproduce**

1. Go to any lemmy instance, find a post, open it on jerboa, and web browser
2. Tap on downvote
3. Refresh post in browser / on phone
4. Post is upvoted instead of downvoted

How does these PRs actually pass reviews? Do developers actually read it? jerboa is not trustable, this way anything goes

MV-GH:
Fixes #891 

https://github.com/dessalines/jerboa/assets/67873169/4e6997de-b9ad-4e5e-98a5-f775633ae941

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/post/PostActivity.kt
+++ b/app/src/main/java/com/jerboa/ui/components/post/PostActivity.kt
@@ -332,7 +332,7 @@ fun PostActivity(
                                                     post_id = pv.post.id,
                                                     score = newVote(
                                                         postView.my_vote,
-                                                        VoteType.Upvote,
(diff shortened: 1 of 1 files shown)


MV-GH:
CreatePostLike is a autogenerated type generated from the Lemmy rust backend. We don't modify those, but you can make a issue at lemmy to refactor those types.

PR #738, opened 2023-06-18
Added close on click  functionality to ImageViewerDialog. Solves #736

sockenklaus:
This closes #736

PR #894, opened 2023-06-27
Fix .zoomable functionality in ImageViewerDialog and add "swipe up to close"

### LINKED ISSUE #892: ImageViewerDialog: Functionality of .zoomable modifier broken
> This means that you can no longer tap to toggle the top bar, double tap to quickly zoom in or double tap and drag do zoom in or out, and it adds an ugly white overlay on tap.

_Originally posted by @pipe01 in https://github.com/dessalines/jerboa/issues/738#issuecomment-1609652575_

@pipe01 is correct. My fix was pretty crude without realizing it would kill the described functionality. I will take another look at this.

sockenklaus:
This closes #892 

With my PR #738 I broke the intended functionality of `com.jerboa.ui.components.common.ImageViewerDialog.kt`.

**The originally intended functionality was:**
* tap dialog to hide / show top icon bar
* double tap to zoom in 
* double tap + drag to zoom in and out

**My mistake:**

I added a `.clickable` modifier to the image. This overwrote the indended functionality (...and added a white overlay according to @pipe01 which I honestly could not reproduce)

**My solution in this PR:**

I removed the `.clickable` modifier to restore the original functionality and also added the `.scrollable` modifier to the surrounding box. The box now listens to a scroll gesture and dismisses the dialog if it detects a scroll with a delta < -70. This translates to a "quickly swipe up" gesture.

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/common/ImageViewerDialog.kt
+++ b/app/src/main/java/com/jerboa/ui/components/common/ImageViewerDialog.kt
@@ -11,6 +11,9 @@ import androidx.compose.animation.core.tween
 import androidx.compose.foundation.Image
 import androidx.compose.foundation.background
 import androidx.compose.foundation.clickable
+import androidx.compose.foundation.gestures.Orientation
+import androidx.compose.foundation.gestures.rememberScrollableState
+import androidx.compose.foundation.gestures.scrollable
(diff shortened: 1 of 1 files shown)


MV-GH (approved):
Looks good and works well

PR #514, opened 2023-06-11
Add setting to blur NSFW

TrollBlox:
Accidentally deleted my old pull request. Please tell me if there is anything I should change.

* I will add an option to toggle from the lateral menu too, unless this is not something the creators want.

PR #896, opened 2023-06-27
Add blur option, fix non blurred community icons/banners

### LINKED ISSUE #529: Add option to blur or unblur NSFW posts
https://github.com/dessalines/jerboa/issues/291
https://github.com/dessalines/jerboa/pull/514

If NSFW is enabled simply give the user a choice between blurring or unblurring posts on their feed.

MV-GH:
Add a blur option in settings, that controls if NSFW images should be blurred.
Fix community icons and banners not being blurred for NSFW communities

Refactor app setting defaults, previously app settings default where littered all over the place. If we wanted to change a default we would have to track down a dozen places. Now it is all stored in a central location.

Fixes #529
Makes #514 obsolete and can be closed

Tested on android 13 & 11

### DIFF
--- a/app/src/main/java/com/jerboa/MainActivity.kt
+++ b/app/src/main/java/com/jerboa/MainActivity.kt
@@ -118,7 +118,7 @@ class MainActivity : AppCompatActivity() {
                 fetchInitialData(accountSync, siteViewModel)
             }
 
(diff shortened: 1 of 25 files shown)


twizmwazin (approved):
Looks great, but in the future I would prefer if the refactor change could come separately from the feature change.

PR #897, opened 2023-06-28
Fix edit comment view text is being empty

### LINKED ISSUE #895: CommentEditActivity: content is empty
**Jerboa Version**
main 0934e39

**Android version + phone**
Android 10.1, OnePlus 5t

**Describe the bug**

- When editing a comment the content is not displayed. `var content` is empty.
- I can enter new text and it gets saved properly.

**To Reproduce**

1. Find one of your comments.
2. Click on the three-dot menu
3. Select "Edit"
4. **Don't see your old comment in the text box**
5. Enter some text, click save
6. See your newly entered comment in the comment view...

MV-GH:
Fixes the edit comment view being empty

It was composed first before the initialize was called which sets the commentView. Causing it to set the remember with empty text. I changed the remember to change when commentView changes. Not sure if it is normal that it compose before initialize fires. @nahwneeth ?

Fixes #895 

https://github.com/dessalines/jerboa/assets/67873169/12ccdcee-6aca-46a1-96c9-cca6d416809a

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/comment/edit/CommentEditActivity.kt
+++ b/app/src/main/java/com/jerboa/ui/components/comment/edit/CommentEditActivity.kt
@@ -36,8 +36,8 @@ fun CommentEditActivity(
         commentEditViewModel.initialize(commentView)
     }
 
-    var content by rememberSaveable(commentEditViewModel.commentView.value, stateSaver = TextFieldValue.Saver) {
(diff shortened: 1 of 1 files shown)


nahwneeth:
Yeah I must have forgotten or lost in merge conflict ig. One suggestion: the commentView to be edited is passed to the composable / activity. That could be used directly instead of reading from the viewmodel. The launched effect won't be needed then.

MV-GH:
Yeah that worked too. Kinda do feel like that Viewmodel isnt even need there

PR #899, opened 2023-06-28
Add DB migrations tests

### LINKED ISSUE #898: Add DB migration tests
We need an additional check that verifies that all the migrations are correct. It is easy to make a mistake here. And it wont get caught unless you actively test for it.

MV-GH:
Adds DB migration tests.
Added it to woodpecker too.
The tests need the schemas thus they are included in git. (They were a PITA to get)
Btw why are we allowing queries on the main thread? @dessalines 

Fixes #898


MV-GH:
@dessalines is it possible to get hardware acceleration in the pipeline? The migrations need to run as a instrumentation test which needs to run in a emulator which needs hardware acceleration to run. So I am unable to add this check to the CI.

MV-GH:
image
How is this passing?

dessalines:
> @dessalines is it possible to get hardware acceleration in the pipeline?

I don't know if that android image supports it. God that's annoying that a full emulator is required just to test some migrations.

> Btw why are we allowing queries on the main thread? @dessalines

iirc its necessary for synchronous queries. You could test it without that on tho, and see if you get any errors.

MV-GH:
> I don't know if that android image supports it. God that's annoying that a full emulator is required just to test some migrations.

All instrumentation tests will need it else it will take like 2 hours what would take 5minutes.
see this issue https://github.com/ReactiveCircus/android-emulator-runner/issues/46

Well that check exists so what we dont do any DB queries on ui thread. If we are hitting that check we should instead fix the issue and not disable the check. I'll maybe look into it but that's not for this pr.

app/schemas/com.jerboa.db.AppDB/1.json
dessalines, line 1: So I take it we'll need to remember to add these json files (which should be auto-generated), as part of the migration PR? I assume the lint will fail otherwise.

dessalines (commented):
I didn't see the migration tests being run: https://woodpecker.join-lemmy.org/repos/137/pipeline/494/6

app/schemas/com.jerboa.db.AppDB/1.json
MV-GH: Yes, the schema of current version gets generated each time when run. This is needed for the migration test to compare the version after migration to the outputted schema of the actual DB.

MV-GH:
It is not possible to add the migration tests without hardware support. Thus i removed them from CI. But i added comment that devs who add migrations should check it.

dessalines (approved):

Issue #506, opened 2023-06-10
Add "View source" option for posts (to match comments)


a1studmuffin:
We should add a "View source" option to posts in the same way we have this functionality for comments. It's useful for copying text, inspecting markdown etc.

ZJouba:
Should be able to close this one now

PR #901, opened 2023-06-28
Adding Show Post Source button

ZJouba:
Add "View Post Source" button to Post Options Dialog to show raw text

https://github.com/dessalines/jerboa/assets/42601911/861d6670-b3a4-4181-988b-629222110720

Addressing: #506 

Translations with ChatGPT


MV-GH (changes requested):
Looks good, and works but for the translations next time if you don't know the language don't add them. As you have no guarantee that ChatGPT doesn't generate nonsense.

One thing I would like to have changed though is when you want to see the original comment back the text still displays `View post source` Maybe change this to something like `View original text` (or something better)

app/src/main/res/values-da/strings.xml
dessalines, line 1: Remove all these translations for languages you don't speak.

MV-GH: NL translation is good, you can that one

ZJouba:
@MV-GH I would like to change the button text but how do I do that without passing the viewSource boolean around, it seems messy. Best way that I can figure is to actually create a boolean on the postView.post saying "isSource" or something.
What would you suggest?

twizmwazin:
Can you reuse the viewSource variable?

EntelectChallenge:
I'd have to pass it in everywhere, bit of a hassle for just a text change

app/src/main/res/values/strings.xml
```diff
@@ -335,4 +335,5 @@
     <string name="create_edit_post_close">Close</string>
     <string name="lang_language">Language</string>
     <string name="blur_nsfw">Blur NSFW images</string>
+    <string name="post_listing_view_source">View Post Source</string>
```
dessalines, line 338: Agree with @twizmwazin , either re-use the `comment_node_view_source` string, or just make this one `View Source`

ZJouba: I think he meant can't I reuse the `viewSource` variable to change the button text, but @twizmwazin passing in that variable, all the way from PostListing, just to change the button text feels bad

ZJouba: @dessalines Isn't it better that it has it's own resource string? Somewhere along the line someone might change `comment_node_view_source` not thinking that it will also change for the Post Listing...

ZJouba:
@MV-GH I've updated to use the bool for changing the button from `View Source` to `View Original`

MV-GH (approved):
Tested it and works well, thx for work! Passing the bool down was the best way. For blurNSFW i had to pass it down like 10 composables

PR #904, opened 2023-06-28
Change app display name

Undearius:
This will change the name shown on the android home screen and settings to `Jerboa` while leaving the name on the Play Store as `Jerboa for Lemmy`

This gives a bit of a cleaner look as long app names will either overflow to a new line or will be truncated with an ellipse (eg. `Jerboa for Le...`) but will retain the searchability for users searching "Lemmy" on the Play Store and it's still clear from the title what the app is for.

This does remove the localization but I believe it should be possible to restore this by creating a title.txt file in the respective locale folder under `fastlane/metadata/android/`, although I don't know Fastlane well enough to confirm this.


dessalines (approved):
Thanks, I like this better.

Issue #851, opened 2023-06-23
Jerboa wants Lemmy 0.18 which isn't released


jonsparks:
**Jerboa Version**
0.0.35-alpha

**Describe the bug**
App throws an error on launch asking for a newer version of Lemmy server. I'm currently running 0.17.4 which is the latest release on https://github.com/LemmyNet/lemmy. App should not be ahead of the Lemmy releases, I'm certainly not updating my instance to a beta RC.

**To Reproduce**
Update to latest app version and connect to literally any Lemmy server aside from lemmy.ml.

**In the case of a crash**

Please include the logs.
N/A

MV-GH:
Lemmy 0.18 is supposed to release today, we did a release yesterday so that it would at least be available already on fdroid

beatgammit:
Released half hour ago. 😄

twizmwazin:
The backwards incompatible changes in Lemmy 0.18 have frustrated a lot of people, myself included. Hopefully the main Lemmy folks will take note of this and avoid breaking the API in the next version, at least allowing one version of backwards compatibility to allow for a smooth transitions for users. Unfortunately there's not much that we can do about it right now, so hopefully this can be a lesson for the next cycle.

MJBrune:
If you can detect the version why not support the other API version? Seems very short sighed to blindly roll forward breaking the app for the majority of users.

Issue #832, opened 2023-06-22
Fresh install crashes on launch


Rex-On-GitHub:
**Jerboa Version**
0.0.34 via F-Droid

**Describe the bug**
App opens for a second and disappears. It had been working fine for a week or so until earlier today. Issue persists after deleting cache/data and reinstalling.

**To Reproduce**
Install and attempt to run.

`
06-22 16:59:50.382 D/CompatibilityChangeReporter( 1783): Compat change id reported: 161252188; UID 10332; state: ENABLED
06-22 16:59:50.382 I/ActivityTaskManager( 1783): START u0 {act=android.intent.action.MAIN cat=[android.intent.category.LAUNCHER] flg=0x10200000 cmp=com.jerboa/.MainActivity bnds=[620,1340][824,1572]} from uid 10169

06-22 16:59:50.396 V/SplashScreenExceptionList( 1783): SplashScreen checking exception for package com.jerboa (target sdk:33) -> false
06-22 16:59:50.396 D/CompatibilityChangeReporter( 1783): Compat change id reported: 205907456; UID 10332; state: ENABLED
06-22 16:59:50.396 W/TransactionTracing(  642): Could not find layer id -1
06-22 16:59:50.400 D/CompatibilityChangeReporter( 1783): Compat change id reported: 194833441; UID 10332; state: ENABLED

06-22 16:59:50.410 D/Zygote  (  995): Forked child process 21179
06-22 16:59:50.410 I/ActivityManager( 1783): Start proc 21179:com.jerboa/u0a332 for next-top-activity {com.jerboa/com.jerboa.MainActivity}
06-22 16:59:50.411 D/CoreBackPreview( 1783): Window{1d15129 u0 Splash Screen com.jerboa}: Setting back callback OnBackInvokedCallbackInfo{mCallback=android.window.IOnBackInvokedCallback$Stub$Proxy@184d6e5, mPriority=0}

06-22 16:59:50.456 V/GraphicsEnvironment(21179): ANGLE Developer option for 'com.jerboa' set to: 'default'

06-22 16:59:50.456 V/GraphicsEnvironment(21179): ANGLE GameManagerService for com.jerboa: false
06-22 16:59:50.457 V/GraphicsEnvironment(21179): Neither updatable production driver nor prerelease driver is supported.

06-22 16:59:50.518 D/jerboa  (21179): Fetching posts for anonymous user

[...]

Issue #820, opened 2023-06-22
Server version `0.17.4` is under the minimum supported version `0.18`


babeuh:
**Jerboa Version**
0.0.35

**Describe the bug**
Cannot login to instance (lemmy.world). Says server version `0.17.4` is under the minimum supported version `0.18`.
I'm sorry if I am misunderstanding something but I'm pretty sure version `1.17.4` is the latest.

This does log me in anyways though.

**To Reproduce**

1. Go to 'Login'
2. Try to login to lemmy.world
3. See error

daviessm:
https://github.com/LemmyNet/lemmy/releases - the error is correct, v0.18 is the upcoming version. Jerboa <= v0.0.34 will work with Lemmy <= 0.17.4, Jerboa >= v0.0.35 will work with Lemmy servers running >= v0.18.

babeuh:
Ah sorry for the unnecessary issue

jpodpro:
but it doesn't work because no servers are on 18 yet. why would the latest version be one that doesn't actually work yet! now I have to try to figure out how to find and install an older version. this schedule makes no sense!

PR #907, opened 2023-06-29
Made changes to allow logging into unsupported older server versions

ShinyLuxray:
Hello, I have made a version that patches the errors that come up when logging into pre-0.18 servers. This change will allow people to be logged in on 0.17.x and 0.18 on the same Jerboa app.

Full list:
- Changed the outdated messages to a warning
- added checks if default listing types and default sort types were null, defaulted to 0.

I think this change would be highly desirable for people, feel free to mention any issues that this could help solve: E.g. #832, #820, #851, and maybe more.


ShinyLuxray:
I am a big dunce when it comes to git, see my new pull request 908, I didn't think renaming a branch would close it :/

PR #908, opened 2023-06-29
Allow logging into unsupported older server versions

ShinyLuxray:
Hello, I have made a version that patches the errors that come up when logging into pre-0.18 servers. This change will allow people to be logged in on 0.17.x and 0.18 on the same Jerboa app.

Full list:

    Changed the outdated messages to a warning
    added checks if default listing types and default sort types were null, defaulted to 0.

I think this change would be highly desirable for people, feel free to mention any issues that this could help solve: E.g. https://github.com/dessalines/jerboa/issues/832, https://github.com/dessalines/jerboa/issues/820, https://github.com/dessalines/jerboa/issues/851, and maybe more.

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/login/LoginViewModel.kt
+++ b/app/src/main/java/com/jerboa/ui/components/login/LoginViewModel.kt
@@ -17,6 +17,7 @@ import com.jerboa.api.apiWrapper
 import com.jerboa.api.retrofitErrorHandler
 import com.jerboa.compareVersions
 import com.jerboa.datatypes.types.GetSite
+import com.jerboa.datatypes.types.ListingType
 import com.jerboa.datatypes.types.Login
 import com.jerboa.db.Account
 import com.jerboa.db.AccountViewModel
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/jerboa/ui/components/login/LoginViewModel.kt
```diff
@@ -97,22 +97,23 @@ class LoginViewModel : ViewModel() {
...
+
                         val account = Account(
                             id = luv.person.id,
                             name = luv.person.name,
                             current = true,
                             instance = instance,
                             jwt = jwt,
-                            defaultListingType = luv.local_user.default_listing_type.ordinal,
-                            defaultSortType = luv.local_user.default_sort_type.ordinal,
+                            defaultListingType = if (luv.local_user.default_listing_type != null) luv.local_user.default_listing_type.ordinal else 0,
```
dessalines, line 115: These are exactly the hacks I don't want riddling the codebase. They're just going to keep piling on top of each other until it becomes inscrutable.

ShinyLuxray: I totally understand what you mean.

I'm a shitty coder when it comes to best practices. I'm trying to come up with a better way to do it, would a try/catch statement be better in all the spots? I'll make a commit to switch it around.

twizmwazin: The "good" solution would be abstracting away the API entirely, having a version-agnostic frontend that can utilize a per-version backend. Jerboa then would query the frontend asking what features are supported on an instance, and adjust its behavior accordingly. That's quite a bit of engineering though since we're currently directly using the types autogenerated from the Lemmy API directly throughout the app with no abstraction.

dessalines:
The `SiteResponse` has your server version #. You can use this as a switch to use different types and API calls. Those should be segregated into their own functions, and tagged as `@Deprecated`.

PR #915, opened 2023-06-29
Make updateInstances task compatible with configuration cache

AppearamidGuy:
There were basically two problems:
- Task was calling functions defined in the script. Fixed it by moving everything into a proper task class.
- Task was calling `Project.file`. Replaced it with file output properties. Since task now had defined inputs/outputs it would be cacheable, so I made it not tracked (not cacheable and never up-to-date).


MV-GH (approved):
Lovely change, thanks for that 👍

PR #932, opened 2023-07-01
Add caret to indicate thumbnails that open in the image viewer

LilithSilver:
This feature is inspired by Boost, which has differently-covered carets for different types of media. There, images have white triangles, gifs blue, and videos red. This allows the user to easily predict what tapping the thumbnail will do, and make an informed decision on whether they want to follow a link or just open the image viewer. 

The advantages of this feature include: 

* Safety: The user is less likely to click on suspicious links by accident that seem like simple images.
* Predictability: Sometimes, the user wishes to not exit the app and load another page (for example on mobile). This allows them to only click on images, and not links, instead of guessing.
* Future-proofing: Once Jerboa has in-app video support, users will want a differently-colored indicator to indicate that it may use data or play sound. This PR adds a post type enum that can be extended for video or gif content later.

This feature is only applicable in Small Card and List views.

Screenshot: 

Note the lack of border-radius on the bottom-right corner, when the triangle is enabled.

Note to reviewers: I awkwardly stuck the color in `Color.kt`; I'm not sure how to integrate it with the current theming system. If there's a better way please let me know!


MV-GH:
From your screenshot it also marked the external link with the caret. While on Boost it doesn't do that. Is that intended change in behavior?

MV-GH:
Regarding the colors, to integrate that with the theme, use `MaterialTheme.colorScheme.X`

LilithSilver:
> From your screenshot it also marked the external link with the caret. While on Boost it doesn't do that. Is that intended change in behavior?

The top link hasn't properly generated a thumbnail (that happens sometimes), but it's still an imgur link and correctly opens with the in-app viewer!

The bottom link doesn't have a caret, because it redirects to the browser (verge article).

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
```diff
@@ -1230,29 +1236,45 @@ private fun ThumbnailTile(
...
-        } ?: run {
-            Card(
-                colors = CARD_COLORS,
-                modifier = postLinkPicMod,
-                shape = MaterialTheme.shapes.large,
-            ) {
-                Box(
-                    contentAlignment = Alignment.Center,
-                    modifier = Modifier.fillMaxSize(),
+        Box {
```
MV-GH, line 1239: I am guessing this was needed for the clipping to work?

MV-GH (commented):
Looks good and works fine. Regarding the colors, ill let the others decide if it should be part of the themes. If so you would have to do something like this https://gustav-karlsson.medium.com/extending-the-jetpack-compose-material-theme-with-more-colors-e1b849390d50

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
MV-GH: Oh btw could you already add the video enum and color

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
```diff
@@ -271,7 +275,7 @@ fun PostTitleBlock(
     usePrivateTabs: Boolean,
     blurNSFW: Boolean,
 ) {
-    val imagePost = postView.post.url?.let { isImage(it) } ?: run { false }
+    val imagePost = postView.post.url?.let { getPostType(it) == PostType.Image } ?: run { false }
```
MV-GH, line 278: just `?: false`

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
LilithSilver: The box was added to allow the caret to overlap whatever's there -- whether it's a thumbnail or a link-placeholder.

And done!

app/src/main/res/drawable/triangle.xml
```diff
@@ -0,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
+  <path
+      android:pathData="M0,24l24,-24l0,24l-24,0z"
+      android:fillColor="#FFF"/>
```
dessalines, line 8: I spose there's no other way to avoid making this.

app/src/main/java/com/jerboa/ui/theme/Color.kt
```diff
@@ -1061,3 +1061,9 @@ val Color.muted @Composable get() = this.copy(alpha = 0.5F)
 
 val CARD_COLORS @Composable
 get() = CardDefaults.cardColors(containerColor = MaterialTheme.colorScheme.inverseOnSurface)
+
+// TODO integrate into theme?
+val imageHighlight: Color
+    get() = Color(0xCCD1D1D1)
+val videoHighlight: Color
+    get() = Color(0xCCC20000)
```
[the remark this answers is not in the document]
LilithSilver: I'm not so sure about using existing Material color entries, for a few reasons:

- The colors need to be immediately distinctive between each other. 
- The colors need to be chosen specifically to show up properly over most thumbnails, and material theming doesn't intentionally provide that. 
- Boost has white, green, blue, and red. If we want to work up to that behavior we'll run out of slots.

If we want to integrate these into theme colors, we should find a way to extend Material colors IMO. Thoughts?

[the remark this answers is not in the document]
MV-GH: Yes i agree, and i posted a link how to extend Material colors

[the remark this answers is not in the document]
LilithSilver: Added custom color capability with the theming system according to that article; let me know what you think. The biggest diversion I made was storing a reference to the default material theme rather than mirroring each and every single property; I think it's more future-proof this way, because if Material gets an update with more/less properties we won't have to change that mirroring solution. (Plus there are just a lot). 

Other than that it should be pretty straightforward.

LilithSilver:
Just getting back to this now, apologies for the delay. I pushed a proper color scheme implementation; let me know if this solution works!

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
```diff
@@ -1267,11 +1271,12 @@ private fun ThumbnailTile(
     openImageViewer: (url: String) -> Unit,
 ) {
     postView.post.url?.also { url ->
+        val postType = getPostType(url)
 
         val postLinkPicMod = Modifier
             .size(POST_LINK_PIC_SIZE)
             .clickable {
-                if (isImage(url)) {
+                if (postType != PostType.Link) {
```
MV-GH, line 1279: I prefer `postType == PostType.Image`

MV-GH (changes requested):
I tested this and it works well. Thanks !

app/src/main/java/com/jerboa/ui/theme/Theme.kt
```diff
@@ -34,7 +40,20 @@ fun JerboaTheme(
 
     // Dynamic schemes crash on lower than android 12
     val dynamicPair = if (android12OrLater) {
-        Pair(dynamicLightColorScheme(ctx), dynamicDarkColorScheme(ctx))
+        val jerboaImageHighlight = Color(0xCCD1D1D1)
```
dessalines, line 43: This is wild that add custom colors requires this much code bloat. Is it really that necessary? Now all theme designers have to hack on these colors too, rather than just using the defaults.

The tradeoff of this massive code bloat is not worth it IMO. If our three options are:

- Use the existing colors: primary, tertiary, etc, but made semi-transparent for these carets.
- Less optimal - hard code the colors
- Add these custom colors into every theme, and extend jetpack-compose theming like you've done here.

Please use option 1 or 2, this extra complication is not worth it at all.

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
MV-GH: Not the way I was imagining it, I was going to add a separate videoViewer. If imageViewer would support videos I would have rather that it be named mediaViewer. Or something generic.

LilithSilver: I was also imagining a separate video viewer. This bit of code doesn't really matter tbh until we integrate the image viewer, but I can change it to ==.

app/src/main/java/com/jerboa/ui/theme/Theme.kt
LilithSilver: Option 1 definitely won't work; we need new colors here (see https://github.com/dessalines/jerboa/pull/932#discussion_r1251311355 for my reasoning why).

A note on the implementation: Designers will still be able to use the default colors through `MaterialTheme.colorScheme`; it's just the new `MaterialTheme.jerboaColorScheme` that provides custom colors like these. I don't believe this issue will be unique; there may very well be other times we need colors that aren't present in Material Design 3. We can definitely hardcode each new color, but they will never be compatible with the theming system that users can select, if we do that.

I do agree that it's kind of bloaty, though. I'm fine with reverting to option 2 if it comes down to that.

Thoughts from @MV-GH since they requested the color theming?

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
MV-GH: I was fine with the hard coding tbh. But since we already have this implemented with the theme might as well keep it. @twizmwazin wanna be the tie braker?

app/src/main/java/com/jerboa/ui/theme/Theme.kt
MV-GH: I was fine with the hard coding tbh. But since we already have this implemented with the theme might as well keep it. @twizmwazin wanna be the tie braker?

dessalines (approved):

PR #937, opened 2023-07-01
Show the users' federated name in drawer.

camporter:
This lets users know which instance they are on more easily.

Also change the 'Switch to' name/server format to match the federated name format using `@`.

federated_name_in_drawer


MV-GH:
What happens if the name is way too long? Does it completely clip it out? If so you should probably change the text clipping to do the ellipsis thing.

camporter:
The federated name was previously given one line and clip. Gave it two and ellipsis and it seems like that should be plenty 🙂 

The existing `PersonName` behavior I've left alone.

Both were able to knock the expandable arrow out of the row, so set weights.

app/src/main/java/com/jerboa/ui/components/home/Home.kt
```diff
@@ -436,13 +441,28 @@ fun AvatarAndAccountName(myPerson: Person?, showAvatar: Boolean) {
...
-            person = myPerson,
-            color = MaterialTheme.colorScheme.onSurface,
-        )
+        Column() {
+            PersonName(
+                person = myPerson,
+                color = MaterialTheme.colorScheme.onSurface,
+            )
+            Text(
+                text = myPerson?.let { federatedNameShown(it) } ?: run { "" },
```
MV-GH, line 450: just `?: ""`

camporter: Thanks, done :+1:

MV-GH:
Greek string file should be updated now too

dessalines (approved):
Looks good, thanks!

PR #943, opened 2023-07-01
Fix jerboa icon resize.

camporter:
In 7d9c5a68c47f568d5a17f6630450e2bd267220ac the jerboa icon changed size, but the paths were still at the previous size.

Converted the drawable to SVG, resized using `rsvg-convert`, then exported back to a drawable.

Before:
jerboa_icon_before

After:
jerboa_icon_after


MV-GH (approved):
Thanks

PR #944, opened 2023-07-01
Add Greek Language Support on Jerboa - values-el strings.xml

sv1sjp:
If something is missing please send me a message and I will add/fix it as soon as possible!


MV-GH (changes requested):
Update locales_config.xml too

sv1sjp (commented):
Okay, < locale android:name="el" > has been added in locales_config.xml

MV-GH (approved):

Issue #681, opened 2023-06-15
Top-level exception handling


krestenlaust:
- [x] Did you check to see if this issue already exists?
- [x] Is this only a single feature request? Do not put multiple feature requests in one issue.
- [x] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/jerboa

"We should probably implement some sort of top level exception handling that allows the user to submit a bug report directly from the app to make it easier to get a full backtrace. If you can get the full backtrace from logcat that'd be great." - https://github.com/dessalines/jerboa/issues/538#issuecomment-1586320696

MV-GH:
How will the stacktraces look from a release build?

i ve seen some before and they are basically useless

example
```
	at r.u0.<init>(Unknown Source:72)
	at f1.c.y0(Unknown Source:28)
	at d5.j.X(Unknown Source:173)
	at n0.b.a(Unknown Source:50)
	at n0.b.X(Unknown Source:8)
	at s.j0.T(Unknown Source:181)
	at n0.b.b(Unknown Source:50)
	at n0.b.T(Unknown Source:8)
	at o.t.a(Unknown Source:2503)
	at o.t.Y(Unknown Source:317)
	at n0.b.Y(Unknown Source:58)
	at f7.i.t(Unknown Source:182)
	at x.c1.k(Unknown Source:182)
	at s.t.T(Unknown Source:152)
	at n0.b.b(Unknown Source:50)
	at n0.b.T(Unknown Source:8)
	at t.c.g(Unknown Source:80)
	at s.v.g(Unknown Source:61)
	at t.a.g(Unknown Source:67)
	at s.w.g(Unknown Source:12)
	at n.l0.a(Unknown Source:169)
	at n.l0.Y(Unknown Source:161)
	at n0.b.Y(Unknown Source:58)
	at f7.i.t(Unknown Source:182)
	at p0.h.f(Unknown Source:113)
	at t.k0.f(Unknown Source:32)
	```
We could probably disable obfusication if isnt already since it's foss anyway but minimization/release optimizations will still make it look like that

beatgammit:
It looks like [GlitchTip](https://glitchtip.com/sdkdocs/java-android) is FOSS and supports android reporting, but I have no experience with it. I also don't know how this interacts with optimizers.

[...]

PR #945, opened 2023-07-02
Add local crash tracking and viewing

camporter:
Uses https://github.com/FunkyMuse/Crashy to save and access previous app crashes.

https://github.com/dessalines/jerboa/assets/494446/82a447cc-f972-4104-80a2-4a9ac7ba3bba


camporter:
I've left this in draft just to gauge the feelings behind local crash reporting. There was an issue discussing this and other solutions, but I can't seem to find it now.

MV-GH:
I think you mean #681. This pretty great but I'm guessing if you crash at startup this won't be useable?

camporter:
Right. Start up crashes should still be recorded depending on how bad they are, but won't be accessible to a user without doing something to avoid the crash.

This avoids transmitting crashes somewhere that users might not like or trust, and is at least better than no way (besides system crash reporting) to get crashes to report.

dessalines (commented):
Seems like a good idea to have this, thanks. Fix conflicts then we can test.

camporter:
:+1: rebased

app/src/main/res/values/strings.xml
```diff
@@ -102,6 +102,10 @@
     <string name="community_sortBy">Sort by</string>
     <string name="community_subscribe">Subscribe</string>
     <string name="community_users_month">%1$s users / month</string>
+    <string name="crashes">Crashes</string>
```
twizmwazin, line 105: Maybe "Crash logs", "All crash logs deleted", "Copy crash log" and "Delete all crash logs"?

camporter: Yes that's much easier to read, changed 👍

dessalines (approved):
Tested this, and it seems to work. Thx!

MV-GH (approved):

PR #948, opened 2023-07-02
Update subscribed communities in sidebar when user subscribes

twizmwazin:


app/src/main/java/com/jerboa/ui/components/community/CommunityActivity.kt
```diff
@@ -235,6 +236,11 @@ fun CommunityActivity(
                                                             auth = acct.jwt,
                                                         ),
                                                     )
+                                                    siteViewModel.getSite(
```
MV-GH, line 239: Does this guarantee that getSite only is executed once followCommunity is done?

twizmwazin: Fixed!

app/src/main/java/com/jerboa/ui/components/community/CommunityViewModel.kt
```diff
@@ -144,6 +145,8 @@ class CommunityViewModel : ViewModel(), Initializable {
 
                 else -> {}
             }
+
+            onComplete()
```
MV-GH, line 149: I would only do it on Success, there is no point on updating if the following has failed

app/src/main/java/com/jerboa/ui/components/community/CommunityActivity.kt
```diff
@@ -234,6 +235,13 @@ fun CommunityActivity(
                                                             follow = cfv.subscribed == SubscribedType.NotSubscribed,
                                                             auth = acct.jwt,
                                                         ),
+                                                        onComplete = {
```
dessalines, line 238: Nice, this is simple enough and works.

MV-GH (approved):

Issue #927, opened 2023-06-30
App v0.0.37 crashes when hitting "Back" in a post


scottwallacesh:
**Jerboa Version**
0.0.37

**Android version + phone**
Android 13, Pixel 4a

**Describe the bug**
When reading a post and hitting the "Back" button, the app crashes.

**To Reproduce**
1. Read a post
2. Hit "Back"

**In the case of a crash**
<details><summary>LOGS</summary>

```text
--------- beginning of main
06-30 08:00:47.921 13116 13133 E OpenGLRenderer: Unable to match the desired swap behavior.
06-30 08:00:51.028 13116 13116 W OnBackInvokedCallback: OnBackInvokedCallback is not enabled for the application.
06-30 08:00:51.028 13116 13116 W OnBackInvokedCallback: Set 'android:enableOnBackInvokedCallback="true"' in the application manifest.
06-30 08:00:51.048 13116 13116 D jerboa  : got to post activity
06-30 08:00:51.073 13116 18031 I okhttp.OkHttpClient: --> GET https://lemmy.example.com/api/v3/post?auth=[AUTH_TOKEN]&id=[ID]
06-30 08:00:51.073 13116 18031 I okhttp.OkHttpClient: User-Agent: Jerboa
06-30 08:00:51.073 13116 18031 I okhttp.OkHttpClient: --> END GET
06-30 08:00:51.073 13116 13116 D Compose Focus: Owner FocusChanged(true)
06-30 08:00:51.088 13116 13116 D jerboa  : got to home activity
06-30 08:00:51.088 13116 13116 D jerboa  : got to post activity
06-30 08:00:51.121 13116 13116 D jerboa  : got to post activity
06-30 08:00:51.129 13116 18034 D TrafficStats: tagSocket(96) with statsTag=0xffffffff, statsUid=-1
06-30 08:00:51.805 13116 18031 I okhttp.OkHttpClient: <-- 200 https://lemmy.example.com/api/v3/post?auth=[AUTH_TOKEN]&id=[ID] (732ms)
06-30 08:00:51.805 13116 18031 I okhttp.OkHttpClient: content-type: application/json
06-30 08:00:51.806 13116 18031 I okhttp.OkHttpClient: date: Fri, 30 Jun 2023 07:00:51 GMT
06-30 08:00:51.806 13116 18031 I okhttp.OkHttpClient: permissions-policy: interest-cohort=()
[...]

PR #951, opened 2023-07-02
Redesign post listings on main screen

pieterdd:
I've done a few design tweaks that I hope will improve the user experience.

Before:

Screenshot_20230702-115825_Jerboa_(DEBUG)

After:

Screenshot_20230702-115702_Jerboa_(DEBUG)


MV-GH:
Looks great but it still has some issues when the community uses it full name

image

or
image

pieterdd:
Good point.

I'm having trouble finding a way to tell Compose to shorten the community name and/or username in case of overflow. I could resort to tricks like setting a maximum absolute width, but on bigger screens there's no need for such a limit.

Some possible avenues:
- Experiment with BoxWithConstraints
- Remove weighted total (upvotes and downvotes are already shown)

My intention is to reduce clutter by increasing focus on the post (bigger title), decreasing focus on the metadata (moved it to one line) and letting everything breathe a little bit (card layout with whitespace).

app/src/main/java/com/jerboa/ui/components/post/PostListings.kt
```diff
@@ -73,32 +74,39 @@ fun PostListings(
...
-                usePrivateTabs = usePrivateTabs,
-                blurNSFW = blurNSFW,
-            )
-            Divider(modifier = Modifier.padding(bottom = SMALL_PADDING))
+            Card(
+                colors = CardDefaults.cardColors(
+                    containerColor = MaterialTheme.colorScheme.background,
+                ),
+                modifier = Modifier
+                    .padding(12.dp),
```
camporter, line 82: Thoughts on something closer to `SMALL_PADDING` here?

app/src/main/java/com/jerboa/ui/components/post/PostListings.kt
```diff
@@ -73,32 +74,39 @@ fun PostListings(
...
-                postViewMode = postViewMode,
-                showVotingArrowsInListView = showVotingArrowsInListView,
-                enableDownVotes = enableDownVotes,
-                showAvatar = showAvatar,
-                useCustomTabs = useCustomTabs,
-                usePrivateTabs = usePrivateTabs,
-                blurNSFW = blurNSFW,
-            )
-            Divider(modifier = Modifier.padding(bottom = SMALL_PADDING))
+            Card(
```
camporter, line 77: For the list view mode, we might not want it to be surrounded in a card.

camporter (commented):
Thanks, the cards look much better!

Removing the `Spacer` before `ScoreAndTime` and setting a lower weight (I was trying with `0.5f`) for the `CommunityName` and `PersonProfileName` might make it behave a bit better?

Giving `PersonProfileName` its own row seems to add a lot of breathing room at the cost of some extra vertical space.

app/src/main/java/com/jerboa/ui/components/post/PostListings.kt
pieterdd: I was afraid the effect would be too subtle with `SMALL_PADDING`, but it still works. I've removed the bottom padding entirely to ensure the amount of whitespace between cards equals `SMALL_PADDING`.

pieterdd:
As per your suggestion I ended up going back to two rows. Since tidying up involves compromises, I did disable the display of user avatars in post listings. Community avatar and moderator symbol remain in place.

Here's a preview: https://github.com/dessalines/jerboa/assets/1407980/58fc6215-5872-4ca1-8f1d-fd701bd931d1

app/src/main/java/com/jerboa/ui/components/post/PostListings.kt
pieterdd: Now with the smaller padding, the effect is fairly subtle: https://github.com/dessalines/jerboa/assets/1407980/c92a039c-db49-4f50-a7f9-7580002759fd

Is it okay to keep this as-is? It'd be one less code path to maintain.

MV-GH:
image
Should we vertically center those voting arrows?

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
```diff
@@ -307,7 +303,9 @@ fun PostName(
 
     Text(
         text = postView.post.name,
-        style = MaterialTheme.typography.titleLarge,
+        style = TextStyle(
+            fontSize = 22.sp,
```
camporter, line 307: The titles are no longer controllable by the font size selection in settings.

app/src/main/java/com/jerboa/ui/components/post/PostListings.kt
camporter: With the title font size customizable again, it's not too different vertical space-wise from what I can tell.

pieterdd:
>  Should we vertically center those voting arrows?

I tried it but it looks a little odd on cards with bigger heights.

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
pieterdd: Hmm, not immediately sure how to handle this. Can I somehow edit the default font size of `titleLarge`, and is it safe to do so? Should I make a separate constant?

app/src/main/java/com/jerboa/ui/components/post/PostListings.kt
pieterdd: True, it's not too different from having the divider. As for the padding of the arrows, it looks like the alignment of the composables was already correct:

image

I'm not sure if I can control the padding at the top of the text, so leaving it as-is for now.

MV-GH:
We can't downgrade compose, see #927

MV-GH:
Can you expand on issues rendering previews? Half the time they didnt work before either on 1.4.3

dessalines:
For the card-view, I went with the 1-line for each of name and community, because of the issues with really long names, and not wanting them to flow onto new rows. This comes a card view from infinity for reddit, rather than boost. 

For the list-view, I did use a wrapping line, but it also doesn't have that top-right score indicator.

Am I the only one who doesn't like cards that much, and prefers simple lines to separate posts? Boost doesn't use them on its compact views, and it looks so much cleaner.

Screenshot_2023-07-03-10-16-45-220_com rubenmayayo reddit

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
MV-GH: multiply by it?

dessalines: Just use the existing material sizes, otherwise custom sizing won't work correctly. Try titleMedium.

pieterdd:
I've updated the post title typography so it uses a preset. I've spun off the [render preview](https://github.com/dessalines/jerboa/issues/963) issue as it was an existing issue only affecting my IDE.

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
pieterdd: Especially when viewing image posts with titleMedium, the post title could use some extra visibility. I tweaked titleLarge for this purpose.

dallyh:
I like card style posts (coming from another app), however I personally think that this should be an option instead of the default.

dessalines (commented):
I tested this out locally, and I do like most of the changes except for the cards.

dessalines:

PR #953, opened 2023-07-02
Improve login form

### LINKED ISSUE #879: Improve login form
**Jerboa Version**
Main

**Android version + phone**
Oneplus 5t, Android 10.1

**Describe the bug**
The login form has some cosmetic and usability issues:

1. The drop-down with suggested instances stays on top of a "change input language"-dialog.
2. Suggested instances don't filter when typing
3. When entering long passwords the password field expands to MaxWidth
4. The CircularProgressIndicator is bigger than the login button. When pressed, the login button resizes twice which looks wonky.

**To Reproduce**

5. Click on '....'
6. Scroll down to '....'
7. See error

**In the case of a crash**

Please include the logs.

<details><summary>Click here if you don't know how to get the logs</summary>

Follow these steps (skip the ones you already have done):
- Enable developer options and USB DEBUGGING
- Install ADB on your PC
- Connect your phone to your PC
- Execute this command on PC

### To enable developer options and USD DEBUGGING

[Follow this article](https://developer.android.com/studio/debug/dev-options)
In essence you find the build number in About section of settings page and tap it a few times
Then open developer options and enable USB DEBUGGING

### Install ADB on your PC

Download ADB from here [https://adbinstaller.com/](https://adbinstaller.com/)

Unzip it

[(Click here if you wanna know more about ADB)](https://developer.android.com/tools/adb)

### Connect your phone to your PC

Use a USB cable to connect your phone to your PC

### Execute this command on PC
On windows open Powershell in the directory with the adb.exe (where you unzipped it, if you didn't add it to the PATH)

On Unix open a shell in the directory with the adb (where you unzipped it, if you didn't add it to the PATH)

To test if this is working type `adb devices`, this should output a device

Then execute this command `adb shell logcat --pid $(adb shell "timeout 5s logcat | sed -En 's/.*com.jerboa.*, PID: (\S*)/\1/p'| tail -1")`
The above command will print the logs of the latest execution of Jerboa (regardless of which version/build).

That should look like something like this.

Copy the full output. (If not possible just include the few lines below `FATAL EXCEPTION: main` )

Either include it as a file or past it here

```
<details><summary>LOGS</summary>
PASTE HERE AND REMOVE THE ABOVE GUIDE
</details>
```

</details>

MV-GH:
- Filters instance list
- Can type in dropdown if options are showing
- FieldTexts dont grow with text
- LoadingIndicator fits the button now

Could not solve the KeepassDX autofill issue, according to docs it looks for username and password combo. We have that so the issue must be with KeePassDX since other password managers do work fine.

Fixes #879

https://github.com/dessalines/jerboa/assets/67873169/9e77e87e-dfe1-4e40-afdf-6eb3d40aa60f


app/src/main/java/com/jerboa/ui/components/login/Login.kt
```diff
@@ -201,7 +211,7 @@ fun LoginForm(
             modifier = Modifier.padding(top = 10.dp),
         ) {
             if (loading) {
-                CircularProgressIndicator()
+                CircularProgressIndicator(modifier = Modifier.size(24.dp))
```
dessalines, line 214: Would probably be better to have this size come from the material text size in use.

app/src/main/java/com/jerboa/ui/components/login/Login.kt
```diff
@@ -141,30 +144,35 @@ fun LoginForm(
...
-                        text = {
-                            Text(text = selectionOption)
-                        },
-                        onClick = {
-                            instance = selectionOption
-                            expanded = false
-                        },
-                        contentPadding = ExposedDropdownMenuDefaults.ItemContentPadding,
-                    )
+            val filteringOptions = instanceOptions.filter { it.contains(instance, ignoreCase = true) }
```
dessalines, line 147: This worked a little strange, IE you have to type some letters, then hit the popup for it to filter. But I also remember the other one I had to revert because it had really terrible performance.

MV-GH:
Honestly, did not fully understand nor experience what you wrote regarding the filter thing. But i did make a small change that made it better.

dessalines (approved):
Looks good to me, I'll let @twizmwazin test it too.

PR #962, opened 2023-07-03
Fix deprecations

MV-GH:
There where some left behind depreciations from upgrading to compose 15.3


dessalines (approved):
Thanks for grabbin these.

Issue #967, opened 2023-07-04
Navigation drawer should close when system back is pressed


FilippoVigani:
**Jerboa Version**
0.0.38

**Android version + phone**
Any

Currently the modal navigation drawer can only be closed by tapping the nav icon, by tapping outside, or by dragging the drawer.  Once it's opened, we should be able to close it by using the system nav back button (or the back gesture).

FilippoVigani:
Duplicate of #938

PR #941, opened 2023-07-01
When drawer is open, close it if back is pressed.

camporter:
Fixes #938

PR #968, opened 2023-07-04
Fix nav drawer back interaction

FilippoVigani:
Resolves (#967)


silentTeee:
Does the mainlined #941 no longer address this?

PR #973, opened 2023-07-05
db refactor

### LINKED ISSUE #949: Remove the suppression for DB queries on UI thread check.
Currently we have this `.allowMainThreadQueries()` on DB. This allows DB queries to execute on the UI thread. This is currently necessary for getAccountsSync and possibly more. Remove this suppression and fix the queries on UI thread.
see https://github.com/dessalines/jerboa/pull/899#issuecomment-1612396780

yate:
- Reorganize db into different packages (entity, dao, repository)
- Make all queries async, remove need for allowMainThreadQueries() (Fixes #949)
- Refactor the way the viewModels were getting initialized with repositories based on this codelab https://developer.android.com/codelabs/basic-android-kotlin-compose-persisting-data-room#8


dessalines:
I'm fairly busy, but ping me when you're ready for me to take a look.

app/src/main/java/com/jerboa/MainActivity.kt
```diff
@@ -71,35 +71,28 @@ import com.jerboa.ui.components.report.post.CreatePostReportActivity
...
+
+    override fun onCreate() {
+        super.onCreate()
+        container = AppDBContainer(this)
+    }
 }
 
+fun CreationExtras.jerboaApplication(): JerboaApplication =
+    (this[ViewModelProvider.AndroidViewModelFactory.APPLICATION_KEY] as JerboaApplication)
+
```
yate, line 88: This is referenced in view model factories here https://github.com/dessalines/jerboa/pull/973/files#diff-41754fa505a06441b96271ba020ae2e014472578466e14a44fd7ce01a6c578f4R37, based on the compose room workshop again https://github.com/google-developer-training/basic-android-kotlin-compose-training-inventory-app/blob/room/app/src/main/java/com/example/inventory/ui/AppViewModelProvider.kt#L65

app/src/main/java/com/jerboa/MainActivity.kt
```diff
@@ -115,10 +108,17 @@ class MainActivity : AppCompatActivity() {
...
             }
 
-            LaunchedEffect(Unit) {
-                fetchInitialData(accountSync, siteViewModel)
+            val accountViewModel: AccountViewModel = viewModel(factory = AccountViewModelFactory.Factory)
+
+            val account = getCurrentAccount(accountViewModel)
+
+            LaunchedEffect(account) {
+                if (account == null || account.id != -1) {
```
yate, line 116: There was an occasional race condition type issue where `getCurrentAccount` returns an initial value of `null` because it hasn't gotten the data from the database yet. Since there was no initial state set on the `observeAsState` method here https://github.com/dessalines/jerboa/pull/973/files#diff-697367ddff23d571ded643b597f312aeba4b29669b058e9edb03edf9bb030d92R14, it was making two calls to `fetchInitialData`. Once with the `null` initial value, then again with the actual data from the database. This caused two site data calls to go out, depending on which one returned first, the user data in the siteViewModel could set incorrectly set to the wrong value.

As a solution to this, I set an initial state value in  `observeAsState`, so I know if the account id is -1, we haven't actually gotten any data yet from Room, so there is no need to call `fetchInitialData`. If there is no account in Room, it will return `null` and work like before. 

Let me know if there's a better way to handle something like this

app/src/main/java/com/jerboa/MainActivity.kt
```diff
@@ -165,17 +165,14 @@ class MainActivity : AppCompatActivity() {
                     ) {
                         LoginActivity(
                             navController = navController,
-                            accountViewModel = accountViewModel,
```
yate, line 168: I saw in the workshop, the viewModel parameters weren't passed in directly, they used the factory in the composable instead, like https://github.com/dessalines/jerboa/pull/973/files#diff-6ca2c69ae054a88358d16c98f47efd02bb16eae48a09295b4db9bd5e99ba8e76R25.

app/src/main/java/com/jerboa/db/dao/AccountDao.kt
```diff
@@ -0,0 +1,34 @@
...
+import androidx.room.Query
+import androidx.room.Update
+import com.jerboa.db.entity.Account
+
+@Dao
+interface AccountDao {
+    @Query("SELECT * FROM account")
+    fun getAll(): LiveData<List<Account>>
+
+    @Query("SELECT * FROM account where current = 1 limit 1")
```
yate, line 17: Made a query to get the current account instead of getting all and fetching the first here https://github.com/dessalines/jerboa/pull/973/files#diff-697367ddff23d571ded643b597f312aeba4b29669b058e9edb03edf9bb030d92L24

yate:
Thanks @dessalines, it should be ready. I added some initial comments. It's mostly large because of the reorganization of the db package, but I also changed the way the view models/factories were getting initialized which could all be split into a separate PR, or removed, if you prefer.

app/src/main/java/com/jerboa/MainActivity.kt
dessalines: Seems fine to me.

app/src/main/java/com/jerboa/ui/components/post/create/CreatePostActivity.kt
```diff
@@ -52,7 +53,7 @@ data class MetaDataRes(val title: String?, val loading: Boolean)
 
 @Composable
 fun CreatePostActivity(
-    accountViewModel: AccountViewModel,
+    accountViewModel: AccountViewModel = viewModel(factory = AccountSettingsViewModelFactory.Factory),
```
dessalines, line 56: With regards to all of these, I think I'd much prefer them not having defaults, and filling this from MainActivity. Its much more explicit that way, and I've tried to avoid defaults, especially for viewmodels, wherever possible.

app/src/main/java/com/jerboa/MainActivity.kt
```diff
@@ -133,7 +133,7 @@ class MainActivity : AppCompatActivity() {
                     appSettings.usePrivateTabs,
                 )
 
-                ShowChangelog(appSettingsViewModel = appSettingsViewModel)
+                ShowChangelog()
```
dessalines, line 136: Stuff like this scares me, it seems like it has no actual dependencies now, when in reality the appSettings does need to be injected.

dessalines (commented):
Looks good, but please remove the defaults, and make the requirement explicit again.

yate:
@dessalines Thanks, I removed the ViewModel parameter defaults.

MV-GH (approved):
Rather have this one merged before the others that modify appsettings

dessalines (approved):
K I just tested this locally, and it works great. Thx!

PR #975, opened 2023-07-05
Fix recompositions due to autofill

MV-GH:
Fix recompositions issues due to previous autofill, it used `composed` causing constant recompositions on any change.

before:

https://github.com/dessalines/jerboa/assets/67873169/a6115f33-5e07-444b-a04c-d94b13674ff6

after:

https://github.com/dessalines/jerboa/assets/67873169/e47f39ad-8b9a-4e2c-859c-2203a6acf418

After only the button and the column are constantly recompositioning, i looked into it and they login function is a lambda that uses AccountViewmodel (which is unstable to due to AccountRepo) that is unstable, causing the lambda to be unstable and causing the button and column to be unstable.

the AccountViewmodel stableness might be fixed due to #973.


yate:
@MV-GH AccountViewModel is still unstable on my branch
```
unstable class AccountViewModel {
  unstable val repository: AccountRepository
  unstable val currentAccount: LiveData<Account?>
  unstable val allAccounts: LiveData<List<Account>>
  <runtime stability> = Unstable
}
```

MV-GH:
I was more thinking about it injecting the viewmodels instead of passing them

twizmwazin (approved):

PR #980, opened 2023-07-06
Add separate screen for ImageViewer

### LINKED ISSUE #966: Back and download buttons difficult to click due to status bar overlapping them (when viewing images)
Jerboa v0.0.38

Android 13, OnePlus 7 Pro (Pixel Experience)

Status bar covers back button and download button at the top of the screen, making it very difficult to click these buttons. 

**To Reproduce**

1. Open any image in Jerboa
2. Tap to expose back button and download button
3. Try to click either button (and note the majority of the button is unclickable due to the status bar).

MV-GH:
All though #966 could simply be fixed by doing `statusBarsPadding()` for android 13. I wanted to fix the the different levels of positioning of the buttons across the supported android versions.

I have attempted many things to get it work for a dialog. But i could never get it to work for all Android versions.

Examples:
Existing behaviour on older than android 13. The dialog overlaps half the status bar. Only doesn't overlap the buttons due to the padding on the buttons. Possibly will overlap on phones with bigger status bar due to a bigger notch.
studio64_jsS9PPjaqd

Messing with fitsSystemWindows &  WindowCompat.setDecorFitsSystemWindows, etc
you get this result on some android versions
studio64_tp90ORfXYT

After wasting several hours on this, it is in my opinion you should never bother with fullscreen dialogs. Instead I turned it into a screen/activity. And now it not only consistent with positioning across all android versions. It is placed exactly where other back buttons would be placed.

How it looks now
studio64_eMGN9EIJ26

It retains all existing functionality.

Fixes #966


twizmwazin:
Is there any way with this approach to retain the transparency effect?

MV-GH:
Not sure but I'll take a look. Didn't really think about that.

MV-GH:
I did not find a way to do this yet. But I took a look at the other apps and none allow you to see through them. So instead I can make it always completely black

dessalines:
I'd be fine with a black background also. Whatever is easier to maintain and less buggy.

MV-GH:
Improved the UX of the screen viewer

https://github.com/dessalines/jerboa/assets/67873169/2bce13c2-5085-4999-bd31-05747cff732c

dessalines (approved):
I tested and it works great.

lawrence0arabia:
Great stuff @MV-GH 🔥

PR #984, opened 2023-07-07
Add contenttypes to lazy columns

MV-GH:
https://developer.android.com/jetpack/compose/lists#content-type

Adds content types to the lazy columns. From my limited testing, I did not notice any significant performance changes.


app/src/main/java/com/jerboa/ui/components/community/list/CommunityList.kt
```diff
@@ -76,6 +76,7 @@ fun CommunityListings(
         items(
             communities,
             key = { it.community.id },
+            contentType = { "communitylink_${it.counts.users_active_month == 0}" },
```
dessalines, line 79: just needs to be "communityLink"

dessalines (commented):
According to the docs, this is just supposed to refer to the type of item, for lists which have different types. So it just needs to be "comment", "post", etc.

app/src/main/java/com/jerboa/ui/components/community/list/CommunityList.kt
dessalines: bump

app/src/main/java/com/jerboa/ui/components/comment/CommentNode.kt
```diff
@@ -233,7 +233,7 @@ fun LazyListScope.commentNodeItem(
         .commentView.counts.child_count > 0 && !isFlat
 
     increaseLazyListIndexTracker()
-    item(key = commentId) {
+    item(key = commentId, contentType = { "contentNodeItemL${node.depth}" }) {
```
MV-GH, line 236: Eh it's just this one that worries me, I ll leave this one out for now and when I get some more time I ll try to get a microbench for this, bc the macrobench currently is not showing any significant perf changes. Same was for the scrolling with baselines/without but from personal expercience the baselines make a bigger impact than they are showing from the tests. So they aren't very reliable imo.

dessalines: Gotcha. Yep I'm guessing these aren't going to really matter until we have combined lists, like putting all the notification types (messages, comment replies, and mentions), in a single LazyList.

dessalines (approved):
Seems good, not sure why build is failing, I'll try a restart.

PR #990, opened 2023-07-07
Adding spoiler tag support

ZJouba:
Addressing: #627 
Adding `::: spoiler` tag support for Jerboa

https://github.com/dessalines/jerboa/assets/42601911/9b1649d6-ef1f-4f7a-a9f5-8326fcaebc9d

Nested tags are not supported and parsing can be weird if the syntax isn't correct
I'll wait for #734 to be merged, update and then put mine up for review

Post used for testing: https://voyager.lemmy.ml/post/1053


MV-GH:
Looks great but I would probably put that `class SpoilerPlugin` in Utils

twizmwazin:
I disagree, we already have a markwon plugin in this file and we shouldn't be dumping everything in Utils if there is a better place for it. If the custom plugins become too much to contain in this file, it'd be better to make a package for them.

MV-GH:
I don't mean the utils.kt, the package Utils as a class

ZJouba:
@MV-GH I hope I understood you correctly with this change 🤣

dessalines (commented):
I just tested this, and it works great, but unfortunately the `click to collapse comment` overrides it. Although you can long-click to reveal the spoiler.

@MV-GH @twizmwazin I'm not sure if we should merge as is, and tackle that one separately.

ZJouba:
Hey @dessalines, thanks for the 👀

I'm happy to tackle it pronto, I just need a teensy bit of guidance. Does Compose have a 'preventDefault` option?

twizmwazin:
Is the issue that tapping to reveal conflicts with tapping to collapse a comment? If so, the desired behavior I think is that for text that is intractable, such as a spoiler, that text's behavior should take precedence over collapsing the comment. There is space in the comment header that will always be empty/non-interactable to collapse a comment even if the entire body is interactable. For implementation, it would be intuitive to me if setting `Modifier.clickable(), then that composable should consume the click event rather than the parent. If that's not the case then I'm unsure.

ZJouba:
The spoiler tag is only a `ClickableSpan` and the entire comment body has the `toggleExpanded` on click listener. I'm not sure how to have it stop after the spoiler. Because both onClicks get called. Its not disregarding the spoiler onClick

app/src/main/java/com/jerboa/util/MarkwonSpoilerPlugin.kt
```diff
@@ -0,0 +1,139 @@
...
+                spanned.setSpan(
+                    spoilerTitle,
+                    spoilerStart,
+                    spoilerStart + spoilerTitle.length,
+                    Spanned.SPAN_EXCLUSIVE_EXCLUSIVE,
+                )
+
+                val wrapper = object : ClickableSpan() {
+                    override fun onClick(p0: View) {
+                        textView.cancelPendingInputEvents()
```
ZJouba, line 99: Turns out it's as easy as this

ZJouba:
If everyone agrees, it's easy to remove

ZJouba:
I can add a `preview` boolean to the plugin which is set when previewMarkwon is used which will remove the onClick functionality

https://github.com/dessalines/jerboa/assets/42601911/404103fd-b5b4-4d31-8ffe-1b7eedbcf7ba

twizmwazin (commented):
LGTM once interaction is disabled for preview. Your approach sounds good.

app/src/main/java/com/jerboa/util/MarkwonSpoilerPlugin.kt
```diff
@@ -16,8 +16,7 @@ import io.noties.markwon.image.AsyncDrawableScheduler
 data class SpoilerTitleSpan(val title: CharSequence)
 class SpoilerCloseSpan
 
-class MarkwonSpoilerPlugin : AbstractMarkwonPlugin() {
-
+class MarkwonSpoilerPlugin(val preview: Boolean) : AbstractMarkwonPlugin() {
```
twizmwazin, line 19: Instead of `preview` could we perhaps use `enableInteraction`?

app/src/main/java/com/jerboa/util/MarkwonSpoilerPlugin.kt
```diff
@@ -0,0 +1,140 @@
...
+                spanned.setSpan(
+                    spoilerTitle,
+                    spoilerStart,
+                    spoilerStart + spoilerTitle.length,
+                    Spanned.SPAN_EXCLUSIVE_EXCLUSIVE,
+                )
+
+                val wrapper = object : ClickableSpan() {
+                    override fun onClick(p0: View) {
+                        if (!enableInteraction) {
```
twizmwazin, line 98: Does the value of this need to be flipped too?

ZJouba: Yeah, sorry about that

twizmwazin (approved):

PR #991, opened 2023-07-07
Add drawer to all main tabs and general drawer changes

### LINKED ISSUE #914: Make NavigationDrawer accessible from all top level views

**Describe the feature request below**
At the moment the Navigation Drawer is onl accessible from the home view.
It would be more logical to have access from alle top level views, namely:

- Communities / "Search"
- Inbox
- Bookmarks
- Profile

I'd also argue that the Navigation Drawer should be accessible from a single community but you could also say that a single community is in fact a sub page of Communities.
This should be discussed, especially because that would raise the question what should happen the back button in the top bar of a single community.

MV-GH:
- Add drawer to all the main tabs. 
- Made it so that it doesn't display the nav also in the drawer if the buttomNavBar is visible.
- Removed lots of duplicate logic surrounding the navBottom. 
- Also fixed switching being tabs that back still would return sometimes. Same fix as #988.
- Removed duplicate string resources. 
- Also moved drawer stuff into its own package.

https://github.com/dessalines/jerboa/assets/67873169/1a66223a-5815-4de6-8bd9-23d18584eecd

Fixes #914 

@nahwneeth What's the reason for the duplicate navhost in BottomNavActivity? I think it can be removed.


nahwneeth:
The second nav host only has pages that are on the bottom nav (home, search, inbox, saved, profile). And the bottom nav bar is outside this second nav host. This way we can avoid having the bottom nav on every page (which was how it was previously).

nahwneeth:
Then the bottom nav has to be duplicated on those 5 pages right? Coz what I found was the bottom nav can only be specified on the scaffold, and I couldn't come up with a better way than using a second nav host within a Scaffold.

MV-GH:
I think i understand what you mean. If i move the drawer stuff and navBottomBar to the root. It would mean all pages would have the the drawer and navBottomBar? But maybe it can still be done

app/src/main/java/com/jerboa/ui/components/drawer/Drawer.kt
```diff
@@ -0,0 +1,414 @@
...
+import com.jerboa.ui.components.person.PersonName
+import com.jerboa.ui.theme.DRAWER_BANNER_SIZE
+import com.jerboa.ui.theme.LARGE_PADDING
+import com.jerboa.ui.theme.SMALL_PADDING
+import com.jerboa.ui.theme.XL_PADDING
+import com.jerboa.ui.theme.muted
+
+@Composable
+fun Drawer(
+    siteRes: ApiState<GetSiteResponse>,
```
dessalines, line 71: Just have this pass in `myUserinfo?` . I've tried hard to avoid passing the ApiStates to anything but the top-level screens, and especially not to the lower-level composables.

app/src/main/java/com/jerboa/ui/components/drawer/DrawerActivity.kt
```diff
@@ -0,0 +1,103 @@
...
+fun MainDrawer(
+    siteViewModel: SiteViewModel,
+    navController: NavController,
+    accountViewModel: AccountViewModel,
+    homeViewModel: HomeViewModel,
+    scope: CoroutineScope,
+    drawerState: DrawerState,
+    onSelectTab: (NavTab) -> Unit,
+    blurNSFW: Boolean,
+    showBottomNave: Boolean,
```
dessalines, line 30: Nave .

app/src/main/java/com/jerboa/ui/components/drawer/Drawer.kt
MV-GH: That was there before, i just moved it to a new location, but i have changed it.

MV-GH:
@dessalines Fixed

dessalines (approved):

PR #871, opened 2023-06-24
Add local search history to the community list screen

wkoomson:
This change adds a search history Dao/Repository/ViewModel so that we can save user searches in Room, and display them on the community list screen. Users can click the history item to search it, or click the "X" button to delete it.

This is for issue  #811

PR #995, opened 2023-07-08
Add toggle in settings to hide descriptions in navbar

Sir-Photch:
Addresses #818


app/src/main/res/values/strings.xml
```diff
@@ -180,6 +180,7 @@
     <string name="look_and_feel_secure_window">Prevent Screenshots</string>
     <string name="look_and_feel_show_action_bar_for_comments">Show action bar by default for comments</string>
     <string name="look_and_feel_show_navigation_bar">Show navigation bar</string>
+    <string name="look_and_feel_nav_bar_only_icons">Icon-only navigation bar</string>
```
twizmwazin, line 183: "Show text descriptions in navigation bar" is probably better.

app/src/main/java/com/jerboa/db/AppDB.kt
```diff
@@ -121,6 +121,11 @@ data class AppSettings(
         defaultValue = "1",
     )
     val blurNSFW: Boolean,
+    @ColumnInfo(
+        name = "nav_bar_only_icons",
+        defaultValue = "0",
+    )
+    val navBarOnlyIcons: Boolean,
```
twizmwazin, line 128: `showTextDescriptionsInNavbar`

app/schemas/com.jerboa.db.AppDB/17.json
```diff
@@ -64,7 +64,7 @@
...
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`id` INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, `font_size` INTEGER NOT NULL DEFAULT 16, `theme` INTEGER NOT NULL DEFAULT 0, `theme_color` INTEGER NOT NULL DEFAULT 0, `viewed_changelog` INTEGER NOT NULL DEFAULT 0, `post_view_mode` INTEGER NOT NULL DEFAULT 0, `show_bottom_nav` INTEGER NOT NULL DEFAULT 1, `show_collapsed_comment_content` INTEGER NOT NULL DEFAULT 0, `show_comment_action_bar_by_default` INTEGER NOT NULL DEFAULT 1, `show_voting_arrows_in_list_view` INTEGER NOT NULL DEFAULT 1, `show_parent_comment_navigation_buttons` INTEGER NOT NULL DEFAULT 1, `navigate_parent_comments_with_volume_buttons` INTEGER NOT NULL DEFAULT 0, `use_custom_tabs` INTEGER NOT NULL DEFAULT 1, `use_private_tabs` INTEGER NOT NULL DEFAULT 0, `secure_window` INTEGER NOT NULL DEFAULT 0, `blur_nsfw` INTEGER NOT NULL DEFAULT 1, `nav_bar_only_icons` INTEGER NOT NULL DEFAULT 0)",
```
twizmwazin, line 67: Please make a new version and a migration rather than edit the existing schema.

gradle.properties
```diff
@@ -1,25 +1,19 @@
-# Project-wide Gradle settings.
```
twizmwazin, line 1: This file shouldn't be changed at all in this PR

twizmwazin (changes requested):

app/schemas/com.jerboa.db.AppDB/17.json
Sir-Photch: I've added a migration to V18, and that generated a new schema.json. Please check

app/src/main/java/com/jerboa/ui/components/settings/lookandfeel/LookAndFeelActivity.kt
```diff
@@ -224,6 +227,15 @@ fun LookAndFeelActivity(
                     },
                     onCheckedChange = { updateAppSettings() },
                 )
+                AnimatedVisibility(visible = showBottomNavState.value) {
```
twizmwazin, line 230: I'm not sure how I feel about making a setting visible based on another setting. I think this should just be always shown.

app/schemas/com.jerboa.db.AppDB/18.json
```diff
@@ -2,7 +2,7 @@
   "formatVersion": 1,
   "database": {
     "version": 18,
-    "identityHash": "214278f3dfc312f44830b58678afc3eb",
+    "identityHash": "9a881a3925f19533ba5f6d339f9d0461",
```
twizmwazin, line 5: It seems like this whole file shouldn't exist yet. @MV-GH do you know what's up here?

twizmwazin (changes requested):

app/schemas/com.jerboa.db.AppDB/18.json
MV-GH: Yeah that was a mistake, it was added already by accident. It was a leftover from testing #871

app/src/main/java/com/jerboa/ui/components/settings/lookandfeel/LookAndFeelActivity.kt
Sir-Photch: Are you sure? This setting has no effect when the navbar is disabled. So it makes sense to hide it imo. If you insist, I will remove that.

twizmwazin: It has no effect, but I am concerned that hiding it would be confusing and frustrating to users who are trying to find that specific option. It would make more sense perhaps to disable/gray out the option so that users can at least see that it would be there than hiding it entirely.

Sir-Photch: I guess that's a compromise.

twizmwazin (approved):

PR #998, opened 2023-07-08
Add confirmation on exit

MV-GH:
Adds a confirmation when trying to exit the app, to prevent accidental exits. I read somewhere that someone wanted this. Can't find the issue back though.

https://github.com/dessalines/jerboa/assets/67873169/42893df2-9373-4005-b5cd-967efe36473b


twizmwazin:
Instead of using toasts, could we use a dialog? Other apps with a similar feature will pop up a "are you sure you want to exit?" dialog, usually with a "don't ask me again" checkbox option, and an exit and cancel button.

MV-GH:
I just copied the behaviour of what Boost did.
I would prefer a toast much over a dialog. I don't want a intrusive dialog each time I am trying to exit the app.
Adding a option in settings is possible but does add some complexity.


https://github.com/dessalines/jerboa/assets/67873169/4ce0f13a-6f15-414b-8ebb-ae4924f1d71d


Boost options but honestly i don't think anyone uses the dialog thing.
image

twizmwazin:
I might have been the one user of that dialog option, lol.

app/src/main/res/values/strings.xml
```diff
@@ -333,4 +333,5 @@
     <string name="dialogs_top_nine_month">Top Nine Months</string>
     <string name="post_listing_view_source">View Source</string>
     <string name="post_listing_view_original">View Original</string>
+    <string name="back_warning">Repeat action to exit</string>
```
twizmwazin, line 336: Are there are other actions besides pressing the back button/gesture equivalent that we would need to be concerned about? If not, "Press back again to exit" sounds more natural to me. Otherwise LGTM.

MV-GH: Yeah I didn't want to put tap because of the back gesture, but that could work too.

dessalines: Ya I also like "Press back again to exit"

dessalines (commented):
To break the tie 😆 , I think the simple toast works fine, and an extra setting is probably overkill.

MV-GH:
I was going to add the setting, it can always be expanded to add the dialog.

MV-GH:
Added all the options

https://github.com/dessalines/jerboa/assets/67873169/e70d8bc9-f2b3-4947-a903-113df747d17d

twizmwazin (approved):
Love it

PR #1001, opened 2023-07-09
Make the FAB configurable

camporter:
Leaving in draft for thoughts.

To start off with, modes include:
* creating a post (the current default)
* scrolling to the top

Positions include:
* End (Right, current default)
* Center

https://github.com/dessalines/jerboa/assets/494446/fff197e6-756a-4a91-bf29-cff20157981d

Might need an additional place to create a post from such as the drawer.

### DIFF
--- a/app/src/main/java/com/jerboa/db/AppDB.kt
+++ b/app/src/main/java/com/jerboa/db/AppDB.kt
@@ -121,6 +121,16 @@ data class AppSettings(
         defaultValue = "1",
     )
     val blurNSFW: Boolean,
+    @ColumnInfo(
+        name = "fab_action_mode",
+        defaultValue = "0",
+    )
+    val fabActionMode: Int,
+    @ColumnInfo(
+        name = "fab_position_mode",
+        defaultValue = "0",
+    )
+    val fabPositionMode: Int,
 )
 
 val APP_SETTINGS_DEFAULT = AppSettings(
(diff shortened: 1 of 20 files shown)


app/src/main/java/com/jerboa/db/AppDB.kt
```diff
@@ -121,6 +121,16 @@ data class AppSettings(
         defaultValue = "1",
     )
     val blurNSFW: Boolean,
+    @ColumnInfo(
+        name = "fab_action_mode",
+        defaultValue = "0",
+    )
+    val fabActionMode: Int,
+    @ColumnInfo(
+        name = "fab_position_mode",
```
dessalines, line 130: Rename this `fab_position`

app/src/main/java/com/jerboa/db/AppDB.kt
```diff
@@ -121,6 +121,16 @@ data class AppSettings(
         defaultValue = "1",
     )
     val blurNSFW: Boolean,
+    @ColumnInfo(
+        name = "fab_action_mode",
```
dessalines, line 125: This one doesn't make too much sense to me, especially since we might add different actions to different pages. Since there won't be a "universal" FAB action on all pages, it probably doesn't make sense to have this.

The post page, like boost, could have a multi-action FAB, to account for extra actions.

camporter: Sure, we could add a bottom sheet on long press to customize the FAB per page type instead.

I had looked into using something like https://github.com/leinardi/FloatingActionButtonSpeedDial but I was concerned it might not be 'material 3'-ey enough. There's a branch of it that is designed for material 3, but it seems like some other way of customizing the FAB, or changing the FAB behavior, or adding more actions would be more appropriate with a bottom sheet or menu.

dessalines: Ya, I'm surprised android doesn't already have something set up for multiple floating actions / expanding.

dessalines:

PR #1015, opened 2023-07-11
Save comment/saved/profile/inbox sort options

### LINKED ISSUE #921: Default comment sorting

Add the ability to set default comment sorting the same way you can set default post sorting. As it is in 0.0.37, comment sorting resets to hot on every post. In 0.0.36, it was possible to set it to top, and it would stay that way for every post until the app was restarted. Now it resets to hot every time. I'm not sure if this is a bug, but default comment sorting options would be nice to have either way.

### LINKED ISSUE #976: Default comment sort option

I'd like a setting to change the default sort type for comments on a post.  Every time I open a post, the comments are sorted by "Hot" but I normally prefer "Top".

MV-GH:
Save comment/saved/profile/inbox sort options. These should be the remaining options that were not retained.

Also added missing sort type `old`

Fixes #976
Fixes #921

### DIFF
--- a/app/src/main/java/com/jerboa/db/AppDB.kt
+++ b/app/src/main/java/com/jerboa/db/AppDB.kt
@@ -131,8 +131,6 @@ data class AppSettings(
         defaultValue = "1",
     )
     val backConfirmationMode: Int,
-
-    // TODO migrations
     @ColumnInfo(
         name = "comment_sorting_mode",
         defaultValue = "0",
@@ -140,17 +138,17 @@ data class AppSettings(
     val commentSortingMode: Int,
     @ColumnInfo(
         name = "profile_sorting_mode",
-        defaultValue = "0",
+        defaultValue = "2",
     )
     val savedSortingMode: Int,
     @ColumnInfo(
         name = "saved_sorting_mode",
-        defaultValue = "0",
+        defaultValue = "2",
     )
     val profileSortingMode: Int,
     @ColumnInfo(
         name = "inbox_unread",
-        defaultValue = "0",
+        defaultValue = "1",
     )
     val inboxMode: Int,
 )
@@ -175,9 +173,9 @@ val APP_SETTINGS_DEFAULT = AppSettings(
     showTextDescriptionsInNavbar = true,
     backConfirmationMode = 1,
     commentSortingMode = 0,
-    savedSortingMode = 0,
-    profileSortingMode = 0,
-    inboxMode = 0,
+    savedSortingMode = 2,
+    profileSortingMode = 2,
+    inboxMode = 1,
 )
 
 @Dao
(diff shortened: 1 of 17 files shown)


app/src/main/java/com/jerboa/db/AppDB.kt
```diff
@@ -131,6 +131,26 @@ data class AppSettings(
         defaultValue = "1",
     )
     val backConfirmationMode: Int,
+    @ColumnInfo(
+        name = "comment_sorting_mode",
+        defaultValue = "0",
+    )
```
dessalines, line 137: Whoa, probably should asked about this one before going ahead with it.

- These should be stored in lemmy's back end, otherwise no other app / front end will be able to use them. 
  - Same goes for default sorts for specific communities, which might also need back-end settings.
- So now we have sorts stored in 3 (maybe 4) different places: 
  - Your lemmy backend profile
  - The jerboa `Account` table (which is where these should live, for fast initial app loading)
  - And now also the jerboa `AppDB` table for some reason, which overrides each account?

MV-GH: I did ask about this in matrix chat. If these are stored on the backend, we would need to add them first there. And then we would also have the compatibility issue again that older lemmy servers won't have these. Further anonymous account wont be able to use these either. This would only add the sorting options that don't exist yet. I have also been contemplating to make the anon account, an actual "special" account that is stored in DB. This would allow us to have anon all share the features of a logged in account. All global settings now can be saved per account based, such as the "censor NSFW".  This would a huge migration but better earlier than later.

MV-GH:
@dessalines What's the way forward now? Add these to the backend and then implement support for them? If that is the case I wont be able to help. As I don't know rust or the lemmy codebase.

twizmwazin:
IMO we should not be letting features go unimplemented in Jerboa if they can be implemented locally without major tradeoffs. In the future if Lemmy supports this we can then adapt the feature to that, though even in that case I think an argument could be made for keeping a local preference that may be different from the account preference.

dessalines:
If there isn't one already, create an issue on lemmy to discuss potential "user-saveable" default sorts and view types. 

These could be for specific views / pages, like the ones you proposed here, or they could be for specific communities. 

I could see it getting complicated pretty quickly, because then there has to be some precedence order for defaults (which already exists because there are admin sort defaults, and user sort defaults which override those).

I won't have time to work on it, but there are plenty of other contributors now that potentially could. But I see this being a pratty popular feature request because a lot of reddit apps hacked this feature on in their own way, because reddit didn't add it.

dessalines:
Stale PR. Those additional sort options should be added in the back-end before we save them here, in the same way that we have `defaultSortType`, etc.

PR #1022, opened 2023-07-12
Fix and update of the Ukrainian translation

Digharatta:
Thank you for implementing the Ukrainian translation!
First live tests revealed that some strings need to be fixed, so I fixed them and also updated some other strings.


app/src/main/res/values-uk/strings.xml
```diff
@@ -11,11 +11,11 @@
     <string name="account_settings_all">Усі</string>
     <string name="account_settings_avatar">Аватар</string>
     <string name="account_settings_banner">Банер</string>
-    <string name="account_settings_bio">Біографія</string>
+    <string name="account_settings_bio">Невеличкий життєпис/string>
```
MV-GH, line 14: `/string>` is missing `<`

MV-GH (changes requested):
Looks good otherwise than that small issue

app/src/main/res/values-uk/strings.xml
Digharatta: Thank you, fixed it.

dessalines (approved):

MV-GH (approved):

app/src/main/res/values-uk/strings.xml
```diff
@@ -156,14 +156,14 @@
...
-    <string name="login_totp">Введіть токен двофакторної автентифікації, якщо її увімкнено</string>
-    <string name="login_password">Пароль</string>
-    <string name="login_view_model_incorrect_login">Невірний логін</string>
+    <string name="_back">Повернутися</string>
+    <string name="_email_or_username">Адреса електронної пошти або ім\'я користувача</string>
+    <string name="_instance">Примірник (інстанс)</string>
+    <string name="_instance_placeholder">наприклад: lemmy.ml</string>
+    <string name="_">Увійти</string>
+    <string name="_totp">Введіть токен двофакторної автентифікації, якщо її увімкнено</string>
+    <string name="_password">Пароль</string>
```
MV-GH, line 165: These names are not the same

MV-GH (changes requested):

MV-GH (approved):

PR #1024, opened 2023-07-12
Update italian translations

andscape-dev:


app/src/main/res/values-it/strings.xml
```diff
@@ -119,7 +127,10 @@
     <string name="dialogs_top_month">Top del mese</string>
     <string name="dialogs_top_week">Top della settimana</string>
     <string name="dialogs_top_year">Top dell\'anno</string>
+    <string name="dialogs_old">Vecchi</string>
     <string name="dialogs_unread">Non letti</string>
+    <string name="dialogs_server_version_outdated">La versione del server (%1$s) è inferiore alla minima supportata (%2$s).\\n\\nInforma i tuoi amministratori ed effettua il login su un\'altra istanza, o usa quella di default anonimamente.</string>
+    <string name="dialogs_server_version_outdated_short">La versione del server (%1$s) è troppo bassa.</string>
```
andreaippo, line 133: I wonder if `vecchia` would be a better fit than `bassa`, given the context?

app/src/main/res/values-it/strings.xml
```diff
@@ -149,6 +160,7 @@
     <string name="login_instance">Istanza</string>
     <string name="login_instance_placeholder">ad es.: lemmy.ml</string>
     <string name="login_login">Login</string>
+    <string name="login_totp">Inserisci il token 2FA se è abilitato</string>
```
andreaippo, line 163: What about:
```suggestion
    <string name="login_totp">Inserisci il token 2FA, se abilitato</string>
```
?

app/src/main/res/values-it/strings.xml
```diff
@@ -159,16 +171,28 @@
     <string name="look_and_feel_post_view_card">Scheda</string>
     <string name="look_and_feel_post_view_list">Lista</string>
     <string name="look_and_feel_post_view_small_card">Scheda piccola</string>
+    <string name="look_and_feel_secure_window">Previeni screenshot</string>
```
andreaippo, line 174:
```suggestion
    <string name="look_and_feel_secure_window">Impedisci screenshot</string>
```
?

app/src/main/res/values-it/strings.xml
```diff
@@ -159,16 +171,28 @@
...
     <string name="look_and_feel_theme_system">Di sistema</string>
     <string name="look_and_feel_theme_system_black">Di sistema nero</string>
+    <string name="look_and_feel_theme_color_dynamic">Dinamico</string>
+    <string name="look_and_feel_theme_color_beach">Spiaggia</string>
+    <string name="look_and_feel_theme_color_blue">Blu</string>
+    <string name="look_and_feel_theme_color_crimson">Cremisi</string>
+    <string name="look_and_feel_theme_color_green">Verde</string>
+    <string name="look_and_feel_theme_color_grey">Grigio</string>
+    <string name="look_and_feel_theme_color_pink">Rosa</string>
+    <string name="look_and_feel_theme_color_purple">VIola</string>
```
andreaippo, line 194:
```suggestion
    <string name="look_and_feel_theme_color_purple">Viola</string>
```

app/src/main/res/values-it/strings.xml
```diff
@@ -262,4 +297,44 @@
...
+    <string name="post_listing_view_source">Mostra markdown</string>
+    <string name="post_listing_view_original">Mostra originale</string>
+    <string name="look_and_feel_show_text_descriptions_in_navbar">Mostra descrizioni nella barra di navigazione</string>
+    <string name="person_iconBot">È un bot</string>
+    <string name="download">Download</string>
+    <string name="back_confirmation">Premi indietro nuovamente per uscire</string>
+    <string name="no_back_confirmation">Non chiedere conferma per uscire</string>
+    <string name="press_again_confirmation">Premi indietro due volte per uscire</string>
+    <string name="confirm_exit">Conferma azione di uscita</string>
+    <string name="back_dialog">Mostra un dialogo di conferma per uscire</string>
```
andreaippo, line 337: I think that the correct translation for a dialog would be the somewhat verbose `finestra di dialogo`
What about replacing `dialogo` with `popup`, instead?
Or, more freely, `Richiedi conferma per uscire`?

app/src/main/res/values-it/strings.xml
```diff
@@ -54,19 +54,25 @@
     <string name="comment_edit_type_your_comment">Digita il tuo commento</string>
     <string name="comment_mention_node_block">Blocca %1$s</string>
     <string name="comment_mention_node_copy_permalink">Copia permalink</string>
+    <string name="comment_mention_node_copy_comment">Copia testo del commento</string>
     <string name="comment_mention_node_go_to">Vai a</string>
     <string name="comment_mention_node_permalink_copied">Permalink copiato</string>
+    <string name="comment_mention_node_comment_copied">Testo copiato</string>
     <string name="comment_mention_node_report_comment">Segnala commento</string>
     <string name="comment_mention_node_view_source">Mostra markdown</string>
+    <string name="comment_next_parent">Vai al prossimo commento padre</string>
```
andreaippo, line 63: I wonder if we could find a better word for `parent`.
`superiore`could be an option, but in some context it may be mistaken for "the one above".
Although, thinking of it, in file system navigation this stuff gets usually translated that way (e.g. vai alla cartella superiore).
In this case it could become:
```suggestion
    <string name="comment_next_parent">Vai al commento superiore successivo</string>
```

app/src/main/res/values-it/strings.xml
andscape-dev: Done

app/src/main/res/values-it/strings.xml
andscape-dev: I like `Richiedi conferma per uscire`, done

app/src/main/res/values-it/strings.xml
andreaippo: I agree about the potential confusion indeed. There's no good translation I can think of, so maybe let's leave your original translation and call it a day.
I had also thought of `commento radice` for a sec, but technically speaking it would be a bit different, plus as you imply it's also not a very common concept outside of data structures (although most of lemmy users ATM are probably geeks running Linux anyway, so...)

andscape-dev: Yeah I can't think of a better translation either. I liked your `successivo` better than `prossimo` so I updated it, but I guess let's leave `padre` in there until someone can come up with a better idea.

Also, I think this text is not visible in the UI. I'm guessing it's for blind / low vision users. Not to throw them under the bus, but probably almost nobody will read or hear it.

andscape-dev:
@andreaippo thanks for the review and suggestions! I think we're ready to merge here

Issue #535, opened 2023-06-11
Mark post as read when scrolled past


humanplayer2:
- [y] Did you check to see if this issue already exists?
- [ y] Is this only a single feature request? Do not put multiple feature requests in one issue.
- [n] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/jerboa

** Describe the feature request below **

The *Hide read posts* option is great. I still see many of the same posts, because I don't open most posts. They are therefore not marked as read. 

Having the option to mark posts as read when scrolled past would be nice.

Issue #403, opened 2023-06-04
Posts are not immediately marked as viewed after viewing them in Home-Screen


oscarnylander:
Expected behaviour is that posts get marked as viewed immediately after having been viewed.

dessalines:
This should come after I have time to work on, and merge, the `autogen_types` branch, which has the `MarkPostAsRead` data class this requires.

oscarnylander:
Sounds great. I won't pick this up, in that case.

dessalines:
Definitely not a good first issue. But there are plenty of other ones you could take.

Schoolkid1:
Thanks for your quick reply. Do you perhaps have any recommendations? I like a good challenge to work on, ultimately I want to learn and upgrade my knowledge.

dessalines:
#425 is a good smaller one to take. 

When I get the time, I'll go through and organize these issues a bit better.

MV-GH:
Hmm confirmed, weird did not touch anything regarding that

MV-GH:
[app-release.zip](https://github.com/dessalines/jerboa/files/12423204/app-release.zip) has the fix

Issue #916, opened 2023-06-29
Make all `List`s into `ImmutableList`


dessalines:
- [x] Did you check to see if this issue already exists?
- [x] Is this only a single feature request? Do not put multiple feature requests in one issue.
- [x] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/jerboa

** Describe the feature request below **

Context: https://github.com/dessalines/jerboa/pull/809

https://issuetracker.google.com/issues/199496149

MV-GH:
Note that only lists passed as argument to a composable function needs to be changed.

dessalines:
This isn't the preferred way anymore.

Issue #807, opened 2023-06-22
Refactor CommunityViewModel, HomeViewModel, InboxViewModel, PersonProfileViewModel


MV-GH:
The VMs: CommunityViewModel, HomeViewModel, InboxViewModel, PersonProfileViewModel all have significant duplicated code such as the pages, posts ... We should refactor that into a ancestor called something like PostsViewModel containing that code.

dessalines:
This one is done, at least for Home and Community

MV-GH:
Most of these changes are still sitting in my draft PR for search expansion. I'll extract them into separate PRs. Atm we still have significant duplicate code that needs to be addressed

PR #1025, opened 2023-07-12
Mark as Read on Open and Scroll

ZJouba:
Addresses: #403, #535 

Mark on scroll works beatifully:

Mark on open doesn't update the LazyColumn on Home. In fact, any update in PostActivity doesn't reflect on Home until refresh:  
NotUpdating

[val read: MutableState<Boolean> = mutableStateOf(false))](https://developer.android.com/codelabs/jetpack-compose-state#11) seems to be the answer but GSON doesn't know how to deserialize MutableState 🤷‍♂️

Any advice would be appreciated


MV-GH:
Instead of passing that through the navigation as a JSONed string, pass the homeviewmodel to the post activity and only update that post. (I think it already does that in some places)

ZJouba:
@MV-GH like so?
```
MarkPostAsRead(
    post_id = it,
    read = true,
    auth = account!!.jwt,
).run {
    postViewModel.markPostAsRead(this)
    homeViewModel.markPostAsRead(this)
}
```
Won't that mark the post as read with the api twice?

MV-GH:
No i was thinking about `updatePost` but this still does a network call, If you your markPostAsRead network call is successfull do a updatePost (make one) that doesnt do the network call and just marks it as read.

ZJouba:
@MV-GH Okay, I think I got it, not 100% sure if its correct though, if a user navigates back to home to quickly, I don't think it will update  
UpdateHome

app/src/main/java/com/jerboa/ui/components/settings/lookandfeel/LookAndFeelActivity.kt
```diff
@@ -71,7 +71,8 @@ fun LookAndFeelActivity(
     )
     val postViewModeState = rememberIntSettingState(settings.postViewMode)
     val showBottomNavState = rememberBooleanSettingState(settings.showBottomNav)
-    val showTextDescriptionsInNavbar = rememberBooleanSettingState(settings.showTextDescriptionsInNavbar)
+    val showTextDescriptionsInNavbar =
+        rememberBooleanSettingState(settings.showTextDescriptionsInNavbar)
```
MV-GH, line 75: Change your column line width to at least 120

app/src/main/java/com/jerboa/ui/components/post/PostListings.kt
```diff
@@ -29,12 +22,11 @@ import com.jerboa.db.entity.Account
 import com.jerboa.isScrolledToEnd
 import com.jerboa.ui.components.common.simpleVerticalScrollbar
 import com.jerboa.ui.theme.SMALL_PADDING
-import kotlinx.collections.immutable.ImmutableList
 import kotlinx.collections.immutable.persistentListOf
 
 @Composable
 fun PostListings(
-    posts: ImmutableList<PostView>,
+    posts: List<PostView>,
```
MV-GH, line 29: Why? See https://github.com/dessalines/jerboa/issues/916

MV-GH (changes requested):
Also, actually i think the viewmodel approach I said earlier, is not the way we should take this. We can open posts from many different locations. And this would only work work with homeviewmodel. It also makes it a global and to make it work with the other viewmodels it would also make them a global. Instead I propose we should use the `navController.ConsumeReturn` Where if you return from a postActivity to a postlisiting. You refresh that post you went to. You will probably have to add this in each relevant activity. Such as PersonProfile, Home and Community

app/src/main/java/com/jerboa/db/AppDBMigrations.kt
```diff
@@ -221,6 +221,47 @@ val MIGRATION_18_19 = object : Migration(18, 19) {
...
+            """
+            INSERT INTO AppSettingsBackup SELECT * FROM AppSettings
+            """,
+        )
+        database.execSQL("DROP TABLE AppSettings")
+        database.execSQL("ALTER TABLE AppSettingsBackup RENAME to AppSettings")
+    }
+}
+
+val MIGRATION_20_21 = object : Migration(19, 20) {
```
MV-GH, line 264: Dont forget to change those 19, 20 numbers

ZJouba:
While I'm busy should I just do this for all activities that navigate to a post?

MV-GH:
Yes

ZJouba:
I'm not super familiar with android architecture, why do we have duplicate functions in each ViewModel, can't they be extracted to a common place?

MV-GH:
Such as? If you mean the nextPage and such yes they can be. See #807

ZJouba:
@MV-GH this approach seems to suffer from race conditions. If I set a breakpoint to inspect, it works. The Home/Community post list is updated and I can see my saved/like/dislike etc. But if I don't have a pause anywhere, the change doesn't reflect until I refresh. Any advice?

MV-GH:
Can you push what you have, I take a look when I have some time

ZJouba:
Okay, so I need to know if this is acceptable. Instead of having to return the post after every action (like, save, dislike) etc. I thought I might just do a request to get the latest post once you navigate back to the HomeActivity, this way, if someone else liked the post while you were navigating back, you'd have the latest info on screen load. 

However, this resulted in the request for the updated post not completing before the composition and thus not showing in the Home posts list. And it's also not updating the post after composition (not sure why). So now I've had to add an "await" of sorts until the latest post is fetched and updated in the list of posts. Is this acceptable?  
await

This is how I update the post in `HomeViewModel.kt`:
```
fun refreshSinglePost(id: Int, account: Account) {
    viewModelScope.launch {
        val postForm = GetPost(id = id, auth = account.jwt)
        postRes = apiWrapper(API.getInstance().getPost(postForm.serializeToMap()))
        val refreshedPost = postRes
        val existing = postsRes
        when {
            refreshedPost is ApiState.Success && existing is ApiState.Success -> {
                val refreshedPosts =
                    findAndUpdatePost(existing.data.posts, refreshedPost.data.post_view)
                val newPosts = ApiState.Success(existing.data.copy(posts = refreshedPosts))
                postsRes = newPosts
            }

            else -> {}
        }
    }
}
```
This line `refreshedPost is ApiState.Success && existing is ApiState.Success` seems to only be true after the Home screen has already loaded.

MV-GH:
I'll take a look at this once the conflicts are fixed

ZJouba:
I've also noticed that voting on a post doesn't show on the home activity if you navigate back too quickly since the request hasn't been completed by the time the return is consumed 😢

ZJouba:
I don't understand why it doesn't re-compose when the postsRes is updated in `HomeViewModel.kt`?

app/schemas/com.jerboa.db.AppDB/21.json
```diff
@@ -1,218 +0,0 @@
-{
-  "formatVersion": 1,
```
MV-GH, line 2: Why is this file being removed?

app/src/main/java/com/jerboa/api/Http.kt
```diff
@@ -26,7 +26,7 @@ import java.util.concurrent.TimeUnit
 import okhttp3.Response as HttpResponse
 
 const val VERSION = "v3"
-const val DEFAULT_INSTANCE = "lemmy.ml"
+const val DEFAULT_INSTANCE = "voyager.lemmy.ml"
```
MV-GH, line 29: Dont commit these changes

MV-GH (changes requested):
Sorry for the delay, I ll test it later today

app/src/main/java/com/jerboa/ui/components/home/HomeActivity.kt
```diff
@@ -344,6 +353,18 @@ fun MainPostListingsContent(
...
+                account?.also { acct ->
+                    homeViewModel.markPostAsRead(
+                        MarkPostAsRead(
+                            post_id = postView.post.id,
+                            read = true,
+                            auth = acct.jwt,
+                        ),
+                    )
+                }
+            },
```
MV-GH, line 367: All these need to check first if the post is already read ifso dont do the markAsRead, this saves some api calls

app/src/main/java/com/jerboa/ui/components/post/PostActivity.kt
```diff
@@ -167,6 +172,20 @@ fun PostActivity(
     InitializeRoute(postViewModel) {
         postViewModel.initialize(id = id)
         postViewModel.getData(account)
+        account?.also {
+            id.fold({
+                postViewModel.markPostAsRead(
```
MV-GH, line 177: Use `if (id.isLeft())` much clearer imo

MV-GH:
Alright I got this working.

https://github.com/dessalines/jerboa/assets/67873169/873da2a0-373b-4969-b977-19b9f1110c76


Regarding your refreshPost approach,  I don't think we should do this we can perfectly fine achieve this functionality without doing another unnecessary API call (see vid).
 
So what I had to was a few things, first of all: 
I used a ConsumeReturn in HomeActivity for a postview
```kt
    appState.ConsumeReturn<PostView>(PostViewReturn.POST_VIEW) { pv ->
        if (homeViewModel.initialized) homeViewModel.updatePost(pv)
    }
```
In PostActivity I override Back click to add first the AddReturn if the postRes is succesfull
```kt
 onClick = {
                                when (val res = postViewModel.postRes) {
                                    is ApiState.Holder -> appState.addReturn(
                                        PostViewReturn.POST_VIEW,
                                        res.data.post_view
                                    )

                                    else -> {}
                                }

                                appState.navigateUp()
                            },
```
To get this to work for the navigation back btn, I did not implement this yet but for that you will have to use a `onBackPressedDispatcher` to intercept the back call and do the above and call navigateUp, you will have to add this in a `DisposableEffect` so that you can remove the callback in `onDispose`

This makes it already functional but it is not instantly .e.g we have to wait for the result of markRead 
(And currently it marks already read posts a read)

I transformed markPostAsRead into the beast below
Which has a few changes, it checks if the current Post is actually loaded, so that we don't mark posts as read that failed to load.
and it checks if the post is already read. We also instantly mark the post as read (even if the network call fails, if it fails we just correct it)), we do this already with the votes. 

Now we have one problem left is that markRead Api call will be cancelled when the viewmodel is cleared (when we leave the PostScreen) and thus not complete the markread API call. So instead of using the viewmodelscope use the appState scope only for the api call keep the rest of that logic using viewmodelscope, as it does not need to happen if the viewmodel is cleared.

For now i called this from getData. But tbh this should be done from PostActivity in the when(postRes) section on sucessfull to there you can directly pass the data to markRead so that additional when is not needed that checks for the ApiState. Also will be able to pass the scope from appState.
```kt
  fun markPostAsRead(
        form: MarkPostAsRead,
    ) {

        when(val res = postRes) {
            is ApiState.Holder -> {

                if (res.data.post_view.read) return
                postRes = ApiState.Success(res.data.copy(post_view = res.data.post_view.copy(read = true)))
                viewModelScope.launch {

                    markPostRes = ApiState.Loading
                    markPostRes = apiWrapper(API.getInstance().markAsRead(form))

                    when (val markRes = markPostRes) {
                        is ApiState.Success -> {
                            updatePost(markRes.data.post_view)
                        }

                        else -> {}
                    }
                }
            }

            else -> {}
        }


    }
  ```



*(You can see all my changes on the branch issue/403_modified)*

File: app/schemas/com.jerboa.db.AppDB/21.json (hunk 4 above)
Comment (MV-GH) [reply app/schemas/com.jerboa.db.AppDB/21.json:2]: Also add your 22 schema

Comment (ZJouba) [reply app/schemas/com.jerboa.db.AppDB/21.json:2]: I have gone through so many schema migrations with other PRs being merged I'm not sure which should be included or not...

File: app/src/main/java/com/jerboa/api/Http.kt (hunk 5 above)
Comment (ZJouba) [reply app/src/main/java/com/jerboa/api/Http.kt:29]: Oops! Sorry

**Comment by ZJouba** [comment]
> Alright I got this working.
> 
>  studio64_GOShuwRTj3.mp4 
> Regarding your refreshPost approach, I don't think we should do this we can perfectly fine achieve this functionality without doing another unnecessary API call (see vid).
> 
> So what I had to was a few things, first of all: I used a ConsumeReturn in HomeActivity for a postview
> 
> ```kotlin
>     appState.ConsumeReturn<PostView>(PostViewReturn.POST_VIEW) { pv ->
>         if (homeViewModel.initialized) homeViewModel.updatePost(pv)
>     }
> ```
> 
> In PostActivity I override Back click to add first the AddReturn if the postRes is succesfull
> 
> ```kotlin
>  onClick = {
>                                 when (val res = postViewModel.postRes) {
>                                     is ApiState.Holder -> appState.addReturn(
>                                         PostViewReturn.POST_VIEW,
>                                         res.data.post_view
>                                     )
> 
>                                     else -> {}
>                                 }
> 
>                                 appState.navigateUp()
>                             },
> ```
> 
> To get this to work for the navigation back btn, I did not implement this yet but for that you will have to use a `onBackPressedDispatcher` to intercept the back call and do the above and call navigateUp, you will have to add this in a `DisposableEffect` so that you can remove the callback in `onDispose`
> 
> This makes it already functional but it is not instantly .e.g we have to wait for the result of markRead (And currently it marks already read posts a read)
> 
> I transformed markPostAsRead into the beast below Which has a few changes, it checks if the current Post is actually loaded, so that we don't mark posts as read that failed to load. and it checks if the post is already read. We also instantly mark the post as read (even if the network call fails, if it fails we just correct it)), we do this already with the votes.
> 
> Now we have one problem left is that markRead Api call will be cancelled when the viewmodel is cleared (when we leave the PostScreen) and thus not complete the markread API call. So instead of using the viewmodelscope use the appState scope only for the api call keep the rest of that logic using viewmodelscope, as it does not need to happen if the viewmodel is cleared.
> 
> For now i called this from getData. But tbh this should be done from PostActivity in the when(postRes) section on sucessfull to there you can directly pass the data to markRead so that additional when is not needed that checks for the ApiState. Also will be able to pass the scope from appState.
> 
> ```kotlin
>   fun markPostAsRead(
>         form: MarkPostAsRead,
>     ) {
> 
>         when(val res = postRes) {
>             is ApiState.Holder -> {
> 
>                 if (res.data.post_view.read) return
>                 postRes = ApiState.Success(res.data.copy(post_view = res.data.post_view.copy(read = true)))
>                 viewModelScope.launch {
> 
>                     markPostRes = ApiState.Loading
>                     markPostRes = apiWrapper(API.getInstance().markAsRead(form))
> 
>                     when (val markRes = markPostRes) {
>                         is ApiState.Success -> {
>                             updatePost(markRes.data.post_view)
>                         }
> 
>                         else -> {}
>                     }
>                 }
>             }
> 
>             else -> {}
>         }
> 
> 
>     }
> ```
> 
> _(You can see all my changes on the branch issue/403_modified)_

This is what I had initially so that's fine then

File: app/src/main/java/com/jerboa/ui/components/post/PostActivity.kt [hunk 8]
```diff
@@ -324,8 +343,6 @@ fun PostActivity(
                                                 ),
                                             )
                                         }
-                                        // TODO will need to pass in postlistingsviewmodel
-                                        // for the Home page to also be updated
                                     },
                                     onDownvoteClick = { pv ->
                                         account?.also { acct ->
```
ZJouba, line 348: > we do this already with the votes.
I'm not sure

ZJouba:
Okay, I've added the `isRead` checks and also put the api call on the appState coroutines @MV-GH

app/src/main/java/com/jerboa/model/HomeViewModel.kt
```diff
@@ -172,7 +187,29 @@ class HomeViewModel : ViewModel(), Initializable {
...
         updateSortType(SortType.values().getOrElse(account.defaultSortType) { sortType })
-        updateListingType(ListingType.values().getOrElse(account.defaultListingType) { listingType })
+        updateListingType(
+            ListingType.values().getOrElse(account.defaultListingType) { listingType },
+        )
+    }
+
+    fun refreshSinglePost(id: Int, account: Account) {
+        viewModelScope.launch {
+            val postForm = GetPost(id = id, auth = account.jwt)
```
MV-GH, line 197: If this is not used anymore it should be removed

app/src/main/java/com/jerboa/ui/components/post/PostActivity.kt
MV-GH: Yeah i meant that it instantly updates the votes visually not check in the postview itself, we should also make it do the votes using the appState but that's for another PR

app/src/main/java/com/jerboa/model/PostViewModel.kt
```diff
@@ -311,4 +315,22 @@ class PostViewModel : ViewModel(), Initializable {
...
+
+            when (val markRes = markPostRes) {
+                is ApiState.Success -> {
+                    updatePost(markRes.data.post_view)
+                }
+
+                else -> {}
+            }
+        }
+    }
```
MV-GH, line 335: The only concern if I have here is, if the if viewmodel gets cleared it wont cancel this coroutine and note sure what this updatePost() will do then, It's why i rather had that `API.getInstance().markAsRead(form)` only this was done in that appState scope. If it is possible can you change that or, will the code below not do anything in that regard then?

app/src/main/java/com/jerboa/ui/components/post/PostActivity.kt
```diff
@@ -300,6 +305,19 @@ fun PostActivity(
                     is ApiState.Failure -> ApiErrorText(postRes.msg)
                     is ApiState.Success -> {
                         val postView = postRes.data.post_view
+                        account?.also { _ ->
+                            if (!postView.read) {
+                                appState.addReturn(PostViewReturn.POST_VIEW, postView.copy(read = true))
```
MV-GH, line 310: This works for this ig, but I will probably change this later to behaviour i described before so that it will also update the saved , likes, etc

MV-GH (commented):
I will test the functionality of this later, but this PR seems ready

twizmwazin (approved):
LGTM but in the future please avoid reformatting code unrelated to your change.

app/src/main/java/com/jerboa/db/AppDBMigrations.kt
```diff
@@ -287,7 +287,7 @@ val MIGRATION_22_21 = object : Migration(22, 21) {
     }
 }
 
-val MIGRATION_21_22 = object : Migration(21, 22) {
+val MIGRATION_22_23 = object : Migration(21, 22) {
```
MV-GH, line 290: You forgot the numbers in the migration

ZJouba: 🤦‍♂️Sorry

app/src/main/java/com/jerboa/ui/components/post/PostActivity.kt
```diff
@@ -305,6 +310,19 @@ fun PostActivity(
                     is ApiState.Failure -> ApiErrorText(postRes.msg)
                     is ApiState.Success -> {
                         val postView = postRes.data.post_view
+                        account?.also { _ ->
```
MV-GH, line 313: This now needs to be If(!account.isAnon())

app/src/main/java/com/jerboa/ui/components/person/PersonProfileActivity.kt
```diff
@@ -617,6 +626,21 @@ fun UserTabs(
                                     openImageViewer = appState::toView,
                                     openLink = appState::openLink,
                                     showPostLinkPreviews = showPostLinkPreviews,
+                                    markAsReadOnScroll = markAsReadOnScroll,
+                                    onMarkAsRead = {
+                                        if (!it.read) {
+                                            account?.also { acct ->
```
MV-GH, line 632: same here

app/src/main/java/com/jerboa/ui/components/home/HomeActivity.kt
```diff
@@ -397,6 +406,21 @@ fun MainPostListingsContent(
             showPostLinkPreviews = showPostLinkPreviews,
             openImageViewer = appState::toView,
             openLink = appState::openLink,
+            markAsReadOnScroll = markAsReadOnScroll,
+            onMarkAsRead = { postView ->
+                if (!postView.read) {
+                    account?.also { acct ->
```
MV-GH, line 412: also here

app/src/main/java/com/jerboa/ui/components/community/CommunityActivity.kt
```diff
@@ -456,6 +463,21 @@ fun CommunityActivity(
                             showPostLinkPreviews = showPostLinkPreviews,
                             openImageViewer = appState::toView,
                             openLink = appState::openLink,
+                            markAsReadOnScroll = markAsReadOnScroll,
+                            onMarkAsRead = { postView ->
+                                if (!postView.read) {
+                                    account?.also { acct ->
```
MV-GH, line 469: this too

MV-GH (changes requested):
This changes are since that I have made account not nullable anymore, so now instead you have to check if it is Anon account or not

app/src/main/java/com/jerboa/ui/components/post/PostActivity.kt
```diff
@@ -310,7 +311,7 @@ fun PostActivity(
                     is ApiState.Failure -> ApiErrorText(postRes.msg)
                     is ApiState.Success -> {
                         val postView = postRes.data.post_view
-                        account?.also { _ ->
+                        if (!account.isAnon()) {
                             if (!postView.read) {
```
MV-GH: Merge the two ifs

app/src/main/java/com/jerboa/ui/components/person/PersonProfileActivity.kt
```diff
@@ -629,12 +629,12 @@ fun UserTabs(
                                     markAsReadOnScroll = markAsReadOnScroll,
                                     onMarkAsRead = {
                                         if (!it.read) {
-                                            account?.also { acct ->
+                                            if (!account.isAnon()) {
```
MV-GH, line 632: same here

app/src/main/java/com/jerboa/ui/components/home/HomeActivity.kt
```diff
@@ -409,12 +409,12 @@ fun MainPostListingsContent(
             markAsReadOnScroll = markAsReadOnScroll,
             onMarkAsRead = { postView ->
                 if (!postView.read) {
-                    account?.also { acct ->
+                    if (!account.isAnon()) {
```
MV-GH: same here

app/src/main/java/com/jerboa/ui/components/community/CommunityActivity.kt
```diff
@@ -466,12 +466,12 @@ fun CommunityActivity(
                             markAsReadOnScroll = markAsReadOnScroll,
                             onMarkAsRead = { postView ->
                                 if (!postView.read) {
-                                    account?.also { acct ->
+                                    if (!account.isAnon()) {
```
MV-GH: also here

MV-GH (changes requested):

MV-GH:
I did some testing, and if you open a post and close it before it even gets to load and thus fire our markRead, and then open it again let it properly load and then go back you will see that the post in the feed does not turn read. That is because when you open it again and thus request the post again. The read is already true and it wont fire markRead and thus not the `(read= true)`

So what I think is happening is that as soon as you even request to load the post. It will automatically be set to read, by lemmy making our markRead changes unnecessary but I have not confirmed this yet.

MV-GH:
Alright just confirmed this with postman. If you open a unread post, you will get a response that with read=false, if you request that post again it will be read=true.

I'm thinking as soon as you click on a post it should mark it instantly as read in the feed (the feed itself updates the read). The problem is, what if you click on a post and it fails loading it.

app/src/main/java/com/jerboa/ui/components/post/PostActivity.kt
```diff
@@ -305,6 +311,17 @@ fun PostActivity(
...
+                        if (!account.isAnon() && !postView.read) {
+                            appState.addReturn(PostViewReturn.POST_VIEW, postView.copy(read = true))
+                            postViewModel.markPostAsRead(
+                                MarkPostAsRead(
+                                    post_id = postView.post.id,
+                                    read = true,
+                                    auth = account.jwt,
+                                ),
+                                appState,
+                            )
```
MV-GH, line 323: Well this isn't needed anymore and you can remove that function too as I don't think we will ever need it now.

MV-GH (changes requested):
I think the approach we should take is, just keep the addReturn were it is now but it should always do it if you are logged in. Even when the post is` read=true`. The feed will only show it as read if the user sees the post. This will also prevent that previous issue where the post is not turning read in the feed if you open and close it quickly.

app/src/main/java/com/jerboa/ui/components/post/PostActivity.kt
ZJouba: Okay, I've removed it and it will only add the consumable return when the post has successfully loaded.

app/src/main/java/com/jerboa/ui/components/post/PostActivity.kt
```diff
@@ -305,7 +309,7 @@ fun PostActivity(
                     is ApiState.Failure -> ApiErrorText(postRes.msg)
                     is ApiState.Success -> {
                         val postView = postRes.data.post_view
-
+                        appState.addReturn(PostViewReturn.POST_VIEW, postView.copy(read = true))
```
MV-GH, line 312: The anon check needs to be still there. Read feature is only available to real accounts

app/src/main/java/com/jerboa/ui/components/post/PostActivity.kt
```diff
@@ -309,7 +310,7 @@ fun PostActivity(
                     is ApiState.Failure -> ApiErrorText(postRes.msg)
                     is ApiState.Success -> {
                         val postView = postRes.data.post_view
-                        appState.addReturn(PostViewReturn.POST_VIEW, postView.copy(read = true))
+                        if (!account.isAnon()) appState.addReturn(PostViewReturn.POST_VIEW, postView.copy(read = true))
```
MV-GH, line 313: I think you need the curly braces or klint will balk at it

ZJouba: I run ktlint as a precommit and it passed?

MV-GH (approved):

MV-GH:
Ah pipeline died, you can try `git commit --allow-empty -m "Trigger woodpecker" && git push `

ZJouba:
> Ah pipeline died, you can try `git commit --allow-empty -m "Trigger woodpecker" && git push `

Oh I wish. This is a windows laptop 😁 we use `;` here not `&&`

MV-GH:
It does actually work in CMD and since PS 7

PR #1026, opened 2023-07-12
Add general small fixes

MV-GH:
These were mostly included in #1015


app/src/main/java/com/jerboa/ui/components/common/AccountHelpers.kt
```diff
@@ -26,5 +26,5 @@ fun getCurrentAccount(accountViewModel: AccountViewModel): Account? {
 }
 
 fun getPostViewMode(appSettingsViewModel: AppSettingsViewModel): PostViewMode {
-    return PostViewMode.values()[(appSettingsViewModel.appSettings.value ?: APP_SETTINGS_DEFAULT).postViewMode]
+    return getEnumFromIntSetting<PostViewMode>(appSettingsViewModel.appSettings) { it.postViewMode }
```
dessalines, line 29: Doesn't seem any less verbose, but it works.

dessalines (approved):

PR #1027, opened 2023-07-12
Change default of show_parent_comment_navigation_buttons

MV-GH:
Doing it this way, wont affect existing users, just new install and new users migrating from a older install

### DIFF
--- a/app/schemas/com.jerboa.db.AppDB/16.json
+++ b/app/schemas/com.jerboa.db.AppDB/16.json
@@ -64,7 +64,7 @@
       },
       {
         "tableName": "AppSettings",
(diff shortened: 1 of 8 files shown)


MV-GH:
Nvm room doesnt play nice

```
  Caused by: java.lang.IllegalStateException: Room cannot verify the data integrity. Looks like you've changed schema but forgot to update the version number. You can simply fix this by increasing the version number. Expected identity hash: ad45348d127850603284a08efa23652a, found: 48a7d36d42e1a9901e39fcb739f396f7
```

Sadly sqllite doesnt have a each way to change the default of a column, only way is to drop whole table and recreate. This is of course not wanted at all.

Any ideas?

twizmwazin:
Look at this migration: https://github.com/dessalines/jerboa/blob/main/app/src/main/java/com/jerboa/db/AppDBMigrations.kt#L96

twizmwazin (approved):

PR #1029, opened 2023-07-13
Revert #767 should fix post opens the wrong one sometimes

### LINKED ISSUE #1021: Tapping on a post opens the wrong one
**Jerboa Version**
0.0.39

**Android version + phone**
Pixel 4, Android 13

**Describe the bug**
When going through my feed tapping on a post sometimes opens a different one from higher up in the list. It also seems to prefer posts I opened and closed a few seconds before, but I'm not 100% sure on that.

Closing the wrong post and tapping on the desired one again always opens the correct one.

**To Reproduce**

1. Scroll through feed
2. Tap on a post
3. A post from higher up in the list is opened instead
4. Close wrong post
5. Tap on desired post again
6. The correct post is opened

MV-GH:
I tested this build for a half hour and I could not reproduce the issue any more. So believe this fixes it. 
Fixes #1021

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/common/MarkdownHelper.kt
+++ b/app/src/main/java/com/jerboa/ui/components/common/MarkdownHelper.kt
@@ -170,7 +170,6 @@ object MarkdownHelper {
                     }
                     markwon!!.setParsedMarkdown(textView, md)
                 },
-                onReset = {},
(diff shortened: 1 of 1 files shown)


dessalines (approved):
Thanks, I noticed this one occasionally too.

PR #1031, opened 2023-07-13
Change all issue templates to github issue forms

bappitybup:
- Removed old .md issue templates (except questions / discussion)
- Added new issue form github templates ([syntax examples here](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms))
### NEW FEATURE REQUEST ISSUE FORM:
image
### NEW BUG REPORT ISSUE FORM:
image
### in the examples above:
- the summary markdown at the bottom can be viewed and will not be submitted with the issue form
- the issue templates submit the headings and the content entered by users below them (and also the checkboxes)
- some input/textarea/checkbox fields have been made "required" so the submit button can only be pressed when they have content / have been updated
- I have changed the question on the last checkbox to "This is not a question or discussion" rather than "Is this a question or a discussion" so that it can be checked instead of left alone

### Some notes:
- individual checkboxes can be made required or not required so I could change the last checkbox back to how it was and just make it not required.
- I can change the Jerboa Version input field to a dropdown list of versions but then I'd have to update the dropdown list every time a new version is released which seems like a hassle


bappitybup:
https://github.com/bappitybup/testrepo/issues/new/choose <-- examples with the issue forms that you can test

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,95 @@
...
+      label: In the case of a crash
+      description: Please include the logs.
+    validations:
+      required: false
+  - type: markdown
+    attributes:
+      value: |
+        <details><summary>Click here if you don't know how to get the logs</summary>
+
+
```
dessalines, line 53: Instead of this, there's now a crash logs helper. Tell them to go to `settings -> about -> Crash Logs`

dessalines (commented):
Thanks a ton!

> individual checkboxes can be made required or not required so I could change the last checkbox back to how it was and just make it not required.

Yeah make them required if you would.

A dropdown for versions isn't necessary.

.github/ISSUE_TEMPLATE/bug_report.yaml
MV-GH: Keep that one, cuz the crash log helper wont help if the crash is on startup. And we have had a handfull of those.

bappitybup:
> Thanks a ton!
> 
> > individual checkboxes can be made required or not required so I could change the last checkbox back to how it was and just make it not required.
> 
> Yeah make them required if you would.
> 
> A dropdown for versions isn't necessary.

no prob 👌 all of them are already required

.github/ISSUE_TEMPLATE/bug_report.yaml
bappitybup: should I just add some text in this summary dropdown for the crash log helper? or replace it altogether?

MV-GH: Tell them to use the previous method if the crash is on startup else use the crash logs

bappitybup:
I have added two sections for startup crashes and general crashes, updated the example as well
--> https://github.com/bappitybup/testrepo/issues/new?assignees=&labels=bug%2Ctriage&projects=&template=bug_report.yaml

.github/ISSUE_TEMPLATE/bug_report.yaml
bappitybup: see latest comment in PR, added two sections
image
**FOR CRASHES ON STARTUP** has the old content
**FOR GENERAL CRASHES** has this:
image

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,108 @@
...
+        ### How Do I Get Crash Logs?
+        **<details><summary>FOR CRASHES ON STARTUP (Click to expand)</summary>**
+        > 
+        > Follow these steps (skip the ones you already have done):
+        > - Enable developer options and USB DEBUGGING
+        > - Install ADB on your PC
+        > - Connect your phone to your PC
+        > - Execute this command on PC
+        > 
+        > ### To enable developer options and USD DEBUGGING
```
MV-GH, line 62: Fix my typo there, and ill approve it

MV-GH (approved):

PR #1033, opened 2023-07-14
Add option to show/hide post previews

twizmwazin:
In an issue somewhere someone suggested that titles and bodies were being duplicated when it was actually just that the user copied the title and article description into their post. I thought maybe adding an option to disable the preview would be a good idea.

When enabled/current behavior:
image

After with disabled:
image

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
+++ b/app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
@@ -409,6 +409,7 @@ fun PostBody(
     useCustomTabs: Boolean,
     usePrivateTabs: Boolean,
     blurNSFW: Boolean,
+    showPostPreview: Boolean,
     openImageViewer: (url: String) -> Unit,
     navController: NavController,
     clickBody: () -> Unit = {},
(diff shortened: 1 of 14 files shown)


bappitybup:
can't view these images, I get this
# #
Screenshot_20230714-100654

MV-GH:
I think this is better named post link preview. Not to be confused with post markdown preview.

MV-GH:
Another option we can look into is, automatically hide the meta description if its the same as the meta title, and the meta title if its the same as the post title.

MV-GH (approved):

dessalines:
@twizmwazin I'm getting the same github image error, could you re-upload the before / after pics?

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
```diff
@@ -429,7 +430,7 @@ fun PostBody(
         )
 
         // The metadata card
-        if (fullBody && post.embed_title !== null) {
+        if (fullBody && showPostLinkPreview && post.embed_title !== null) {
```
dessalines, line 433: Ohh okay this is about the metadata cards.

dessalines (approved):
I don't need to see the before and after, its just about the metadata cards.

LGTM

PR #1044, opened 2023-07-15
Make datatypes Parcelable and Stable

### LINKED ISSUE #810: Make autogen types extend Parcelable or Serializable
Like the title suggests, make the autogenerated types extend Parcelable or Serializable to be usable with SavedStateHandle and Bundles.

MV-GH:
I used v0.18.0 to generate the types it seems to added a few properties. Not sure if I did hot_rank correctly.

This pr makes all the datatypes Parceable and makes the one that weren't stable, stable due to them using List.
This fixes 33 datatypes that were previously unstable.

Example

```
unstable class MyUserInfo {
  stable val local_user_view: LocalUserView
  unstable val follows: List<CommunityFollowerView>
  unstable val moderates: List<CommunityModeratorView>
  unstable val community_blocks: List<CommunityBlockView>
  unstable val person_blocks: List<PersonBlockView>
  unstable val discussion_languages: List<Int{ com.jerboa.datatypes.types.LanguageId }>
  <runtime stability> = Unstable
}

unstable class PictrsImages {
  stable val msg: String
  unstable val files: List<PictrsImage>?
  <runtime stability> = Unstable
}
```

after

```
stable class MyUserInfo {
  stable val local_user_view: LocalUserView
  unstable val follows: List<CommunityFollowerView>
  unstable val moderates: List<CommunityModeratorView>
  unstable val community_blocks: List<CommunityBlockView>
  unstable val person_blocks: List<PersonBlockView>
  unstable val discussion_languages: List<Int{ com.jerboa.datatypes.types.LanguageId }>
}

stable class PictrsImages {
  stable val msg: String
  unstable val files: List<PictrsImage>?
}

```

Fixes #810 

@nahwneeth not sure for what you needed them Parcelable though


nahwneeth:
https://github.com/dessalines/jerboa/blob/main/app/src/main/java/com/jerboa/ui/components/common/Navigation.kt#L54

If the object is parcelable, then here rememberSaveable can be used directly (no need to manually convert to json).

MV-GH:
@nahwneeth 

I took a stab at replacing it but it does not work with nullable types.
How would I solve this

```kt
inline fun <reified D : Parcelable?> NavController.takeDepsFromRoot(): State<D> {
    val deps = rootChannel<D>().take()

    // This will survive process death
    val depsSaved = rememberSaveable { deps!! }

    // After process death, deps will be null
    return remember(depsSaved) {
        derivedStateOf {
            deps ?: depsSaved
        }
    }
}
```

copy_generated_types_from_lemmy_js_client.sh
```diff
@@ -45,15 +54,44 @@ for filename in *.ts;
...
+  # Add @Parcelize annotation
+  sed -i 's/data class/import android.os.Parcelable\nimport kotlinx.parcelize.Parcelize\n\n@Parcelize\ndata class/g' "$kt_file"
+
+  # Add Parcelable inheritance
+  if grep -q "data class" "$kt_file"; then
+      echo " : Parcelable" >> "$kt_file"
+  fi
+
+  # Adds @Immutable annotation to classes containing lists, to mark them stable
+  if grep -q "List<" "$kt_file"; then
```
dessalines, line 77: Since you're regenerating all of these anyway, now would probabably be a good time to convert all the `List` to `ImmutableList`, `L41` above.

MV-GH: I tried that approach but refrofit/whatever makes the api would not work with it. Was easier to mark it stable.

nahwneeth:
I think a null check will solve this. If the deps is null, then it'll be null anyway after process death also.

copy_generated_types_from_lemmy_js_client.sh
dessalines: Gotcha, nbd.

MV-GH:
Yes but for a non nullable on process death, `rootChannel<D>().take()` will also return null in which case it has to return the savedDeps. How do I differentiate the two.

nahwneeth:
oh okay, I see. Can we make D non-nullable and return State<D?>

MV-GH:
How would that work for nullable types? Currently we have one, createPost has `Community?` which returns the community if created from a CommunityActivity otherwise null.

nahwneeth:
I didn't expect nullability to be a problem tbh. Ig we need to create a wrapper for that which contains a Community? within it. Removing one hack is requiring another. smh.

MV-GH:
Alright, I ll make a separate issue for this and keep `takeDepsFromRoot` as is.

twizmwazin (approved):

PR #1048, opened 2023-07-16
Fix links looking like lemmy links being used as one

### LINKED ISSUE #1020: App crashes when opening specific link
**Jerboa Version**
Tried on 0.0.38 and 0.0.39

**Android version + phone**
Pixel 4, Android 13

**Describe the bug**
The following post and others with the same link crash the app every time when opening the link: https://lemmy.ml/post/1956845

Other links seem to work just fine.

**To Reproduce**

1. Just tap on the link in the post

**In the case of a crash**
I'm on my phone right now, will add the logs later.

MV-GH:
- Updates instance list
- Use a set instead of array for instance list (for O(1) search)
- Fixes lemmy thinking /c/ or /u/ as lemmy instances. Now it does a check if the host is not in LEMMY_INSTANCES, which then gets cached, for links formatted as lemmy instances it directly assumes they are lemmy instances
- Fixes instance/c/community@otherinstance not being displayed correctly in community (basically just errors)

studio64_w1TfQLJKuE

basically what would happen is: 
at openLink receives
-> https://lemmy.one/c/technology@beehaw.org
Transformed and navigated to
-> lemmy.one/c/technology@beehaw.org
-> instance/c/community@otherinstance
 by logic in navhost turned into
-> community@otherinstance@instance
due to {instance}/c/{name} -> {name}@{instance}
studio64_w1TfQLJKuE

*Shows the fix, this would previously show only errors*

https://github.com/dessalines/jerboa/assets/67873169/f93009d8-453c-40a1-80ec-bc224ee699fd

Fully Fixes #1020
This link was previously thought as a community by Jerboa, (and crashed before my other fix)

https://github.com/dessalines/jerboa/assets/67873169/96d3cc29-d3a0-4974-b512-ce9e6366b0cc


app/src/main/java/com/jerboa/api/Http.kt
```diff
@@ -24,6 +24,8 @@ const val DEFAULT_INSTANCE = "lemmy.ml"
 const val MINIMUM_API_VERSION: String = "0.18"
 val REDACTED_QUERY_PARAMS = setOf("auth")
 val REDACTED_BODY_FIELDS = setOf("jwt", "password")
+val TEMP_RECOGNISED_AS_LEMMY_INSTANCES = mutableSetOf<String>()
+val TEMP_NOT_RECOGNISED_AS_LEMMY_INSTANCES = mutableSetOf<String>()
```
dessalines, line 28: Where are these getting filled? Should you just use the hardcoded `DEFAULT_INSTANCES` ?

Globals should probably go inside the singleton if you really need them, but I don't see why you do, because `DEFAULT_INSTANCES` already exists.

app/src/main/java/com/jerboa/api/Http.kt
```diff
@@ -251,18 +253,22 @@ interface API {
...
         fun getInstance(): API {
             if (api == null) {
-                api = buildApi()
+                api = buildApi(buildUrl())
             }
             return api!!
         }
 
-        private fun buildApi(): API {
+        fun createTempInstance(host: String): API {
```
dessalines, line 267: What is this about?

app/src/main/java/com/jerboa/Utils.kt
```diff
@@ -311,33 +326,33 @@ fun LazyListState.isScrolledToEnd(): Boolean {
  * - !community@instance -> https://instance/c/community
  * - @user@instance -> https://instance/u/user
  */
-fun parseUrl(url: String): String? {
+fun parseUrl(url: String): Pair<Boolean, String>? {
```
dessalines, line 329: What's the Boolean for? Should probably add a comment.

app/src/main/java/com/jerboa/api/Http.kt
MV-GH: I needed to create a temporary  API Instance using the new host to test if its a Lemmy instance. Currently that was not possible without removing the existing API.

app/src/main/java/com/jerboa/Utils.kt
MV-GH: Added a comment

app/src/main/java/com/jerboa/api/Http.kt
MV-GH: I moved it and checkIfLemmyInstance to the singleton

PR #1062, opened 2023-07-19
Update Brazilian Portuguese translation

Gustavo-Martins:
- Fixes a typo in pt-BR dir name;
- Updates Brazilian Portuguese translation with new strings.


MV-GH (changes requested):
Put the old dir name back, it is not a typo. It is required to put `r` for region infront of regions.

Gustavo-Martins:
> Put the old dir name back, it is not a typo. It is required to put `r` for region infront of regions.

Will do, thanks for the correction!

Gustavo-Martins:
I dropped the commit with my incorrect renaming, thanks again @MV-GH for the correction.

PR #1063, opened 2023-07-19
Fix dropdown popup issue

MV-GH:
Currently the popup for login disappears when typing. This actually works fine in the emulator but not on actual device.

Still not fan of this solution since you can see the popup disappear and appear when typing but it is improvement until a better fix is found.

Originally i had no `onDismissRequest = {}` but that also means that you have to select an option from the list or will not disappear. But when you add it, it works on emulator but not on actual device. Which makes it disappear when you type. 

I tried a bunch of other solutions like disabling the dissmiss when typing but that had race condition issues.

before/current

https://github.com/dessalines/jerboa/assets/67873169/56fb4b2d-ba94-4c84-bc53-9b406600db26

after

https://github.com/dessalines/jerboa/assets/67873169/6f897ba2-2752-4dd5-a464-135053e52c2c


dessalines (approved):
Tested this and it works great, thx.

PR #1064, opened 2023-07-19
Add divider to section for report actions

MV-GH:
Someone kept accidantally reporting posts instead of sharing it, So I moved it to the bottom and added a small section for it.

Tbh all those dialogs need to be redone at some point to add subnavigation menus because it is getting to big and that padding at the bottom is annoying. Maybe we should use https://m3.material.io/components/menus/guidelines#691d98ba-0375-4681-8c47-a5c3e7b58798

How it looks now
N22YDUP2Gy

I had recently added the missing old sort type but I had used the wrong icon. Now it has the same icon as the old sort used already somewhere.


twizmwazin (changes requested):
I think there needs to be more padding around the separator. I also think report might want to go before block community/user, though that's more subjective.

MV-GH:
image
Updated the padding

twizmwazin (approved):

PR #1068, opened 2023-07-20
Move unExpandedComments and commentsWithToggledActionBar from PostActivity to PostViewModel

### LINKED ISSUE #643: Comments Uncollapse Upon Reply
** Jerboa Version **
Jerboa 0.0.34

**Describe the bug**
When I reply to a comment in a given comment section, all of the collapsed comments expand and I'm reset to the top of the comment section.

**To Reproduce**

1. Go to any comment section
2. Collapse some of the comments
3. Reply to a comment
4. Bug triggers

twizmwazin:
Right now, if you've collapsed some comments or toggled their action bars and then navigate to a new page, when you return back to that post the comment state is reset because it is tied to the composable rather than the view model. Moving this state to the view model fixes this issue.

Is there other state where we would want to do this as well?

Fixes #643


MV-GH:
fix the formatting

twizmwazin:
Most likely, yes. I'll test it.

MV-GH (approved):

MV-GH:
Would this also work for saving the current scroll position?

twizmwazin:
Any state that we want to tie to the lifetime of the ViewModel this method should work for. So, if the HomeViewModel isn't destroyed and recreated when navigating on the navbar between tabs, then this method should work. A case where this wouldn't work is when a user backs out of a post and then re-opens the post from the home or community view, because that would cause the view model to be destroyed and then recreated.

dessalines (approved):
I tested a few scenarios:

action | works
--- | ---
Navigating from post -> a user or community page | Yes
Navigating from post to a create / edit comment | Yes
Navigating from post back to home, then same post | No

I'd say even with those two improvements, this is good.

twizmwazin:
If we want to fix "Navigating from post back to home, then same post", then we would probably need to move the postviewmodel reference to the home/community activity instead of the post activity, and then when a user opens a post, re-use that viewmodel if it is the same as the previously opened post, otherwise replace it. I can give this a shot later.

PR #1077, opened 2023-07-21
Instantly mark posts as read once opened in PostListings

twizmwazin:
This stores read posts in the relevant view model and renders them as viewed even if the PostView is out of date. Similar to the instant votes system already in use.

### DIFF
--- a/app/src/main/java/com/jerboa/model/CommunityViewModel.kt
+++ b/app/src/main/java/com/jerboa/model/CommunityViewModel.kt
@@ -59,6 +59,8 @@ class CommunityViewModel : ViewModel(), Initializable {
     var page by mutableIntStateOf(1)
         private set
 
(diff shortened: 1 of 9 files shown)


MV-GH:
Not sure if should post this here or the other PR or just as a separate issue but posts are marked as read in saved an profile which shouldnt imo.

image

image

dessalines (commented):
This isn't the right way to handle this. There's a LOT of other info that needs to get synced from the postviewmodel to other viewmodels and vice versa, things like votes, saved, read, newcomments, etc. Read is just one of many.

twizmwazin:
That's a seperate issue but one I can quickly fix later today.

MV-GH:
Could we just do a consumeReturn on the modified postview and use that to update the list?

MV-GH:
Will be obsolete once #1025 is merged

Issue #931, opened 2023-07-01
Unsuccessful login attempt still shows name and avatar in sidebar (as if successful login)


Snapz54:
**Jerboa Version**
37

**Android version + phone**
Android 9, Samsung Galaxy Note 8

**Describe the bug**
Beehaw server isn't updated, so login attempt to Jerboa shows as unsuccessful with appropriate error message, but backing out of login screen, my account name and avatar are now shown in sidebar as if login was successful (though posting, commenting and voting are unsuccessful as app isn't actually logged into account).

**To Reproduce**

1. Go to sidebar
2. Click on login
3. Enter credentials to an instance where server isn't updated (e.g. Beehaw at the time of this bug report)
4. See error message for server needing update
5. Back out of login screen and see account info present in sidebar

Note: ideally server not being upgraded could still allow login, just with notification that update is needed. Not sure implications of allowing this though.

PR #1082, opened 2023-07-21
Prevent duplicate accounts from logging in

### LINKED ISSUE #987: Allows you to login with an already logged in account
**Jerboa Version**
0.38

**Android version + phone**
Type here the android version you are using and which phone

**Describe the bug**
It allows you to login multiple times with the same account that is already logged in.
image
But upon switching it does seem to remove the duplicate account.

EDIT:
I just noticed there is no sign out option either huh

https://github.com/dessalines/jerboa/assets/67873169/fc1244d5-8e03-4c42-82e5-ffa5602eb14d

Edit:
At this point i think its just replacing the sign out button. Each additional attempt to add more duplicate accounts doesnt add another entry.

**To Reproduce**

login with a account, then login again

MV-GH:
Fixes #987 

https://github.com/dessalines/jerboa/assets/67873169/8fbaf24c-07cb-4616-a721-10d9bfa8e4ed

I used name + instance to define duplicate because person.Id could be duplicate for two separate accounts on different instances.

I do feel like this whole login function with its numerous changeLemmyInstances should be refactored to use createTempInstance and only be swapped on login.

### DIFF
--- a/app/src/main/java/com/jerboa/model/LoginViewModel.kt
+++ b/app/src/main/java/com/jerboa/model/LoginViewModel.kt
@@ -22,6 +22,7 @@ import com.jerboa.getHostFromInstanceString
 import com.jerboa.serializeToMap
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.launch
+import java.lang.RuntimeException
 
 class LoginViewModel : ViewModel() {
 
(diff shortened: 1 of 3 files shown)


MV-GH (commented):
How does this section prevent logging in with a non supported account?

```kt
             if (compareVersions(siteVersion, MINIMUM_API_VERSION) < 0) {
                        val message = ctx.resources.getString(
                            R.string.dialogs_server_version_outdated_short,
                            siteVersion,
                        )
                        Toast.makeText(ctx, message, Toast.LENGTH_SHORT).show()
                    }
```

Hmm this related #931 ?

twizmwazin (approved):

PR #1084, opened 2023-07-21
Add switch to anonymous option

MV-GH:
Small QoL, many times I wanna test something in anon but I dont want to remove my currently logged in account. This will solve that.

https://github.com/dessalines/jerboa/assets/67873169/b4858b5d-4eaf-4cbe-b2fe-f54a763017c5

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/drawer/Drawer.kt
+++ b/app/src/main/java/com/jerboa/ui/components/drawer/Drawer.kt
@@ -70,6 +70,7 @@ fun Drawer(
     onAddAccount: () -> Unit,
     onSwitchAccountClick: (account: Account) -> Unit,
     onSignOutClick: () -> Unit,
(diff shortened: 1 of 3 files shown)


twizmwazin (approved):
In the future it would also be nice to support anonymous on a user-entered instance.

MV-GH:
For that it would probably be better  to be able to configure the default API instance

twizmwazin:
Suppose someone prefers not to have an account but wants to read the local feed on a few different instances without logging in. I think it would make more sense to present it as an option in the account switcher than forcing the user to go and update the default instance each time.

MV-GH:
So something like anon@lemmy.ml anon@beehaw.org? And where would we make such account? in create account?

twizmwazin:
Yeah, probably something like "login" or "save as anonymous"

PR #1091, opened 2023-07-23
Go to comment when clicking on reply comment in inbox, fixes #1089

### LINKED ISSUE #1089: Clicking a comment in my inbox doesn't bring up the comment
### Pre-Flight checklist

When I have a reply in my inbox, clicking on the comment doesn't actually bring up the comment, it brings the comment thread the reply is in. To get to that exact reply, I'd have to scroll through the whole thread to find it. If it is a big thread, forget it. 
Clicking the reply should take me straight to it. 
Note: I did search the open issues and couldn't find anything. If this is duplicate, my apologies in advance.

twizmwazin:


app/src/main/java/com/jerboa/ui/components/inbox/InboxActivity.kt
```diff
@@ -407,7 +407,8 @@ fun InboxTabs(
                                                 }
                                             },
                                             onPostClick = { postId ->
-                                                appState.toPost(id = postId)
+                                                goToComment(commentReplyView)
+                                                markAsRead(commentReplyView)
```
MV-GH, line 411: This will cause it to mark already marked as read comments as not read yet.
Do the same as
```kt
                              // Do not mark already read reply as read
                                                if (!crv.comment_reply.read) {
                                                    markAsRead(crv)
                                                }
```

MV-GH (changes requested):

app/src/main/java/com/jerboa/ui/components/inbox/InboxActivity.kt
MV-GH: Also if that `postId` isnt used then refactor that function to not require it

dessalines:
Does this overlap with  #1127?

dessalines (approved):

PR #1102, opened 2023-07-27
Add Norwegian Nynorsk translation

huftis:
Here‘s a complete translation of Jerboa into Norwegian Nynorsk.

I have a question about some of the sorting strings, though. They look like this:

    NewComments
    TopMonth
    TopSixHour

They look like internal, machine-readable strings, which should *not* be translated or translatable, i.e., should *not* be present in `strings.xml`. Are they included by mistake?

I did, however, translate them (as similar machine-friendly strings), as I see they are translated in string files for the other languages.

Note that there are also corresponding ‘human-friendly’ strings included, e.g., ‘New Comments’ and ‘Top Month’. And these are the ones actually used in the UI. One exception is `TopAll`, which doesn’t have a human-friendly version in (‘Top All’) in `strings.xml`. And in the UI in the app, the option is indeed listed as ‘TopAll’, while all the other sorting options are listed using the human-friendly versions.


MV-GH:
They are included and visible to the user, It is just a shorter version of the longer names on places where we have less screen real estate.

app/src/main/res/values-nn/strings.xml
```diff
@@ -0,0 +1,348 @@
+<?xml version='1.0' encoding='utf-8'?>
```
MV-GH, line 1: This is not needed

huftis:
> They are included and visible to the user, It is just a shorter version of the longer names on places where we have less screen real estate.

OK. Though it looks like a ‘Top All’ string is missing at
https://github.com/dessalines/jerboa/blob/main/app/src/main/java/com/jerboa/Utils.kt#L1117

That is, `MAP_SORT_TYPE_LONG_FORM` reuses the `R.string.sorttype_topall` (`TopAll`) string instead of having a long-form `R.string.dialogs_top_all` string.

app/src/main/res/values-nn/strings.xml
huftis: It’s automatically generated by the tooling I use (android2po/`a2po`) to manage the translation. Is it a problem that it’s included? If so, I can of course manually remove it.

MV-GH: Not really, it was just for parity with the other translations

MV-GH:
image
Pipeline having issues,

MV-GH (approved):

PR #1107, opened 2023-07-27
Use marquee for instance in topbar, and add it to communitysidebar

MV-GH:
I prefer this over the ellipsis

https://github.com/dessalines/jerboa/assets/67873169/4f959e28-c8df-419a-b8e9-6a8cd2ef76d2


MV-GH:
It only does the scrolling thing when names are too long 99% it does not show, and any other action needs UX changes so that users will know they can do an action

twizmwazin (approved):
LGTM after rebase

MV-GH:
Tbh I am not a fan of it either but I still prefer it over the ellipsis thing as the point of these changes are so that you can see the instance that community is from. And ellipsis prevents it. Ideally we add it somewhere else on a better place. And then we can revert these topbar changes and have it the name only. But this works for now.

PR #1108, opened 2023-07-27
Add direct message person option and screen

MV-GH:
Also fixed a bug where it shows all the more options in the dialog for non logged in users and all those option need the user to be logged in.

https://github.com/dessalines/jerboa/assets/67873169/b4b76129-9707-46ee-aa7b-a8af5bd6ee4f


app/src/main/res/values/strings.xml
```diff
@@ -347,4 +347,8 @@
     <string name="failed_saving_image">Failed to save image!</string>
     <string name="home_switch_anon">Switch to anonymous</string>
     <string name="login_already_logged_in">You are already logged in with this account</string>
+    <string name="person_profile_dm_person">Message this person</string>
```
twizmwazin, line 350: `Message <username>` is preferable

app/src/main/res/values/strings.xml
```diff
@@ -347,4 +347,8 @@
...
+    <string name="private_message_failed">Sending message failed, retrying...</string>
+    <string name="private_message_placeholder">Type here your message</string>
```
twizmwazin, line 352: `Type your message here`

twizmwazin (commented):
LGTM with those two string changes

app/src/main/res/values/strings.xml
MV-GH: It's because the other two options are X this person

MV-GH: Message <username> can get possibly pretty long depending on the username, what about `Send message` ?

PR #1111, opened 2023-07-29
Enchance imageviewer: Downloadprogressbar, error placeholder + retry, large images subsampling, SVG support

MV-GH:
Before we were recreating a bunch of imageloaders which were creating a bunch of HttpClients. This pr aims to solve this.

- Make imageloader a singleton that gets reused
- Make HTTPclient a singleton that gets resused
- Add SVG support
- Moved application to a seperate file
- Added a loadingbar to the imageviewer that tracks the download progress of the image
- increase the ability to zoom in
- Opened: https://github.com/coil-kt/coil/issues/1833
- Used a ZoomableAsyncImage instead now, which supports subsampling larges images
- Added error placeholder (I drew that myself found no SVG CC error placeholder images)
- Added foundation to track upload/download progress
- Removed useless repository section in build :Jerboa, it gets overwritten from settings.gradle anyway, wasted some time on that

> OkHttp performs best when you create a single OkHttpClient instance and reuse it for all of your HTTP calls. This is because each client holds its own connection pool and thread pools. Reusing connections and threads reduces latency and saves memory. Conversely, creating a client for each request wastes resources on idle pools.

Loading bar example

https://github.com/dessalines/jerboa/assets/67873169/4b8d8683-1c0f-4fea-827e-93ec72921290

error failure

https://github.com/dessalines/jerboa/assets/67873169/93d5c9a0-019f-469f-bca7-d795ffc1de07

Large image subsampling

https://github.com/dessalines/jerboa/assets/67873169/b1038b47-7054-4d19-95ae-14470636e28f

ZoomableAsyncImage does have some slight issues. I can revert back to AsymcImage untill they are fixed.

see https://github.com/saket/telephoto/issues


twizmwazin (approved):

app/src/main/java/com/jerboa/util/downloadprogress/ProgressEvent.kt
```diff
@@ -0,0 +1,19 @@
+package com.jerboa.util.downloadprogress
```
dessalines, line 1: Wish these things were already built in to the http clients. oh well.

dessalines (approved):
Works well, thx!

PR #1120, opened 2023-08-01
Update deps, gradle and target SDK 34

MV-GH:
Deps needed to be updated, compose 1.5 requires that we target SDk 34. So  I did that and test the app. It works fine.

The only issue is that the status bar seems to be wrong.

That's probably an issue with compose though cuz we don't manually set nowhere how high the status bar should be.
I'll make a separate issue to look into it later.

image


twizmwazin (approved):
LGTM after rebase

Issue #443, opened 2023-06-06
Subscriptions don't display until app is closed and reopened


tsgagnon:
When I subscribe to a community within the app, that community won't show in the Subscriptions list until I close and reopen the app.

dessalines:
Thanks for this, ya we do need to update those in the viewmodels after a subscribe.

TrollBlox:
or an un-subscribe ;)

PR #1131, opened 2023-08-03
Update drawer Subscriptions

ZJouba:
Addresses #443
  
Subscriptions - Made with Clipchamp

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/drawer/Drawer.kt
+++ b/app/src/main/java/com/jerboa/ui/components/drawer/Drawer.kt
@@ -44,6 +44,7 @@ import androidx.lifecycle.viewmodel.compose.viewModel
 import com.jerboa.R
 import com.jerboa.datatypes.samplePerson
 import com.jerboa.datatypes.types.Community
+import com.jerboa.datatypes.types.CommunityFollowerView
 import com.jerboa.datatypes.types.ListingType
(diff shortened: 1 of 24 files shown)


app/src/main/java/com/jerboa/ui/components/drawer/Drawer.kt
```diff
@@ -70,6 +71,7 @@ import com.jerboa.ui.theme.muted
 @Composable
 fun Drawer(
     myUserInfo: MyUserInfo?,
+    follows: List<CommunityFollowerView>,
```
MV-GH, line 74: Use immutableList if you pass list through composables, list is deemed unstable by compose see https://issuetracker.google.com/issues/199496149

MV-GH (changes requested):

dessalines (approved):
Works well, thx!

app/src/main/java/com/jerboa/ui/components/drawer/Drawer.kt
```diff
@@ -279,6 +284,7 @@ fun DrawerItemsMain(
 fun DrawerItemsMainPreview() {
     DrawerItemsMain(
         myUserInfo = null,
+        follows = listOf<CommunityFollowerView>() as ImmutableList<CommunityFollowerView>,
```
MV-GH, line 287: use `persistentListOf`

PR #1132, opened 2023-08-03
Adding slow paths check to CI.

dessalines:
This will prevent any non-kotlin changes from running the expensive CI steps. So PRs that only update things like the README won't trigger full builds.


MV-GH:
The benchmark module also needs to be added. Unrelated but is woodpecker already updated to v1.0.0? it updated a week ago

dessalines:
Yep woodpecker is updated as of yesterday.

dessalines:
There's so many exceptions, that I might as well close this.

Issue #1065, opened 2023-07-19
Replace all relevant dialogs with M3 menus (Such as Sort/More options/post more options)


MV-GH:
### Pre-Flight checklist

- [X] Did you check to see if this issue already exists?
- [X] This is a single feature request. (Do not put multiple feature requests in one issue)
- [X] This is not a question or discussion. (Use https://lemmy.ml/c/jerboa for that)

### Describe The Feature Request Below

Replace all relevant dialogs with M3 menus (Such as Sort/More options/post more options)

See https://m3.material.io/components/menus/guidelines#834fd8ef-b704-420d-8bf6-f2d21e19d2cc

Also add sub menus. especially for the more options on a post, should add one for Copy (4 copy things), Block (Block user/community), Report (Report / Write full Report) , Share (Link / image / post)

This will also fix bottom padding and should imo be more more cleaner and elegant design.

@dessalines @twizmwazin Do you agree with this?

twizmwazin:
Sub menus: enthusiastic yes, this is desperately needed

M3 menus: Depends. If the menu originates from a top bar or a FAB, definitely yes. If its menu originates from scrollable content like a comment or post, I'm concerned placement will be wonky, and a dialog would be preferable.

MV-GH:
It doesn't have to appear from where it was originated for moveable content, 

We could also do this 
![image](https://github.com/dessalines/jerboa/assets/67873169/76521d0b-928a-4285-aef6-9bb311a0c399)
But submenus will probably not work with that

PR #1146, opened 2023-08-06
Make app bar 'more' buttons use dropdowns rather than dialogs

camporter:
This does some of the work for #1065 by moving more/three-dot buttons on top app bars over to use dropdowns rather than dialogs.

Whether to move other dialogs to dropdowns has some concerns in that issue (such as scrollable areas) and might be facilitated by other components, so this is limited intentionally.


MV-GH (approved):
Ah, looks so much better now, thanks! I am planning on changing all dialogs to M3 menus. And for the post listing more options I was looking into https://github.com/saket/cascade. But bitesized PRs like this are better.

PR #1150, opened 2023-08-07
Use listing type use M3 dropdown instead of dialog

MV-GH:
part of #1065

https://github.com/dessalines/jerboa/assets/67873169/c3bf7b1b-f300-43d8-8006-2f667b2d9513


MV-GH:
for sortTypes it needs to centrally put them like it does now with dialogs

dessalines (commented):
Same with this one, could you post before / after pics?

MV-GH:
There is a video, and the other looks the same as this but with different menu items

dessalines:
Oops my bad. Looks great.

dessalines (approved):

PR #1152, opened 2023-08-07
Use M3 menu dropdown for inbox unread sort

MV-GH:
Part of  #1065


dessalines (commented):
Could you post some before / after pics?

dessalines (approved):

Issue #639, opened 2023-06-14
Action Bar Buttons get pushed off screen when comment count is too high


kuroi-usagi:
** Jerboa Version **
0.0.34

**Describe the bug**
Action Bar Buttons beneath a post are being pushed off screen when the comment counter gets to high or has too much text  alltogether.

**To Reproduce**
1. Open Jerboa
2. Open one of the top comments with more then 100 comments
3. Check if the more action button is still visible
4. It is not (see screenshot)

![Screenshot_20230614_130829_Jerboa (DEBUG)](https://github.com/dessalines/jerboa/assets/40325654/38eb24b5-46d2-42af-ad15-265166ebae4f)

I have a quick fix for this but its not a pretty one.

![Screenshot_20230614_130853_Jerboa (DEBUG)](https://github.com/dessalines/jerboa/assets/40325654/0f443584-0bc0-48f6-851d-1aecfce56278)

servoparmi:
What version of Android is that screenshot from?

kuroi-usagi:
> What version of Android is that screenshot from?

Android 13

noloman:
This is about the hamburger menu, isn't it? the three vertical dots (...)

kuroi-usagi:
Yes exactly those, but it concerns basically all the buttons in this section. In extreme cases all of them will be pushed out of the screen.

noloman:
Ok, I think I can try to take care of this

ZJouba:
@twizmwazin, @MV-GH, @dessalines how do we want to tackle this? Even setting the Arrangement to SpaceAround doesn't leave enough space. We should also decide on a minimum width that we want to cater for, I think 240px is the norm?
![image](https://github.com/dessalines/jerboa/assets/42601911/4b48d0eb-eef2-409c-8c15-54a298c3723f)
It get's especially messy if all the buttons are available
Should we move ActionButtons to the More menu when there isn't enough space?

MV-GH:
I am open for changing `XXX comments (XXX new)` to `XXX (+XXX)` But only for card layout as list layout does not suffer this problem. I was discussing this in matrix too

dessalines:
I think a flowrow, and maybe smaller margins is probably the best way to handle it. That would safely cover all screen widths by wrapping to new lines.

Freeman999:
[...]

PR #1153, opened 2023-08-07
New PostListing CommentCount appearance

ZJouba:
Addresses #639 

image

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/common/AppBars.kt
+++ b/app/src/main/java/com/jerboa/ui/components/common/AppBars.kt
@@ -44,6 +44,7 @@ import com.jerboa.datatypes.samplePerson
 import com.jerboa.datatypes.samplePost
 import com.jerboa.datatypes.types.Person
 import com.jerboa.db.entity.Account
+import com.jerboa.db.entity.AnonAccount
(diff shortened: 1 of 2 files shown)


MV-GH:
image

Missing some logic, if its equal it should not display, and needs some iteration

ZJouba:
Apologies, didn't realise it wasn't just the count. 
What do you mean by
> and needs some iteration  

?

ZJouba:
Oh just changing it to draft. Coolio!

Hopefully less merge commits this time 😁

ZJouba:
@MV-GH is there a commit incoming?

Also, what do you use for dev, emulator or device?

MV-GH:
It is going to be on a different branch, I use a real device but with live edit on, faster than preview rendering, more accurate and interactive

ZJouba:
I like the second one as well, wasn't Boost also like that?

MV-GH:
Not really boost, doesn't display any numbers in the foot actionbar, neither the comments, only the action buttons, 
image


It is closer to the other apps though.

ZJouba:
Shall I close here in the mean time @MV-GH?

MV-GH:
yes

PR #1155, opened 2023-08-07
Post actionbar design rework and options for it.

### LINKED ISSUE #639: Action Bar Buttons get pushed off screen when comment count is too high

MV-GH:
Fixes #639

### DIFF
--- a/app/src/main/java/com/jerboa/db/AppDBMigrations.kt
+++ b/app/src/main/java/com/jerboa/db/AppDBMigrations.kt
@@ -296,6 +296,31 @@ val MIGRATION_22_23 = object : Migration(22, 23) {
     }
 }
 
+val MIGRATION_23_22 = object : Migration(23, 22) {
+    override fun migrate(database: SupportSQLiteDatabase) {
+        database.execSQL(
+            "ALTER TABLE AppSettings DROP COLUMN markAsReadOnScroll",
+        )
+    }
+}
+
+val MIGRATION_23_24 = object : Migration(23, 24) {
+    override fun migrate(database: SupportSQLiteDatabase) {
+        database.execSQL(UPDATE_APP_CHANGELOG_UNVIEWED)
+        database.execSQL(
+            "ALTER TABLE AppSettings ADD COLUMN post_actionbar_mode INTEGER NOT NULL DEFAULT 0",
+        )
+    }
+}
+
+val MIGRATION_24_23 = object : Migration(24, 23) {
+    override fun migrate(database: SupportSQLiteDatabase) {
+        database.execSQL(
+            "ALTER TABLE AppSettings DROP COLUMN post_actionbar_mode",
+        )
+    }
+}
+
 // Don't forget to test your migration with `./gradlew app:connectAndroidTest`
 val MIGRATIONS_LIST = arrayOf(
     MIGRATION_1_2,
@@ -321,4 +346,7 @@ val MIGRATIONS_LIST = arrayOf(
     MIGRATION_21_22,
     MIGRATION_22_21,
     MIGRATION_22_23,
+    MIGRATION_23_22,
+    MIGRATION_23_24,
+    MIGRATION_24_23,
 )
--- a/app/src/main/java/com/jerboa/feat/PostActionbarMode.kt
+++ b/app/src/main/java/com/jerboa/feat/PostActionbarMode.kt
@@ -3,7 +3,6 @@ package com.jerboa.feat
 import androidx.annotation.StringRes
 import com.jerboa.R
 
-
 enum class PostActionbarMode(@StringRes val resId: Int) {
     Long(R.string.post_actionbar_mode_long),
     LeftHandShort(R.string.post_actionbar_mode_short_left),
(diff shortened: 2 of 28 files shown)


MV-GH:
258817525-d46a72ea-0a15-49bc-8a44-802d55a0e207

To

(Only comment text changes)

258831200-418b565d-3ea8-49c7-87ad-ebd660c9a649

MV-GH:
Rearrange footing + change icons

short form

258850794-9408a33a-0c50-4c98-b3f3-2b616122b83f

Long form, Im leaning for this rn

image

MV-GH:
I'm thinking of making the post actionbar looks an option: (default long form), lefthand shortmode, righthand shortmode

dessalines:
I like that a lot, IE the *+6* , with the badge, plus, and italics inline.

I'd def prefer the vote buttons stay on the right side, to the right of the comment icon and counts, at least until a left-handed mode is added.

Either comments icon looks good to me.

MV-GH:
https://github.com/dessalines/jerboa/assets/67873169/dfc117b1-7724-4502-bd78-5e29da7adc89

Added the post actionbar mode settings. Do you still prefer comment to be to the left? @dessalines

dessalines:
That looks good, except that the `short righthand` should probably still left-align the comment info. Most of this was taken from boost, which looks like this:

Screenshot_2023-08-07-15-16-30-700_com rubenmayayo reddit

If that's not doable tho, or would overly-complicate the code, it'd be fine to right-align everything tho.

MV-GH:
image

Not that complex to implement. The lefthandside stays the same?
@dessalines

dessalines:
Nice, that looks great!

Ya, the left-hand can stay the same.

MV-GH:
Alright, I posted it in the community. We ll see feedback then this can be merged

app/src/main/java/com/jerboa/db/AppDBMigrations.kt
```diff
@@ -321,4 +346,7 @@ val MIGRATIONS_LIST = arrayOf(
     MIGRATION_21_22,
     MIGRATION_22_21,
     MIGRATION_22_23,
+    MIGRATION_23_22,
+    MIGRATION_23_24,
+    MIGRATION_24_23,
```
dessalines, line 351: I'm testing these now, but I've never done these down migrations in room. Aren't they going to undo the previous migration, or is that all taken care of by the database version?

app/src/main/java/com/jerboa/ui/components/settings/lookandfeel/LookAndFeelActivity.kt
```diff
@@ -330,6 +334,23 @@ fun LookAndFeelActivity(
                         )
                     },
                 )
+                SettingsList(
+                    title = {
```
dessalines, line 338: I know its somewhat arbitrary, but I think the settings page looks *much* better when the types are grouped. Could you move the `Confirm exit` and `Post Actionbar Mode`, belowe the `Post View` setting, so that all the list-type settings are grouped above the boolean settings?

app/src/main/java/com/jerboa/db/AppDBMigrations.kt
dessalines: Edit: seems to be taken care of by DB version.

MV-GH: They wont cause a downgrade, the version on instance is what causes which version we will be using, This list is just the path the migrator can use to change between versions. If there is no path then room db throws that error on startup that says no migration found. The order of this list doesn't even matter either. Just the version numbers. I just added them because it was trivial and you never know if we have a use case for them.

app/src/main/java/com/jerboa/ui/components/settings/lookandfeel/LookAndFeelActivity.kt
MV-GH: I moved them

app/src/main/java/com/jerboa/feat/PostActionbarMode.kt
```diff
@@ -0,0 +1,10 @@
+package com.jerboa.feat
+
+import androidx.annotation.StringRes
+import com.jerboa.R
+
+enum class PostActionbarMode(@StringRes val resId: Int) {
+    Long(R.string.post_actionbar_mode_long),
```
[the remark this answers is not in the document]
MV-GH: I preferred long tbh

I tweaked lefthand a bit not sure if its better or worse
afbeelding

app/src/main/java/com/jerboa/feat/PostActionbarMode.kt
```diff
@@ -0,0 +1,10 @@
+package com.jerboa.feat
+
+import androidx.annotation.StringRes
+import com.jerboa.R
+
+enum class PostActionbarMode(@StringRes val resId: Int) {
+    LeftHandShort(R.string.post_actionbar_mode_short_left),
```
dessalines, line 7: It should be RightHandShort first.

PR #1170, opened 2023-08-11
Fix statusbar color not updating with theme in certain Android Versions #661

### LINKED ISSUE #661: Black theme doesn't set the android status bar to black
** Jerboa Version **
0.0.33

**Describe the bug**
When in Black theme the status bar on android, remains the same colour as the background colour in Dark mode, rather than matching the theme.

**To Reproduce**

1. Go to 'Settings > Look and feel'
2. Click on 'Theme'
3. Select 'Black'
4. Observe the status bar

Schoolkid1:
Fix for Google Pixel phones


Schoolkid1:
It says All checks have failed, can some one help me out what is going wrong?

MV-GH:
`val toolbar.setBackgroundColor(0xFFFFB0CB)` is not valid code, so the check failed to compile the code

Schoolkid1:
I made a change in the code and committed the file and pushed it but if i want to make a pull request it says there is already a pull request, can I put the request into this request somehow?

MV-GH:
You make the change in this branch (ToolbarFix) then all you have to do is commit and push.

Schoolkid1:
If it is right it should have sended the changes

Schoolkid1:
I added the fix for the toolbar color. I added a TO-DO list below. I cannot do these changes because they are read only documents. 

// TO-DO: add parameter "toolbar: Color = ColorDarkTokens.Toolbar," in function "fun darkColorScheme" in file "ColorScheme.kt"
        // File ColorScheme.kt is read only! Don't forget to add the parameter!
        // Also add value "val Toolbar = PaletteTokens.Black" in file ColorScheme.kt 
        // Make the same changes in dynamicDarkColorScheme.kt, This file is also only read!

MV-GH:
Those "read only" files is the code from the library we are using, Android Studio let's inspect and see exactly which code is behind those library functions you are using.

You are also approaching this fix wrong.

First you need to able to reproduce the original issue, so that you can test your fix to see if it fixes it.
So I suggest downloading the exact same Android version mentioned in the issue in the emulator.

Secondly you are adding two color options to one specific theme. But we already have the color option it just doesn't seem to apply properly in that Android Version.

Now we are already setting the color of the status bar here: 
https://github.com/dessalines/jerboa/blob/9912366caa0af230fe79eb0c173fe00eddba82a8/app/src/main/java/com/jerboa/ui/theme/Theme.kt#L119

Now a possible fix might be to set using the systemUiController which provides support for older android versions
I use this already in the imageViewer
https://github.com/dessalines/jerboa/blob/9912366caa0af230fe79eb0c173fe00eddba82a8/app/src/main/java/com/jerboa/ui/components/imageviewer/ImageViewerActivity.kt#L95

But you would have to test that

dessalines:

PR #1171, opened 2023-08-11
Adding the unblock option

Schoolkid1:
when a community is already blocked the else function should set the block to false


MV-GH:
It fails the formatting check, see https://github.com/dessalines/jerboa/blob/main/CONTRIBUTING.md, in this case you have to run `./gradlew formatKotlin`

Schoolkid1:
Is there a specific prompt in which I can run this command, to be honest I don't know how to run that command in Android Studio...

MV-GH:
There is a terminal in android studio, run the above command in there. You can run it from any terminal/shell that has the repository as its working directory

Schoolkid1:
I found the terminal it gives me the following feedback. Is there something wrong with my java?

  - Incompatible because this component declares a component, compatible with Java 11 and the consumer needed a component, compatible with Java 8
              - Other compatible attribute:
                  - Doesn't say anything about org.gradle.plugin.api-version (required '8.2')
          - Variant 'sourcesElements' capability androidx.benchmark:benchmark-baseline-profile-gradle-plugin:1.2.0-alpha13 declares a component for use during runtime, and its dependencies declared externally:
              - Incompatible because this component declares documentation and the consumer needed a library

MV-GH:
What java version are you using?

Follow this https://stackoverflow.com/questions/72117858/incompatible-because-this-component-declares-a-component-compatible-with-java-11

and set it to JDK 17

Schoolkid1:
I am using java 8

The JDK version is 17.0.6 that should be sufficient right?

MV-GH:
Yeah i think that might be the problem, can you use java 17? or at least 11

MV-GH:
Also try this 
image

Schoolkid1:
I updated my java but i still get the following error:

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:formatKotlinMain'.
> error formatting sources for formatKotlinMain

MV-GH:
Is that the full error? Can you run app in a emulator or you device?

MV-GH:
Your code doesn't compile, it can't format invalid code. Make sure it compiles first

Schoolkid1:
I made some changes in the hope it does compile now, only i cannot check if it does from my laptop. Can you check if it does compile after this commit?

Schoolkid1:
Can some one please check with me what is going wrong i am out of ideas at this point

MV-GH:
Why can't you check if it compiles from your laptop?

Schoolkid1:
I really dont know.Normally i can use find usages on values and i see red lines when there is an error in the code but in this project i cannot use these features.  I think because the project is too big.

MV-GH:
Can you not press the run button at the top that tries to run the app? or build > make project?

Schoolkid1:
I think it has to do with compatibility. But when im working on other apps the version is no problem

Screenshot 2023-08-12 at 07 55 48

MV-GH:
Ah you need to be using at least android studio Giraffe.

Schoolkid1:
Thank you after updating i can finally run the app and gradle and look for bugs. I made some changes if you can check if it is sufficient at this point.

app/src/main/java/com/jerboa/api/Http.kt
```diff
@@ -233,6 +233,9 @@ interface API {
     @POST("community/block")
     suspend fun blockCommunity(@Body form: BlockCommunity): Response<BlockCommunityResponse>
 
+    @POST("community/unblock")
+    suspend fun unblockCommunity(@Body form: UnblockCommunity): Response<UnblockCommunityResponse>
```
MV-GH, line 237: There is no unblock API

app/src/main/java/com/jerboa/datatypes/types/UnblockCommunity.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import android.os.Parcelable
+import kotlinx.parcelize.Parcelize
+
+@Parcelize
+data class UnblockCommunity(
+    val community_id: CommunityId,
+    val block: Boolean,
+    val auth: String,
+) : Parcelable
```
MV-GH, line 11: We didn't really document this well but all the datatypes are auto generated from the API backend. You don't need to make them manually, you just need to use the right one. In this case I think you need to use the blocked API but set the blocked field to false.

app/src/main/java/com/jerboa/util/Initializable.kt
```diff
@@ -5,6 +5,7 @@ import androidx.compose.runtime.LaunchedEffect
 import kotlinx.coroutines.CoroutineScope
 
 interface Initializable {
+
```
MV-GH, line 8: Remove this line

MV-GH (changes requested):
Another thing I noticed is that you added this unblock functionality in the post options. But if the block is working properly the user would never come across any posts from that community and the user won't be able to unblock. For shortterm maybe for now you could show a short "snackbar" (popup with a button) after you block a community to unblock it. And longterm speaking we have to list somewhere all the communities a user has blocked where the user can unblock them.

Schoolkid1:
Hey i saw there were some changes requested, my school project is finished but i would like to keep working. I committed the changes and it says the checks have been passed can you see if it is sufficient?

Also i see there is a import file that is conflicting, i guess that is still about the datatype class right? But I deleted that one, do you know why it still pops up?

MV-GH:
Your branch is behind with the main branch, both have changes in a similar file which is why you have conflicts. Update your forks main branch with  `git pull upstream main` *or press sync fork in github* and merge the changes from your main into this branch. It will give a conflict dialog which you will have to resolve.

app/src/main/java/com/jerboa/api/Http.kt
```diff
@@ -233,6 +233,9 @@ interface API {
     @POST("community/block")
     suspend fun blockCommunity(@Body form: BlockCommunity): Response<BlockCommunityResponse>
 
+    @POST("community/unblock")
+    suspend fun unblockCommunity(@Body form: BlockCommunity): Response<BlockCommunityResponse>
```
MV-GH, line 237: Remove and just use the `blockCommunity` func

Schoolkid1:
The requested changes are processed in the new commit.
I synced the fork as mentioned and got a pop-up with sync is successful, I did not get a conflict dialogue there.

MV-GH:
After you sync your fork you still need to merge the new changes from main into your Fork. I already did that for you now.

dessalines:
Still failing formatting.

dessalines (commented):
This entire PR doesn't make any sense. If you've blocked a community, you will never see its posts or content to begin with after the initial block. This is why in lemmy-ui, you can only unblock communities from a section of your user settings, where you can view the communities you've blocked, and click an X to unblock them.

dessalines:
Closed in favor of #1185

PR #1175, opened 2023-08-12
Fix comment click deadzones

camporter:
Also makes it easier to make selections on comment source, and source now shows in monospace.

[Screen_recording_20230812_000054.webm](https://github.com/dessalines/jerboa/assets/494446/5adf8621-d93e-4ee4-9794-f96673e49455)

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/comment/CommentNode.kt
+++ b/app/src/main/java/com/jerboa/ui/components/comment/CommentNode.kt
@@ -170,6 +170,7 @@ fun CommentBody(
             markdown = content,
             onClick = onClick,
             onLongClick = onLongClick,
+            modifier = Modifier.padding(0.dp, 0.dp, 0.dp, MEDIUM_PADDING),
         )
     }
 }
--- a/app/src/main/java/com/jerboa/ui/components/comment/CommentNode.kt
+++ b/app/src/main/java/com/jerboa/ui/components/comment/CommentNode.kt
@@ -11,9 +11,7 @@ import androidx.compose.foundation.interaction.MutableInteractionSource
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
-import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxWidth
-import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.lazy.LazyListScope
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/jerboa/ui/components/comment/CommentNode.kt
```diff
@@ -315,12 +316,17 @@ fun LazyListScope.commentNodeItem(
...
+                                        }
+                                    },
                                     onLongClick = {
-                                        toggleActionBar(commentId)
+                                        if (!viewSource) {
+                                            toggleActionBar(commentId)
+                                        }
                                     },
                                 )
-                                Spacer(modifier = Modifier.height(MEDIUM_PADDING))
```
camporter, line 323: The combination of this unclickable spacer and the modifier with unclickable padding on the footer line below seems to have been a deadzone when clicking to collapse comments.

app/src/main/java/com/jerboa/ui/components/comment/CommentNode.kt
```diff
@@ -315,12 +316,17 @@ fun LazyListScope.commentNodeItem(
...
+                                    onClick = {
+                                        if (!viewSource) {
+                                            onCommentClick(commentView)
+                                        }
+                                    },
                                     onLongClick = {
-                                        toggleActionBar(commentId)
+                                        if (!viewSource) {
+                                            toggleActionBar(commentId)
+                                        }
```
MV-GH, line 327: What's this for? these click handlers are only for the actual text, they are not hooked on the viewSource text.

camporter: The click events for the comment node overall are triggered when clicking and long clicking while viewing the source of a comment. This is a bit frustrating when trying to copy parts of the source out because either the toolbar will disappear or the whole comment collapses, which resets selection.

MV-GH: Oh nvm they are actually bound to the `Text` remove this modifier instead 

```kt
modifier = Modifier.combinedClickable(
                    onClick = onClick,
                    onLongClick = onLongClick,
                ),
```

camporter: Ah yeah, that makes more sense :+1:

MV-GH:
Can you also change that the View Source text changes to View original after it is clicked like it does with the post

camporter:
I was avoiding touching the dialogs :smile: updated now.

MV-GH (approved):

Issue #1105, opened 2023-07-27
Controversial sorting for posts and comments


iByteABit256:
### Pre-Flight checklist

- [X] Did you check to see if this issue already exists?
- [X] This is a single feature request. (Do not put multiple feature requests in one issue)
- [X] This is not a question or discussion. (Use https://lemmy.ml/c/jerboa for that)

### Describe The Feature Request Below

Jerboa side of [this backend PR](https://github.com/LemmyNet/lemmy/pull/3205)

PR #1176, opened 2023-08-12
Use modal bottom sheets for sorting options.

camporter:
Experiment in changing sort options be modal bottom sheets rather than dialogs. Leaving draft since I'm not sure what the feeling is about bottom sheets vs menus and dialogs in this case.

Screenshot_20230812_022320
Screenshot_20230812_022353


MV-GH:
Hmm not fan, would have to be at least optional where you can switch, I think a fixed centered menu, would still look best. And i would prefer if this gets merged after #1105 so that these hardcoded sorts are gone.

dessalines:
Same, I'm not a fan of this either. The existing dropdown or centered popups look much better.

PR #1185, opened 2023-08-17
Update community block msg to display unblock if blocked already

MV-GH:
- Made text not hardcoded anymore
- Display error on failure to block
- Update view to show unblock if community is blocked

https://github.com/dessalines/jerboa/assets/67873169/520d9b43-de7c-43f2-b07b-648c5f5e7f67

I tried to do the same for postlistings but PostView does not contain a field for blocked community and mapping it to a datatype that does contain it pretty complex and could be done easier when the API rework gets done.
Also see https://github.com/LemmyNet/lemmy/issues/3885


dessalines:
I spose this is okay to add, but its still fairly pointless, because once you refresh the content or navigate away, you won't see the content to be able to unblock it. This is why in lemmy-ui, unblocking communities and persons are a section of your user settings.

dessalines (approved):

PR #1193, opened 2023-08-22
Show local mode on federation enabled

dessalines:


MV-GH:
Wouldn't be clear if federation is disabled that it only shows local instead?

dessalines:
Either one works since they're functionally the same. To me `All / Subscribed` makes more sense for a local-only instance, since local is implied when federation is turned off anyway. IE why would you need to specify *local*, on an instance that is inherently local-only. I'm taking what lemmy-ui shows.

MV-GH:
If lemmy-UI does that I am fine with it but, I do feel that `Local/Subscribed` is more intuitive, If there is no federation there is no all, no content from other instances. 

But what happens if the user sets local in the filter or in usersettings? Do we check for that? I think we should, it should filter local out from the options. It think this should be done like after #1105 so we don't hardcode this unnecessary

dessalines:
> But what happens if the user sets local in the filter or in usersettings?

If federation gets turned off, either setting is still possible, and lemmy's back-end should treat them both like local.

MV-GH:
So, if the user setting is set to local, it should display All for federation disabled instances?

dessalines:
If the server suddenly turns of federation, we don't update all those `default_listing_mode` columns in the user table, since both should function the same (and in case they want to turn federation back on again).

MV-GH:
Yes, I am not arguing that we should change the local data. Just that if it set to local and federation is disabled it should set it all instead of local in the viewmodel. Right now the viewmodel would contain local and that would show at the top while it prevents it from setting to that which kinda a discrepancy.

dessalines:
Stale PR.

PR #1194, opened 2023-08-23
Fix broken Japanese string format for AppBars_users_month

habbbe:
This caused a crash every time you entered Community Info if your phone language is set to Japanese.


MV-GH (approved):
Thanks for catching this!

ShinyLuxray:
Fuck. I think that bug was a mistake I made making the Ja translation. Thank you for fixing it, @habbbe ❤️.

PR #1205, opened 2023-08-31
Sort drawer subscriptions on title instead

### LINKED ISSUE #1203: Sort communities by displayed name rather than by invisible id
### Jerboa Version

0.0.44

### Android Version + Phone

Android 12, Fairphone 4

### Describe The Bug

When opening the side drawer (hamburger icon) the communities are sorted by their identification instead of by their display name, even though only the display name is displayed.

For example, the community "Haupteingang" is after "Linux" and before "Mikrocontroller", because main@feddit.de (which has the display name "Haupteingang") comes after linux@feddit.de and before mikrocontroller@feddit.de

### To Reproduce

1. Subscribe to linux@feddit.de, main@feddit.de and microcontroller@feddit.de
2. Open the side drawer
3. Communities are sorted in this order: "Linux", "Haupteingang", "Mikrocontroller"

It should be "Haupteingang", "Linux", "Microcontroller".

_No response_

MV-GH:
Fixes #1203

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/drawer/DrawerActivity.kt
+++ b/app/src/main/java/com/jerboa/ui/components/drawer/DrawerActivity.kt
@@ -49,7 +49,7 @@ fun MainDrawer(
                 if (!account.isAnon() && account.isReady() && res.data.my_user == null) {
                     accountViewModel.invalidateAccount(account)
                 }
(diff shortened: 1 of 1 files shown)


dessalines (approved):
Agree, that's better.

PR #1206, opened 2023-08-31
Add connect on matrix

MV-GH:
Add connect on matrix feature, this will open the relevant matrix client with the option to dm this user.

I designed the SVG icon of Matrix myself. So hereby I release the icon into the public domain

matrix_icon

Edit:
Someone linked me a smaller icon, I am using that one now.

image


app/src/main/java/com/jerboa/ui/components/common/DropdownMenu.kt
```diff
@@ -40,6 +40,29 @@ fun MenuItem(
...
 }
 
+@Composable
+fun MenuItem(
+    text: String,
+    onClick: () -> Unit,
+    modifier: Modifier = Modifier,
+    textModifier: Modifier = Modifier,
+    icon: @Composable (() -> Unit),
+    textStyle: TextStyle = MaterialTheme.typography.bodyMedium,
```
dessalines, line 50: Ah I see, mainly for the icon.

app/src/main/java/com/jerboa/ui/components/person/PersonProfile.kt
```diff
@@ -278,6 +289,7 @@ fun PersonProfileMoreDropdown(
     onBlockPersonClick: () -> Unit,
     onReportPersonClick: () -> Unit,
     onMessagePersonClick: () -> Unit,
+    openMatrix: (() -> Unit)?,
```
dessalines, line 292: Hrm.... not a fan of optional params, as they often result in places you don't catch them. Probably best would be 

```
hasMatrix: Boolean,
onOpenMatrixClick: () -> Unit
```

app/src/main/res/values/strings.xml
```diff
@@ -412,4 +412,5 @@
     <string name="open_link_external">Open link in external</string>
     <string name="no_activity_found">No activity (app) found that can open this link</string>
     <string name="markdownHelper_insertSpoiler">Insert spoiler</string>
+    <string name="matrix_send_msg">Connect on Matrix</string>
```
dessalines, line 415: I don't know that I've seen this phrase used before, but it makes sense, and is concise.

app/src/main/java/com/jerboa/ui/components/person/PersonProfile.kt
MV-GH: Optional params have indeed been a footgun but this param is not optional. You are forced to pass null or a lambda. Not doing will result in a compile error. *That would not be the case if I did `openMatrix: (() -> Unit)? = null`*

app/src/main/res/values/strings.xml
MV-GH: Yeah, I prefer it to `Send secure message through Matrix` or `Send a secure message`.

app/src/main/java/com/jerboa/ui/components/person/PersonProfile.kt
dessalines: mmmk cool.

Issue #1187, opened 2023-08-19
Error message stuck on top. 


olafur-palsson:
### Jerboa Version

0.0.44

### Android Version + Phone

Android 11, Xiaomi Redmi note 10 pro

### Describe The Bug

Clipped text on top 

"host lemmy.ml. No address associated with hostname"
![Screenshot_2023-08-19-10-53-08-113_com jerboa](https://github.com/dessalines/jerboa/assets/19432526/da8824cd-7edb-438a-bf65-981058c68d08)

### To Reproduce

1. Be offline
2. Load app
3. Go online
4. Load posts.

MV-GH:
Comes from here
https://github.com/dessalines/jerboa/blob/7a90523139bfc499c84b9ff0e22b965ed8978728/app/src/main/java/com/jerboa/ui/components/home/HomeActivity.kt#L218-L228

This error that it shows is intended, it means that It could not load the /site. Which Jerboa needs for lots of things

Jerboa has some checks on startup, you probably saw one of them. A snackbar which showed failed to load instance. This verification triggers for a few things and always on startup. 

Jerboa needs to also do the verification in this specific case. When it loads the post successfully while /site is in a bad state.

dessalines:
Can re-open if this is still an issue, but I don't know a good way to replicate.

MV-GH:
I fixed the case where it hid behind the actionbar. So everything should be as expected

PR #1209, opened 2023-09-01
Add retry button when loading more posts fails

### LINKED ISSUE #806: Reaching bottom of loaded posts sometimes doesn't load more
**Jerboa Version**
0.0.34

**Describe the bug**
sometimes when reaching the bottom of the main feed loaded posts, no more posts will load until you scroll a little bit up and then back down to the bottom.

I'm not exactly sure what's causing it so it may take a couple times reaching the bottom of the feed for the issue to appear, try scrolling down very fast.

**To Reproduce**
I think this could be caused by a timeout when trying to grab posts, try these steps on an overloaded instance to force a timeout

1. Scroll down on main feed to last loaded post until hitting bottom
2. If more posts load, scroll down to the next bottom of the feed
4. Scroll back up at least 2 posts on the feed
5. Scroll back down to the bottom of the feed (new posts should load)

### LINKED ISSUE #1003: If you get an error while scrolling the feed down, you cannot continue
**Jerboa Version**
0.0.38 Alpha from GitHub

**Android version + phone**
Pixel 7a, Android 13

**Describe the bug**
If you get an error from the server when scrolling down, there's no way to retry the request to see the next posts

**To Reproduce**

1. Scroll down the feed until it tries to load more posts
2. Get an error from the server
3. Realize there is no way to retry the request

MV-GH:
- Add button to retry loading more posts if it fails
- Removed page locking
- Add padding so that ApiError is not hiding behind header #1187 

This should provide a better UX than scrolling up and down again to initiate another attempt. The only thing I am not sure of is that should we auto scroll to that button? Most time it appears behind the bottom nav and only visible upon scrolling up.

YD1GONWFd5

Fixes #806
Fixes #1003

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/common/Buttons.kt
+++ b/app/src/main/java/com/jerboa/ui/components/common/Buttons.kt
@@ -1,0 +1,29 @@
+package com.jerboa.ui.components.common
+
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.material3.Button
+import androidx.compose.material3.ButtonDefaults
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.res.stringResource
+import com.jerboa.R
+import com.jerboa.ui.theme.XXL_PADDING
+
+@Composable
+fun RetryLoadingPosts(onClick: () -> Unit) {
+    Button(
+        modifier = Modifier
+            .fillMaxWidth()
+            .padding(horizontal = XXL_PADDING),
+        onClick = onClick,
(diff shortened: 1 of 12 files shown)


app/src/main/java/com/jerboa/ui/components/common/Buttons.kt
```diff
@@ -0,0 +1,29 @@
...
+import com.jerboa.ui.theme.XXL_PADDING
+
+@Composable
+fun RetryLoadingPosts(onClick: () -> Unit) {
+    Button(
+        modifier = Modifier
+            .fillMaxWidth()
+            .padding(horizontal = XXL_PADDING),
+        onClick = onClick,
+        colors = ButtonDefaults.buttonColors(
```
dessalines, line 22: Ah, just to force it as an error color.

dessalines (approved):
Looks good, thx!

PR #1213, opened 2023-09-03
Fix crash in share Media, fix failed download media

### LINKED ISSUE #1204: Sometimes crash when sharing images
### Jerboa Version

0.0.45

### Android Version + Phone

N/A

### Describe The Bug

Sometimes sharing the image causes a crash, but if you do it from the image viewer it does not crash. But both call the exact same function.

Another weird thing is that I have a catch for IOExceptions but somehow it doesn't catch this error

```kt

fun shareMedia(scope: CoroutineScope, ctx: Context, url: String, mediaType: PostType) {
    try {
        val fileName = Uri.parse(url).pathSegments.last()

        val file = File(ctx.cacheDir, fileName)

        scope.launch(Dispatchers.IO) {
            ctx.getInputStream(url).use { input ->
                file.outputStream().use { // From what I can tell this line throws the error
                    input.copyTo(it)
                }
            }
        }

        val uri = FileProvider.getUriForFile(ctx, ctx.packageName + ".provider", file)
        val shareIntent = Intent()
        shareIntent.setAction(Intent.ACTION_SEND)
        shareIntent.putExtra(Intent.EXTRA_STREAM, uri)
        when (mediaType) {
            PostType.Image -> shareIntent.setType("image/*")
            PostType.Video -> shareIntent.setType("video/*")
            PostType.Link -> shareIntent.setType("text/*")
        }
        shareIntent.addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION)
        ctx.startActivitySafe(Intent.createChooser(shareIntent, ctx.getString(R.string.share)))
    } catch (e: IOException) {
        Log.d("shareMedia", "failed", e)
        Toast.makeText(ctx, R.string.failed_saving_media, Toast.LENGTH_SHORT).show()
    } catch (e: Exception) {
        Log.d("shareMedia", "invalid URL", e)
        Toast.makeText(ctx, R.string.failed_saving_media, Toast.LENGTH_SHORT).show()
    }
}
```

### To Reproduce

Long press image then share it, sometimes crash

FATAL EXCEPTION: DefaultDispatcher-worker-9
Process: com.jerboa, PID: 24315
java.io.IOException: closed
	at okio.Buffer$inputStream$1.read(SourceFile:10)
	at java.io.InputStream.read(InputStream.java:101)
	at okio.Utf8.copyTo$default(Unknown Source:9)
	at com.jerboa.feat.UserActionsKt$shareMedia$1.invokeSuspend(Unknown Source:20)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(Unknown Source:8)
	at kotlinx.coroutines.DispatchedTask.run(Unknown Source:109)
	at androidx.core.app.ActivityRecreator$1.run(Unknown Source:15)
	at kotlinx.coroutines.scheduling.TaskImpl.run(Unknown Source:2)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(Unknown Source:93)
	Suppressed: kotlinx.coroutines.internal.DiagnosticCoroutineContextException: [androidx.compose.ui.platform.MotionDurationScaleImpl@f28cf90, androidx.compose.runtime.BroadcastFrameClock@cfe7089, StandaloneCoroutine{Cancelling}@195d88e, Dispatchers.IO]
PerfMonitor installContentProviders : before publish content providers

MV-GH:
Learned quite a bit from this. The original share/download since the introduction of the caching check was only really working in some instances due to some quirks. 

mistakes:
- scope.launch won't rethrow unhandled exceptions but will cancel its parents thus bypassing the try catch I had
- shareMedia used scope.launch for the IO part but it did not actually wait for this complete, thus it was sharing empty files at best crash at worst due to two sources reading/write to a cache file (only really worked in the rare instance its called from imageviewer where the image is loaded in cache and it wins the race condition on reading the file into the cached file)
-  Context.getInputStream shared a closed inputStream, the implementation for snapshot doesn't actually check if it is closed thus that worked. But the byteStream implementation did. Thus the cause of the `IOException: Closed` exception

Fixes #1204

### DIFF
--- a/app/src/main/java/com/jerboa/Utils.kt
+++ b/app/src/main/java/com/jerboa/Utils.kt
@@ -1471,15 +1471,21 @@ fun <I, O> ComponentActivity.registerActivityResultLauncher(
 
 /**
  *  Returns a [InputStream] for the data of the URL, but it also checks the cache first!
+ *
+ *  Doesn't clean up the [InputStream]
+ *
+ *  @throws IOException
+ *  @throws IllegalArgumentException If this is not a well-formed HTTP or HTTPS URL.
  */
 @OptIn(ExperimentalCoilApi::class)
+@Throws(IOException::class)
 fun Context.getInputStream(url: String): InputStream {
     val snapshot = this.imageLoader.diskCache?.openSnapshot(url)
 
-    return snapshot?.use {
-        it.data.toFile().inputStream()
-    } ?: API.httpClient.newCall(Request(url.toHttpUrl())).execute().use { response ->
-        response.body.byteStream()
+    return if (snapshot != null) {
+        snapshot.data.toFile().inputStream()
+    } else {
+        API.httpClient.newCall(Request(url.toHttpUrl())).execute().body.byteStream()
     }
 }
 
--- a/app/src/main/java/com/jerboa/feat/UserActions.kt
+++ b/app/src/main/java/com/jerboa/feat/UserActions.kt
@@ -82,36 +82,38 @@ private suspend fun saveMedia(url: String, context: Context, mediaType: PostType
  * Shares the actual file from the link
  */
 fun shareMedia(scope: CoroutineScope, ctx: Context, url: String, mediaType: PostType) {
-    try {
-        val fileName = Uri.parse(url).pathSegments.last()
+    scope.launch(Dispatchers.Main) {
+        try {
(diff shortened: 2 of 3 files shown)


dessalines (approved):
LGTM

PR #1220, opened 2023-09-05
Fix PullRefreshIndicator theming and placement

### LINKED ISSUE #1011: Swipe to refresh - refresh icon doesn't account for top bar
Hi! I noticed a small visual oddity and figured I'd log it here - I noticed when swiping down on the feed, the circular refresh arrow takes a while to show up? As in, I needed to swipe further down than I'd have expected to make it show up. I believe that this icon is currently set to animate down from the top of the screen, rather than immediately emerging from behind the top bar. Because of this, I frequently refresh the feed having only seen a small portion of the icon, and only for a brief second.

I don't know how easy of a fix it would be, but hopefully updating the refresh icon's starting position to account for the rest of the UI would fix this :)

### LINKED ISSUE #1095: Pull to refresh indicator does not adopt the theme
### Jerboa Version

0.0.41

### Android Version + Phone

11

### Describe The Bug

Pull to refresh indicator does not adopt the theme, it is white with black inner thing on every theme, it should be black with red inner thing according to this theme
studio64_nOc51JxrEc

### To Reproduce

pull the refresh

_No response_

MV-GH:
PullRefreshingIndicator uses Material thus it doesn't work with M3 Theming. So I fixed it.
Also fixed the placement being weird. Instead of doing that offset I had to just add the padding instead. Only on some pullrefreshindicators if they have a header. Those pagers seems to work fine out the box.

image

Fixes #1095 
Fixes #1011

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/common/AppBars.kt
+++ b/app/src/main/java/com/jerboa/ui/components/common/AppBars.kt
@@ -14,10 +14,13 @@ import androidx.compose.foundation.layout.FlowRow
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.LazyListState
 import androidx.compose.foundation.lazy.rememberLazyListState
+import androidx.compose.material.ExperimentalMaterialApi
 import androidx.compose.material.icons.Icons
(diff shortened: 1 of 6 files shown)


dessalines (approved):
Weird, hopefully they'll update / add one at some point.

PR #1223, opened 2023-09-07
Dirty Fix for #1222

0xFOSSMan:
Given my limited experience with Kotlin this maybe complete nonsense but I'll try it anyway


0xFOSSMan:
oh damn i didn't want to put it here, does that break anything with me themeing?

MV-GH:
Not there, look in siteviewmodel where it sets unreadcount when the account changes in the init. There we check if ur not anon and set unread count. Change it so if ur anon that it sets it to 0

MV-GH (changes requested):
see above

0xFOSSMan:
i'll set this fork up again to split that up. thx for the tip

PR #1224, opened 2023-09-07
Possible fix for #1222

0xFOSSMan:
Resetting the inbox count when switching to anon

### DIFF
--- a/app/src/main/java/com/jerboa/model/SiteViewModel.kt
+++ b/app/src/main/java/com/jerboa/model/SiteViewModel.kt
@@ -42,7 +42,7 @@ class SiteViewModel(private val accountRepository: AccountRepository) : ViewMode
 
     private var unreadCountRes: ApiState<GetUnreadCountResponse> by mutableStateOf(ApiState.Empty)
 
-    val unreadCount by derivedStateOf { getUnreadCountTotal(unreadCountRes) }
+    val unreadCount by derivedStateOf { getUnreadCountTotal(unreadCountRes, it.isAnon()) }
 
     init {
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/jerboa/model/SiteViewModel.kt
```diff
@@ -66,6 +66,8 @@ class SiteViewModel(private val accountRepository: AccountRepository) : ViewMode
 
                     if (!it.isAnon()) {
                         fetchUnreadCounts(GetUnreadCount(auth = it.jwt))
+                    } else {
+                        unreadCount = 0
```
dessalines, line 70: There is already a `isAnon` block above. And instead of manually trying to force an immutable var, trace down where unread count is read, and make it zero when its an anonymous account. Or creat a `getUnreadCount(isAnon: boolean)` type function.

MV-GH (commented):
Ah I forgot I made, unreadcount a derivedstate. In the function do a if check for anon and return 0 else let it do the when stuff

0xFOSSMan:
did I change the wrong function?

MV-GH:
No but this line doesn't make sense

      `  return when (val res = unreadCountRes && !isAnon) {`

Did you test this?

0xFOSSMan:
I'll abandon this. I need to learn Kotlin more and before my or your time it's better to let people fix it that can actually be productive.

MV-GH:
You were doing a conjunction of a bool and ApiState wrapper for a response and then matching on a concrete ApiState

Also u were using the implicit `it` for a context that does not have it.

Issue #957, opened 2023-07-03
Add a second NSFW Blur toggle which applies only to NSFW communities


MrMusAddict:
- [x] Did you check to see if this issue already exists?
- [x] Is this only a single feature request? Do not put multiple feature requests in one issue.
- [x] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/jerboa

** Describe the feature request below **

Ideally it'd be great if we could keep NSFW content blurred while browsing mixed communities (all, local, non-NSFW communities that allow NSFW posts), BUT to not blur NSFW content while in a NSFW community. 

I'm thinking we should have 2 toggles;

1. Blur NSFW content on non-NSFW communities, and mixed content pages (all / local, etc)
2. Blur NSFW content on NSFW communities

Or, perhaps a single drop-down to select one of the 3 options;

1. Do not blur
2. Blur everywhere except NSFW Communities
3. Blur everywhere

MakcNmyc:
It's also needed change nsfw type on server

dessalines:
Fixed by above.

PR #1229, opened 2023-09-10
Add BlurNsfwExceptFromNsfwCommunities to blur types

MakcNmyc:
My first pull request. I will be grateful for every advice


MakcNmyc:
It's also needed change nsfw type on server

app/schemas/com.jerboa.db.AppDB/26.json
```diff
@@ -2,7 +2,7 @@
   "formatVersion": 1,
   "database": {
     "version": 26,
-    "identityHash": "bf20da38fc7d03e28121589dfb372eaf",
+    "identityHash": "98ab20c6cd84e7fedbb9eccc2e4c44b1",
```
MV-GH, line 5: Do not change the existing scheme, up the version and push the new version.

I think it's somewhere in AppDB

MV-GH (changes requested):

app/src/main/java/com/jerboa/datatypes/types/Others.kt
```diff
@@ -382,6 +382,22 @@ enum class ModlogActionType {
     AdminPurgeComment,
 }
 
+/**
+ * Different show nsfw states
+ */
+enum class ShowNsfwTypes {
```
MV-GH, line 388: Move this to Feat.kt remove serialize stuff l. That is only needed for custom stuff we get from from the server. This exists only client side thus doesn't need it. And preferable if possible make it Parcelable

app/src/main/java/com/jerboa/datatypes/types/Others.kt
```diff
@@ -382,6 +382,22 @@ enum class ModlogActionType {
...
+
+    @SerializedName("BlurEverywhere")
+    BlurEverywhere,
+
+    @SerializedName("BlurEverywhereExceptNsfw")
```
MV-GH, line 393: Also while ur doing this also add blur only nsfw Community options

app/src/main/java/com/jerboa/db/AppDBMigrations.kt
```diff
@@ -354,6 +354,123 @@ val MIGRATION_26_25 = object : Migration(26, 25) {
     }
 }
 
+val MIGRATION_26_27 = object : Migration(26, 27) {
```
MV-GH, line 357: I am not home and doing this on my phone so i can't properly vet this SQL migration. But does this pass the migration test?

app/src/main/java/com/jerboa/datatypes/types/Others.kt
```diff
@@ -382,6 +382,22 @@ enum class ModlogActionType {
...
+    BlurEverywhereExceptNsfw,
+
+    @SerializedName("DoNotBlur")
+    DoNotBlur;
```
MV-GH, line 397: Also change the order of this from doNotBlur, BlurEverywhere, other.

This keeps the order the same as before (index order) This should allow you to simply the migration by a lot. I am honestly not sure why you were doing so much changes there. But I can't really check it atm.

app/src/main/java/com/jerboa/ui/components/settings/lookandfeel/LookAndFeelActivity.kt
```diff
@@ -353,12 +354,26 @@ fun LookAndFeelActivity(
                     },
                     onCheckedChange = { updateAppSettings() },
                 )
-                SettingsCheckbox(
+                SettingsList(
                     state = blurNSFW,
+                    items = listOf(
+                        stringResource(R.string.account_settings_blur_everywhere),
+                        stringResource(R.string.account_settings_blur_everywhere_except_nsfw),
+                        stringResource(R.string.account_settings_do_not_blur),
```
MV-GH, line 362: Make sure this follows the order of the enum

MV-GH (changes requested):

MV-GH:
> It's also needed change nsfw type on server

What do you mean by this? Currently it is not even saved server side.

MV-GH:
Also move the listsetting to be below the other list settings

MakcNmyc:
> > It's also needed change nsfw type on server
> 
> What do you mean by this? Currently it is not even saved server side.

I'm sorry. Perhaps I misunderstand how this works. I saw that the data in AccountSettings.kt is taken from siteViewModel.siteRes. There they are initialized as API.getInstance().getSite. Could you tell me where I can see how it works?
My response from the API server - lemmynsfw.com

MakcNmyc:
Made changes

app/schemas/com.jerboa.db.AppDB/26.json
MakcNmyc: Can you explain this more please? Now I did this without changing the base, but I'm still curious what you meant.

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
```diff
@@ -1380,7 +1381,7 @@ private fun ThumbnailTile(
             postView.post.thumbnail_url?.also { thumbnail ->
                 PictrsThumbnailImage(
                     thumbnail = thumbnail,
-                    blur = blurNSFW && nsfwCheck(postView),
+                    blur = needNsfwBlur(blurNSFW.toEnum(), postView.community.nsfw, postView.post.nsfw),
```
MV-GH, line 1384: Add override for postview that calls this basically. And remove all occurances of this

app/src/main/java/com/jerboa/ui/components/settings/lookandfeel/LookAndFeelActivity.kt
```diff
@@ -381,6 +374,17 @@ fun LookAndFeelActivity(
                     },
                     onCheckedChange = { updateAppSettings() },
                 )
+                SettingsListDropdown(
```
MV-GH, line 377: This still needs to be moved below the other settingslist

MV-GH (changes requested):
Looking better

MakcNmyc:
Made changes

app/src/main/java/com/jerboa/feat/BlurNsfwTypes.kt
```diff
@@ -0,0 +1,28 @@
...
+    DoNotBlur(R.string.app_settings_do_not_blur),
+    BlurEverywhere(R.string.app_settings_blur_everywhere),
+    BlurEverywhereExceptNsfw(R.string.app_settings_blur_everywhere_except_nsfw),
+    BlurOnlyNsfwCommunity(R.string.app_settings_blur_only_nsfw_community);
+}
+
+fun Int.needNsfwBlur(postView : PostView) =
+    this.needNsfwBlur(postView.community.nsfw, postView.post.nsfw)
+
+fun Int.needNsfwBlur(isCommunityNsfw: Boolean, isPostNsfw: Boolean = isCommunityNsfw): Boolean {
```
MV-GH, line 21: Can it be the enum nsfwTypes instead of Int that you extend?

MV-GH (changes requested):
Other than that it looks good but until I am home I can't verify if the actual functionality works.

app/src/main/java/com/jerboa/feat/BlurNsfwTypes.kt
MakcNmyc: @MV-GH do you shure? I think it's 
blurNSFW.needNsfwBlur(communityView.community.nsfw)
looks better than 
blurNSFW.toEnum<BlurNsfwTypes>().needNsfwBlur(communityView.community.nsfw)

MV-GH: Yeah, it's not about looking cleaner. The original pollutes autocompletion for all Int. Instead for just BlurNfwsType

MakcNmyc:
Made changes

app/src/main/java/com/jerboa/feat/BlurNsfwTypes.kt
```diff
@@ -0,0 +1,28 @@
...
+    BlurEverywhereExceptNsfw(R.string.app_settings_blur_everywhere_except_nsfw),
+    BlurOnlyNsfwCommunity(R.string.app_settings_blur_only_nsfw_community);
+}
+
+fun BlurNsfwTypes.needNsfwBlur(postView : PostView) =
+    this.needNsfwBlur(postView.community.nsfw, postView.post.nsfw)
+
+fun BlurNsfwTypes.needNsfwBlur(isCommunityNsfw: Boolean, isPostNsfw: Boolean = isCommunityNsfw): Boolean {
+    return isPostNsfw && when(this){
+        BlurNsfwTypes.DoNotBlur -> false
```
MV-GH, line 23: Possible to change this to:
- Nothing
- All
- OnlyInsideNsfwCommunities
- OnlyOutsideNsfwCommunities

These are shorter and more descriptive, we already know they blur from the EnumName

Change the translation to fit these new names too.

MV-GH (changes requested):
Thanks for the changes, I will have tmmr some time to physically test these changes

MakcNmyc:
Made changes

app/schemas/com.jerboa.db.AppDB/26.json
MV-GH: you changed the scheme before upping the version, (db classes). This caused it to generated a new scheme for 26. This scheme is needed to test the migrations. The test for example upgrades a database from 25 to 26 and then compares it to this json scheme format. Well If it contains changes from 27 it will no longer be accurate and the test will fail

MV-GH:
Alright, I have confirmed the functionality and added my changes:

- I moved the setting to bottom of the SettingsList
- Added a Icon for the SettingsList option
- Changed naming for the options again (This is clearer and shorter imo)
- Probably by confusion before I had named them opposite (or you had the logic reversed)
- Fixed the build errors due to referencings outdated stringRes. Now we use it directly from the enum to satisfy SSOT principles
- Included the changes from main (Keep your PR upto date with main, smaller but more often merges keep the complexity lower)

MV-GH (approved):
I have tested the functionality and did some changes

MV-GH:
Since I have made some changes, @dessalines, could you review my changes

dessalines:
Sure, testing now.

app/src/main/java/com/jerboa/feat/BlurNsfwTypes.kt
```diff
@@ -0,0 +1,27 @@
+package com.jerboa.feat
+
+import android.os.Parcelable
+import androidx.annotation.StringRes
+import com.jerboa.R
+import com.jerboa.datatypes.types.PostView
+import kotlinx.parcelize.Parcelize
+
+@Parcelize
+enum class BlurNsfwTypes(@StringRes val resId: Int) : Parcelable {
```
dessalines, line 10: Rename this to `BlurNsfw` ... but again this is going to conflict with your real `blur_nsfw` user setting saved in the back end.

app/src/main/java/com/jerboa/feat/BlurNsfwTypes.kt
```diff
@@ -0,0 +1,27 @@
...
+    Nothing(R.string.app_settings_nothing),
+    All(R.string.account_settings_all),
+    OnlyFromNsfwCommunity(R.string.app_settings_blur_only_from_nsfw_community),
+    OnlyFromSfwCommunity(R.string.app_settings_blur_only_from_sfw_community),
+}
+
+fun BlurNsfwTypes.needNsfwBlur(postView: PostView) =
```
dessalines, line 17: These two functions you can probably rename to just `blur` or `isBlurred` . See comment below for the options

app/src/main/java/com/jerboa/feat/BlurNsfwTypes.kt
```diff
@@ -0,0 +1,27 @@
...
+import androidx.annotation.StringRes
+import com.jerboa.R
+import com.jerboa.datatypes.types.PostView
+import kotlinx.parcelize.Parcelize
+
+@Parcelize
+enum class BlurNsfwTypes(@StringRes val resId: Int) : Parcelable {
+    Nothing(R.string.app_settings_nothing),
+    All(R.string.account_settings_all),
+    OnlyFromNsfwCommunity(R.string.app_settings_blur_only_from_nsfw_community),
```
dessalines, line 13: Change to `OnlyNsfwCommunity`

app/src/main/java/com/jerboa/feat/BlurNsfwTypes.kt
```diff
@@ -0,0 +1,27 @@
...
+    OnlyFromSfwCommunity(R.string.app_settings_blur_only_from_sfw_community),
```
dessalines, line 14: Not sure I understand this one. Only blur items from safe communities? I don't think that needs to be an option.

app/src/main/java/com/jerboa/db/AppDB.kt
```diff
@@ -32,7 +32,7 @@ val APP_SETTINGS_DEFAULT = AppSettings(
     useCustomTabs = true,
     usePrivateTabs = false,
     secureWindow = false,
-    blurNSFW = true,
+    blurNSFW = 1,
```
dessalines, line 35: blur_nsfw is a `LocalUser` / back end setting, so it should probably not be here. Otherwise they're duped, and one is ignored.

app/src/main/java/com/jerboa/feat/BlurNsfwTypes.kt
```diff
@@ -0,0 +1,27 @@
...
+}
+
+fun BlurNsfwTypes.needNsfwBlur(postView: PostView) =
+    this.needNsfwBlur(postView.community.nsfw, postView.post.nsfw)
+
+fun BlurNsfwTypes.needNsfwBlur(isCommunityNsfw: Boolean, isPostNsfw: Boolean = isCommunityNsfw): Boolean {
+    return isPostNsfw && when (this) {
```
dessalines, line 21: The logic here doesn't follow. We already have a `blur_nsfw` setting in the back-end. And that should blur any NSFW post or NSFW community posts. I don't see why other options would be necessary.

The *only* other case I could think of, would be  a third option which is to blur everything, but that should probably be handled in the back end.

dessalines (commented):
I'll allow myself to be overruled, but I don't think this is the right way to handle this. We already have a back-end `blur_nsfw` user setting, that appears to be ignored right now in favor of an app-specific setting, which we should do away with / deprecate.

I also don't see the additional blur types making too much sense. There's a blur_nsfw boolean, and the only additional one that might make sense, is the addition in the back end of a `blur_all_previews` setting.

app/src/main/java/com/jerboa/db/AppDB.kt
MV-GH: That setting existed before the backend option was added. I am not sure when it was added but we would need to implement a way to keep this backwards compatible with previous versions. So until then, it is better to just keep it local

app/src/main/java/com/jerboa/feat/BlurNsfwTypes.kt
MV-GH: That one means only blur NSFW posts from SFW communities.

app/src/main/java/com/jerboa/feat/BlurNsfwTypes.kt
MV-GH: This PR was made to solve 
https://github.com/dessalines/jerboa/issues/957

Which wants a more finegrained control over which images get blurred

MV-GH: But I don't think this PR fully solves that FR, as that one wants to only not blur while it is in a communty mode view and still blur NSFW images that are in the feed

MV-GH: Also this option does exists in other Reddit clients like IfR

dessalines: We should at least get the options figured out. The ones I can make sense of

- Blur everything
- Blur nothing
- Blur NSFW
- Blur NSFW except from NSFW communities ( I think that handles #957 , but I can't really tell)

MV-GH: If we are doing Blur everything then, the title for the option should change from `Blur NSFW images` to `Blur images`

@MakcNmyc What do you think of this direction since its your PR

MakcNmyc: Honestly, I don't catch idea why anyone would want to blur all posts from their feed. When I started working on the issue, I thought that we only needed to add an option for those who want to blur too explicit content (in NSFW communities), but save not quite hot (outside NSFW communities, but still NSFW). 
But I just started working on this project, I just started learning how to work in the GitHub open source community. And I’m ready to accept and correct the commit, no matter what you choose.

MV-GH: `Blur everything` could be covered by #1239 where you can disable loading all images instead of just blurring them all instead

I would prefer that over extending the scope of this feature from NSFW to all images. What do you think @dessalines ?

Then we are back to 3 options

dessalines: That sounds good to me.

MV-GH: @MakcNmyc Do you want to make that change? or else I'll do it

MakcNmyc: > But I don't think this PR fully solves that FR, as that one wants to only not blur while it is in a communty mode view and still blur NSFW images that are in the feed

@MV-GH  Maybe fix it to work this way?

MakcNmyc: > @MakcNmyc Do you want to make that change? or else I'll do it

I will do this if you explain what exactly needs to be done. Do you need the options to work for all images (not just nsfw)? But how will “NSFW Blur other than NSFW Communities” work?

dessalines: Address my comments above, but also make sure there are only 3 enum types, and that they make logical sense (explained above).

MakcNmyc: I did the second option with some changes

MakcNmyc:
Made changes

app/src/main/java/com/jerboa/feat/BlurTypes.kt
```diff
@@ -0,0 +1,45 @@
...
+    NsfwExceptFromNsfwCommunities(R.string.app_settings_blur_nsfw_except_from_nsfw_communities),
+    ExceptFromNsfwInsideCommunity(R.string.app_settings_blur_nsfw_except_from_nsfw_inside_community);
+
+    companion object {
+        fun usersEntries() : List<BlurTypes> = entries.filter {
+            it != ExceptFromNsfwInsideCommunity
+        }
+
+        fun changeBlurTypeInsideCommunity(blurTypes: Int): Int =
+            if (blurTypes.toEnum<BlurTypes>() == NsfwExceptFromNsfwCommunities) {
```
MV-GH, line 24: compare on ordinal

MV-GH: and blurType

app/src/main/java/com/jerboa/feat/BlurTypes.kt
```diff
@@ -0,0 +1,45 @@
...
+import com.jerboa.R
+import com.jerboa.datatypes.types.PostView
+import com.jerboa.toEnum
+import kotlinx.parcelize.Parcelize
+
+@Parcelize
+enum class BlurTypes(@StringRes val resId: Int) : Parcelable {
+    Nothing(R.string.app_settings_nothing),
+    NSFW(R.string.app_settings_blur_nsfw),
+    NsfwExceptFromNsfwCommunities(R.string.app_settings_blur_nsfw_except_from_nsfw_communities),
```
MV-GH, line 15: Remove this type, as discussed we keep only, blur nothing, blur nsfw and blur nsfw but not in community view

MakcNmyc: @MV-GH I didn't understand you, only these 3 options remain. 4 option (ExceptFromNsfwInsideCommunity) it technical option. In community activity I swap  NsfwExceptFromNsfwCommunities option for it.

MV-GH: That's not how I meant that second approach. The second approach was, turn NSFWExceptInsideCommunity into Nothing if passed to communityActiviity else convert it to NSFW and then you only have to handle those two cases

MakcNmyc: > else convert it to NSFW and then you only have to handle those two cases

I can't figure out how and where to do this. In Main Activity for every Activity?

MV-GH: yes but you can do it once under appsetting and pass that. except for community activity where you then pass the other option. But first I have to know if Dessalines would accept such approach.

@dessalines Would you be against the above approach or do you rather that we have a extra param indicating that this is community view. Or do you have a better approach in mind?

dessalines: > NSFWExceptInsideCommunity into Nothing if passed to communityActiviity

Yes that seems fine. No point in adding an extra enum when it can figure that out by context, in the CommunityActivity.

MakcNmyc: @MV-GH Am I understanding correctly that "NsfwExceptFromNsfwCommunities" works as "Nothing" only for nsfw communities, otherwise it works as "NSFW"? If this is true then I can't change it in MainActivity because I need the community type.

MV-GH: For what do you need the communityType? You map it there either to Nothing or NSFW ordinals depending if its for CommunityView

MakcNmyc: @MV-GH what behavior should there be with the nsfw posts if Community.nsfw = false and blur option = "NsfwExceptFromNsfwCommunities"?

MV-GH: Remove the community nsfw option, we don't care anymore if the post comes from a nsfw community or not. And if the function receives NsfwExceptFromNsfwCommunity then throw with message likr this type should ve already been transformed into another

MV-GH:
You still need to resolve the conflicts,

dessalines:
Once CI passes I'll test to make sure its good too.

MakcNmyc:
fixed

dessalines (commented):
Lots of errors in `PostListing.kt`

MakcNmyc:
@dessalines fixed

app/src/main/java/com/jerboa/ui/components/community/CommunityActivity.kt
```diff
@@ -82,7 +83,7 @@ fun CommunityActivity(
     showVotingArrowsInListView: Boolean,
     useCustomTabs: Boolean,
     usePrivateTabs: Boolean,
-    blurNSFW: Boolean,
+    blurNSFW: Int,
```
dessalines, line 86: Why can't these use the BlurType enum?

MV-GH: We don't do it either for other enums, see postActionMode below it

dessalines (approved):
K I tested and its working fine. @MV-GH you're still on "request changes" btw.

MakcNmyc:
@MV-GH Do I need to do anything else?

MV-GH:
Lemme test this and I'll merge

MV-GH (approved):

PR #1233, opened 2023-09-14
Fix unread count not resetting when swapping to anon

### LINKED ISSUE #1222: Swapping to anon doesn't reset inbox counter

MV-GH:
Fixes #1222 

@0xFOSSMan 

Sorry about before but this one can be trivially solved by resetting the unreadCount response to empty again.

Since we using derivativedState it will trigger on changes to unreadCountRes. Where `getUnreadCountTotal` sets it to zero for that ApiState


0xFOSSMan:
thx for the info and don't worry, I am just starting out and it's okay to make mistakes

dessalines (approved):

PR #1235, opened 2023-09-14
Fix/1231

### LINKED ISSUE #1231: Properly handle datetime parse failures
See #1230

MV-GH:
Fixes #1231

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/common/TimeAgo.kt
+++ b/app/src/main/java/com/jerboa/ui/components/common/TimeAgo.kt
@@ -95,7 +95,6 @@ fun TimeAgo(
  * @param longTimeFormat If true, use a long time format like "2 hours, 3 minutes ago"
  * @return The pretty string, or null if the date string could not be parsed
  */
-@Throws(DateTimeParseException::class)
 fun dateStringToPretty(dateStr: String, longTimeFormat: Boolean = false): String? {
     return try {
         // TODO: Remove this hack once backward API compatibility is implemented
--- a/app/src/main/java/com/jerboa/ui/components/common/TimeAgo.kt
+++ b/app/src/main/java/com/jerboa/ui/components/common/TimeAgo.kt
@@ -108,7 +108,7 @@ fun dateStringToPretty(dateStr: String, longTimeFormat: Boolean = false): String
     }
 }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/jerboa/ui/components/common/TimeAgo.kt
```diff
@@ -40,6 +42,15 @@ fun TimeAgo(
 ) {
     val publishedPretty = dateStringToPretty(published, longTimeFormat)
 
+    if (publishedPretty == null) {
+        Text(
```
dessalines, line 46: Since this is used twice, you might as well extract it into a composable.

app/src/main/java/com/jerboa/ui/components/common/TimeAgo.kt
```diff
@@ -52,25 +63,48 @@ fun TimeAgo(
...
+/**
+ * Converts a date string to a pretty string like "2 hours" or "2h"
+ *
+ * @param dateStr The date string to convert
+ * @param longTimeFormat If true, use a long time format like "2 hours, 3 minutes ago"
+ * @return The pretty string, or null if the date string could not be parsed
+ */
+fun dateStringToPretty(dateStr: String, longTimeFormat: Boolean = false): String? {
+    return try {
+        // TODO: Remove this hack once backward API compatibility is implemented
```
dessalines, line 99: I'm wondering if we should create some sort of standard for naming these hacks and older versions... or even just the TODO comment. Something like `TODO v0.18.4_deprecated`

MV-GH: We could but I don't think we will have many. This is only one right now and I don't foresee any others.

app/src/main/java/com/jerboa/ui/components/common/TimeAgo.kt
MV-GH: Sure, but I usually follow the Rule of three

https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)

dessalines: Haha, never heard of that, but I love it. I always followed the rule of two I spose.

PR #1238, opened 2023-09-15
Revert "Fix zooming difficulties" and temporary remove Swipe to exit in ImageViewer

dessalines:
Reverts dessalines/jerboa#1237


MV-GH:
It should not just revert, those changes, also remove the existing swipe up to exit

dessalines:
There's no rush to do a release anyway, I'll wait on whatever you decide for this.

MV-GH:
Well I rather keep it but both implementations have issues. And I rather do it properly than have a shoddy implemtation

MV-GH (approved):

PR #1242, opened 2023-09-17
 top comment only mode

### LINKED ISSUE #1134: top comment only mode
### Pre-Flight checklist

a selectable mode where only the parent comment shows, no children comments

MakcNmyc:
Closes #1134

### DIFF
--- a/app/src/main/java/com/jerboa/MainActivity.kt
+++ b/app/src/main/java/com/jerboa/MainActivity.kt
@@ -336,6 +336,7 @@ class MainActivity : AppCompatActivity() {
                             onBack = appState::popBackStack,
                             markAsReadOnScroll = appSettings.markAsReadOnScroll,
                             postActionbarMode = appSettings.postActionbarMode,
+                            topCommentOnly =  appSettings.topCommentOnly
                         )
                     }
 
@@ -371,6 +372,7 @@ class MainActivity : AppCompatActivity() {
                             drawerState = drawerState,
                             markAsReadOnScroll = appSettings.markAsReadOnScroll,
                             postActionbarMode = appSettings.postActionbarMode,
+                            topCommentOnly =  appSettings.topCommentOnly
                         )
                     }
 
@@ -478,6 +480,7 @@ class MainActivity : AppCompatActivity() {
                                 blurNSFW = appSettings.blurNSFW,
                                 showPostLinkPreview = appSettings.showPostLinkPreviews,
                                 postActionbarMode = appSettings.postActionbarMode,
+                                topCommentOnly =  appSettings.topCommentOnly
                             )
                         }
                     }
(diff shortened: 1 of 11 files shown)


MV-GH (changes requested):
run `  ./gradlew formatKotlin` to pass the linter, see CONTRIBUTING.md also merge the last changes

MV-GH:
Writing Fixes #issueid makes Github link this PR to that issue

MV-GH:
Also for this feature, I think the creator wanted something more like in Boost. Where you can presss the Floating Action Button to switch to "List view" instead of thread view. So that it only shows the top comments. So this settings does not need in settings saved. Just temporary enabled only for that post. And activated/disabled though the FAB. But for that we need to implement this FAB first. https://github.com/leinardi/FloatingActionButtonSpeedDial

MakcNmyc:
> Also for this feature, I think the creator wanted something more like in Boost. Where you can presss the Floating Action Button to switch to "List view" instead of thread view. So that it only shows the top comments. So this settings does not need in settings saved. Just temporary enabled only for that post. And activated/disabled though the FAB. But for that we need to implement this FAB first. https://github.com/leinardi/FloatingActionButtonSpeedDial

Can I add a FAB from a link?

MV-GH:
Yes you can add the FAB from that repo. Make sure you use the "alpha" m3 version for jetpack compose. You will find it in the last few issues. Make a issue for it here too, so that it is clear what is added. E.g collapsible FAB with, go top, create post, clear read options (clear read you leave off if you don't know how to implement it yet.) And in a post activity it will also need to be added, which expands to jump to top next thread (Like the volume buttons or the bottom parent buttoms do, create post, and this switch for this mode (top comment) between list and thread

MV-GH:
Btw you don't need to write it in the title only in the PR body

dessalines:
I don't even see the use case of that issue. Why is it important to hide children, if the top comment is the first thing you see anyway?

MV-GH:
It's to quickly scroll through all the top comments, I suppose. Either way hiding it in the settings is not a good user experience and will make this feature unusable

MakcNmyc:
> Yes you can add the FAB from that repo. Make sure you use the "alpha" m3 version for jetpack compose. You will find it in the last few issues. Make a issue for it here too, so that it is clear what is added. E.g collapsible FAB with, go top, create post, clear read options (clear read you leave off if you don't know how to implement it yet.) And in a post activity it will also need to be added, which expands to jump to top next thread (Like the volume buttons or the bottom parent buttoms do, create post, and this switch for this mode (top comment) between list and thread

@MV-GH do you mean I need to create new issue, in it I explain why I add new FAB, implement FAB actions (like go top, create post, clear read options, and for that issue top comment mode), add FAB in Activitys (HomeActivity, CommunityActivity, PostActivity)?

MV-GH:
Yes but the issue doesn't need to be in depth. It can similar as above. All  the actions  exist in some form already just need to link em up

MakcNmyc:
@dessalines What about leaving the option in settings and adding "Expand" to FAB when option is active?

MV-GH:
There is no point, Collapse all would cover the top comment only usecase. Both should always display when FAB is pressed in PostActivity

dessalines:
Stale PR.

PR #1243, opened 2023-09-17
Move save to the topbar of user settings and deduplicate topbar code

MV-GH:
- Moved save button to topbar for usersettings
- Add reuseable action topbar
- Remove unused translations (Its a Android Studio tool)

https://github.com/dessalines/jerboa/assets/67873169/df0a74fe-6aaa-4e19-8837-9a5d8a7ded89


app/src/main/java/com/jerboa/ui/components/comment/edit/CommentEdit.kt
```diff
@@ -19,49 +18,6 @@ import com.jerboa.R
 import com.jerboa.db.entity.Account
 import com.jerboa.ui.components.common.MarkdownTextField
 
-@OptIn(ExperimentalMaterial3Api::class)
```
dessalines, line 22: Ah I see, you're abstracting these, nice.

app/src/main/java/com/jerboa/ui/components/common/AppBars.kt
```diff
@@ -642,13 +643,41 @@ fun LoadingBar(
...
+    )
+}
+
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
-fun CreateSubmitHeader(
-    title: String,
-    onClickBack: () -> Unit,
-    onSubmitClick: () -> Unit,
+fun SaveTopBar(
```
dessalines, line 673: I'd change all this `Save` terminology to `Submit` . Because sometime's they're comment edits (where save makes sense), but also comment creates (where save doesn't make sense.)

dessalines (commented):
Nice job! I'd just change save -> submit then we can merge.

app/src/main/java/com/jerboa/ui/components/common/AppBars.kt
MV-GH: Maybe ActionTopBar? I am okay with Submit but then I will have to change the default from save to send

dessalines: That works too. Although you could still have the default as 

`submitText: Int = R.string.save`

app/src/main/java/com/jerboa/ui/components/common/AppBars.kt
```diff
@@ -670,7 +670,7 @@ fun JerboaPullRefreshIndicator(
 
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
-fun SaveTopBar(
+fun ActionTopBar(
     onBackClick: () -> Unit,
     onSaveClick: () -> Unit,
```
dessalines: Change `onSaveClick`, `saveText`, `saveIcon` to

`onSubmitClick`, `submitText`, `submitIcon`, etc. These aren't always saves, but are often creates.

dessalines (approved):
👍

PR #1249, opened 2023-09-24
Fix post delete not working from within PostActivity

### LINKED ISSUE #1248: select 'remove' from post [...] popupmenu does nothing
### Jerboa Version

0.0.46

### Android Version + Phone

Android version 12,  /e/ OS 1.14/

### Describe The Bug

Click 'remove' from the post [...] popup-menu should delete the post, but nothing happens.

### To Reproduce

1 Open the app, (i have Dutch locale enabled)
2 (using an account which has 'bot' property enabled, but also an account without 'bot' enabled)
3 from the main page, click one of your own posts,
4 then from the bottom [...] button, select the last item called 'verwijderen' ('remove' in Dutch language)
5 Then the menu does close, but nothing happens to the post, and no error is displayed.
6 clicking the [...] again, shows the same 'verwijderen' menu item, instead of the 'herstellen'(restore) option.

it doesn't crash

MV-GH:
Also moved its location a bit, UI wise looks a bit better now

Fixes #1248 

https://github.com/dessalines/jerboa/assets/67873169/8d36bdcd-b676-475c-bd9f-554931dd43dd

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/post/PostActivity.kt
+++ b/app/src/main/java/com/jerboa/ui/components/post/PostActivity.kt
@@ -399,7 +399,7 @@ fun PostActivity(
                                             postViewModel.deletePost(
                                                 DeletePost(
                                                     post_id = pv.post.id,
-                                                    deleted = pv.post.deleted,
+                                                    deleted = !pv.post.deleted,
                                                     auth = it.jwt,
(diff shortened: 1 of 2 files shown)


dessalines (approved):
Nice catch, thx!

PR #1255, opened 2023-10-07
Community: Add Share drop down in community view.

lubosz:
I found that there is no button to share a community url. This patch fixes that.
Not sure if this is the best way to create a community URL, let me know if I am missing something.
Also not sure if https can be expected from all instances. Didn't find any places were instance / community URLs are already created.

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/community/CommunityActivity.kt
+++ b/app/src/main/java/com/jerboa/ui/components/community/CommunityActivity.kt
@@ -49,6 +49,7 @@ import com.jerboa.datatypes.types.SubscribedType
 import com.jerboa.db.entity.getJWT
 import com.jerboa.db.entity.isAnon
 import com.jerboa.feat.doIfReadyElseDisplayInfo
+import com.jerboa.feat.shareLink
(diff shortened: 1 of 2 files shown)


MV-GH (approved):
There is a example in ParseUrl. It does look like we assume https from all instances. I don't really see away to check it anyway. And Android forbids most http traffic anyway due to ClearText.

app/src/main/java/com/jerboa/ui/components/community/CommunityActivity.kt
```diff
@@ -200,6 +201,10 @@ fun CommunityActivity(
                                 }
                             },
                             onClickCommunityInfo = { appState.toCommunitySideBar(communityRes.data.community_view) },
+                            onClickCommunityShare = {
+                                val name = communityRes.data.community_view.community.name
+                                shareLink("https://$instance/c/$name", ctx)
```
dessalines, line 206: Use `community_view.community.actor_id`, its the official URL.

dessalines (approved):
Thx!

PR #1260, opened 2023-10-12
Fix search list repopulating on return

MV-GH:
This was a regression introduced in #1208. On return it would repopulate the searchlist with the user their followed communities. Now it only does it on navigation to the search.

https://github.com/dessalines/jerboa/assets/67873169/e1d1508e-c179-4515-8bbf-31a7c1f5a546


dessalines (approved):
Good catch, thx!

PR #1267, opened 2023-10-18
Adding a theming guide.

dessalines:


MV-GH (approved):
It's a bit more work due to those custom colors. But nothing we can't easily refactor if its too complex for someone

PR #1277, opened 2023-11-15
v0.19 upgrade

dessalines:
@MV-GH this is ready for you to look at.

To summarize my changes:

- Upgraded to the new auth method.
- Upgraded all the structs, and fixed some issues with the `copy_generated_types` script.
- Added the new pagination_cursor method.
- Ran a code analysis and fixed a lot of issues there.
- Removed all the version-checking sorting enums. 

I tested most things, and they seem to be working correctly.

Limitations:

- No backwards-compatibility. 
  - I didn't change the MINIMUM_VERSION in the code, because viewing the site does work.
- The `validate_auth` sections of the code: I'm not entirely sure how they should work, but that new endpoint is deployed on voyager.lemmy.ml .

Next week we'll be testing some `0.19` on lemmy.ml to catch any last bugs, which will unfortunately probably break apps for a few days, before we do the `0.19.0` lemmy backend release.

I defer you to you on the best way to handle the upgrade to your new backwards-compatibility library / method. Even if we don't get it merged in time for this release, it'll still be great to have for future ones. I hope we don't get trapped in merge hell with our two PRs, but its not the end of the world if we do, we'll work through it. LMK anything I can do to help there.

cc @nutomic


MV-GH:
My Api is getting pretty far too. Doing initial integration attempts. Not sure what the timeline is for v0.19. I probably could get it done in 2weeks. Really depends in what problems in run into.

btw this has a good guide for publishing to maven

https://central.sonatype.org/publish/publish-maven/

dessalines:
Cool. Its tough to say yet when we'll release `v0.19`, definitely not this week, but possibly next week. Obvi I'll make sure jerboa is ready before then. But for now at least I'm just planning on supporting v0.19, considering all the API changes.

If you get your backwards-compatible API done soonish tho, we can use that one instead / in favor.

app/src/main/java/com/jerboa/model/LoginViewModel.kt
```diff
@@ -59,9 +57,13 @@ class LoginViewModel : ViewModel() {
                 Toast.makeText(ctx, msg, Toast.LENGTH_LONG).show()
                 return@launch
             }
+            // Login was sucessful after this point
+
+            // Change the lemmy instance to the new one
+            API.changeLemmyInstance(instance, jwt)
```
MV-GH, line 63: This has some other consequences, there was a reason I put all the way at the end. If the site call fails it will still not allow you to login and force you to retry which is good. For example trying to login with 0.19 now will fail at getSite call but it wont cause you to be logged in. Which would be the case now. The check version check would be even more useless as it would set the client regardless

app/src/main/java/com/jerboa/model/SiteViewModel.kt
```diff
@@ -54,30 +49,26 @@ class SiteViewModel(private val accountRepository: AccountRepository) : ViewMode
                     Log.d("Jerboa", "acc init for id: ${it.id}")
 
                     if (it.isAnon()) {
```
MV-GH, line 51: Don't forget that you need to set a new auth when you change ur account/log out. Look in AccountViewModel (take over the design, I did with my pr) Encapsulate it so we dont have to worry with removeAccount and etc

MV-GH (changes requested):
I was originally gonna just take over the changes from this PR into mine. But that no longer seems manageable, rebase it will be then not easy for sure. 

I don't agree with some of the analysis changes

For the verification changes, the validateAuth, , success -> good, 4XX -> bad, 500 -> retry. So if validateAuth has the same behaviour is good. I have emulate validateAuth with that getMentionsCall for my API (for 0.18). 


Also might have missed it but there should also be changes to the settings, the removal of that one. The postnotif setting do we completely remove it our just the UI parts. (Idc tbh)

MV-GH:
Also you will have to redo the generate Types thing, there were some changes again

app/src/main/java/com/jerboa/api/Http.kt
```diff
@@ -316,6 +320,7 @@ interface API {
     @POST
     suspend fun uploadImage(
         @Url url: String,
+        // TODO get rid of this cookie
```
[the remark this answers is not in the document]
dessalines: I'll get rid of this altogether, since auth is already passed through that interceptor, and the lemmy back-end accepts it now.

app/src/main/java/com/jerboa/model/LoginViewModel.kt
dessalines: Not sure I follow this one. Right now it is:

- Create a temp API for logging in.
- If the login succeeds, change the API in order to set the JWT.
- Fetch the site to get `MyUserInfo`
- Save it to the DB
- Go home

What am I missing?

app/src/main/java/com/jerboa/model/SiteViewModel.kt
dessalines: The `changeLemmyInstance` on L54 sets the auth.

app/src/main/java/com/jerboa/JerboaAppState.kt
```diff
@@ -185,7 +180,7 @@ class JerboaAppState(
      *
      * When you want to pass a [Parcelable] to another screen/activity
      */
-    fun sendReturnForwards(
+    private fun sendReturnForwards(
```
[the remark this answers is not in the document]
dessalines: Fixed.

app/src/main/java/com/jerboa/datatypes/Others.kt
```diff
@@ -10,18 +10,14 @@ import androidx.compose.material.icons.outlined.History
...
 import androidx.compose.material.icons.outlined.NewReleases
+import androidx.compose.material.icons.outlined.Scale
 import androidx.compose.material.icons.outlined.ThumbsUpDown
 import androidx.compose.runtime.Immutable
 import androidx.compose.ui.graphics.vector.ImageVector
 import com.google.gson.annotations.SerializedName
 import com.jerboa.R
-import com.jerboa.api.MINIMUM_API_VERSION
-import com.jerboa.compareVersions
 import kotlinx.parcelize.Parcelize
```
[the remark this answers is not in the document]
dessalines: Thx, got it.

app/src/main/java/com/jerboa/model/LoginViewModel.kt
MV-GH: Fetch Site can fail for multiple reasons.

Currently it fails for incompatible versions.  login succeeds for 0.19 but getSite doesn't in this case you do not want to confirm the api. Currently (with your code) it will display error, login failed but it actually successfully changed the api but it didn't make account. Meaning when ppl back out they are using the new api(for the new account) instead of the actual account that is logged in.

dessalines:
K, addressed some of these above. Re-running `copy_generated_types` didn't seem to pick up anything new.

app/src/main/java/com/jerboa/model/SiteViewModel.kt
MV-GH: Ah yes this is a flow updates the api. mb

app/src/main/java/com/jerboa/model/LoginViewModel.kt
dessalines: Hrm. Shouldn't a failed login be caught in the code just above this?

dessalines: Maybe I should be doing the checks on `tempApi.getSite`, and if those pass, then change the instance and continue with the DB stuff?

MV-GH: > Or maybe, login is succeeding, but GetSite fails with older versions?

Yes, not just with older versions also newer versions.

```kt
       API.changeLemmyInstance(instance, jwt)

            // Fetch the site to get your name and id
            siteViewModel.siteRes = apiWrapper(API.getInstance().getSite())

            try {
                when (val siteRes = siteViewModel.siteRes) {
                    is ApiState.Failure -> {
                        val txt = siteRes.msg.message ?: "FAILURE: NO MESSAGE, probably that version not supported"
                        Toast.makeText(ctx, txt, Toast.LENGTH_SHORT).show()
                        throw RuntimeException(txt)
                    }
```

Here we check if the getSite call failed. and show a error msg


The flow when logging with older version:
- Create a temp API for logging in.
- If the login succeeds, change the API in order to set the JWT.
- Confirms the API instance to the new instance
- Fetch the site to get MyUserInfo
- Site Call fails
- Throw error, not supported version, show this
- User backs out the login
- Screen throws errors everywhere, cuz its doing calls with the newly attempted user to login


Before
- Create a temp API for logging in.
- If the login succeeds, change the API in order to set the JWT.
- Fetch the site to get MyUserInfo
- Site Call fails
- Throw error, not supported version, show this
- User backs out the login
- Everything still works because it keeps using the old instance

So just keep the old logic and confirm the api as last as before

MV-GH: > Maybe I should be doing the checks on tempApi.getSite, and if those pass, then change the instance and continue with the DB stuff?

We already do this

app/src/main/java/com/jerboa/model/CommunityViewModel.kt
```diff
@@ -61,26 +59,20 @@ class CommunityViewModel(account: Account, communityArg: Either<CommunityId, Str
...
-    }
-
-    fun nextPage() {
-        page += 1
-    }
-
-    fun prevPage() {
-        page -= 1
+    private fun resetPage() {
+        pageCursor = null
```
MV-GH, line 72: How does this resetPage logic work? in Appending it resets the value regardless the last pageCursor. Does this reset and cursor state even have purpose? Would the reset not be better be to remove the oldRes itself?

app/src/main/java/com/jerboa/model/LoginViewModel.kt
dessalines: Here's the issue on L63-64 : 

```kt
// Fetch the site to get your name and id
siteViewModel.siteRes = apiWrapper(api.getSite(GetSite(auth)....
```

Its using the tempApi, with a custom JWT, to fetch `MyUserInfo` . That's no longer possible.

I'll see if I can refactor it in a way that works, where it only changes it at the end.

app/src/main/java/com/jerboa/model/CommunityViewModel.kt
dessalines: If pageCursor is null / missing, then it just fetches the first page. If you call `getPosts(page_cursor = ...`, then it fetches the next page of results.

If you trace down resetPage(), you'll see its called for things like changing the sort type, clicking refresh, etc.

app/src/main/java/com/jerboa/model/LoginViewModel.kt
dessalines: Okay I think I got this one now, check the most recent commit or a diff of this file from main.

dessalines:
BTW I also tested image uploading on voyager.lemmy.ml from this PR (I removed that cookie stuff), and it works.

app/src/main/java/com/jerboa/model/LoginViewModel.kt
```diff
@@ -59,11 +59,11 @@ class LoginViewModel : ViewModel() {
             }
             // Login was sucessful after this point
```
MV-GH, line 60: small typo

MV-GH:
I was trying to test this PR but voyager seems to be having issues atm. I will test it later

MV-GH:
Alright it works but it seems constantly display, this account is being verified on return out of a post. (It should not even be verifying the account)

Using 0.18 accs seems to work better than I expected, I can browse but it does display the verification thingy stuck on jwt (prob cuz that doesn't fall back thus jwt verification fails as auth=null) (Maybe a warning? we can keep as is but we will definitely get issues around this)

MV-GH (approved):
I can't reproduce the previous "The account is being verified" issue anymore but there is definitely an edgecase somewhere where it doesn't properly purge the lock

dessalines:
> Using 0.18 accs seems to work better than I expected, I can browse but it does display the verification thingy stuck on jwt (prob cuz that doesn't fall back thus jwt verification fails as auth=null) (Maybe add a version warning? we can keep as is but we will definitely get issues around this)

Where would be the best place to add that toast?

MV-GH:
We had one in root our app by twiz, but that one is based of getSite which wouldn't work as getSite doesn't work across version. You could reuse it but add that nodeinfo stuff to get version but that's a lotta work. Or you could wait until my api is done.(merge as is, release in the meantime) Or we can use my LemmyApi.getVersion that does functionality but we would have cut a prelease first.

dessalines:
I'll wait until your API is done, I'll just merge this as is. Again, sry about the conflicts, lmk if there's anything I can do to help there.

dessalines:
(3) Makes the most sense to me also. Could be a part of the `changeInstance` function, which I believe is called on startup, as well as when switching accounts.

PR #1284, opened 2023-11-25
Initial MVP integration with my BackwardsCompatibleAPI

MV-GH:
Basic functionality seems to work

Still loads of TODOs I have to work on.

This PR is mostly to gauge opinions on some of the structural changes

Auth stays with the API now. No longer need to drag a Account everywhere where you need an API call.
Only remnants of account now are to check if the user is logged to logged in required views. I would like to change this too maybe the navigation to it should do the check somewhere and we then simply assume they are logged in?

Can't keep parcelables, Parcelables is Android Only and My API is written in pure Kotlin so couldn't even target io.Serializable. So back to string.

Removes datatypes from this repo

Removes Enums too, went with "data"  approach

No need for retrofitErrorHandler as the API will already retrieve the error from the response and map to a LemmyBadRequestException with it as msg

https://github.com/dessalines/jerboa/assets/67873169/6ca39f82-7ac6-4ee7-86b1-44ed5929132d

Can't run this yourself atm as I haven't published a version yet.


dessalines:
We are gonna have some wild merge conflicts with #1277 , so before going further I'd rebase this PR on that one, which is mostly done. Obvi in the future I'd prefer we use your library and remove the datatypes folder altogether.

MV-GH:
There is still work todo, but it is in a usable state.

I did some research into some resource leaks we were experiencing. I fixed one our side and discovered one in a third party lib.

using
```
        StrictMode.setVmPolicy(
            StrictMode.VmPolicy.Builder()
                .detectActivityLeaks()
                .detectLeakedClosableObjects()
                .detectContentUriWithoutPermission()
                .detectCleartextNetwork()
                .detectFileUriExposure()
                .detectLeakedRegistrationObjects()
                .detectLeakedSqlLiteObjects()
                .penaltyLog()
                .build())
        StrictMode.setThreadPolicy(
            StrictMode.ThreadPolicy.Builder()
                .detectAll()
                .permitDiskReads()
                .penaltyLog()
                .build())
` 

At the application start 

At some point I will add this always on for debug mode

MV-GH:
Also I can't build the release if you limit gradle deamon to 1GB. Do you experience this?

dessalines:
It passed luckily. Seems to be just CI weirdness again. I'll test this out now.

dessalines (approved):
I tested and it works well on multiple instances, nice job!

I noticed you added some TODOs, is that something you want to handle before a merge, or just get this out and handle them later?

MV-GH:
After, I can turn them into issues if you prefer that

dessalines:
Up to you. Can I start another release with this now?

MV-GH:
Yeah that's fine

MV-GH:
> Also I can't build the release if you limit gradle deamon to 1GB. 
With this i meant locally, debug builds fine but release keeps hitting OOM exceptions.

dessalines:
I'll test that rn.

PR #1288, opened 2023-12-12
Fix crash on opening profile anonymous

MV-GH:
The crash is because the verification tried to make API instance for the Anon account. Which causes a crash as it gives a empty string and thus then tries to do a version call to to localhost. 

I will comeback to this later. This decent fix, but there is another edge case to consider.

Technically its possible that the account will reference a url to a instance that isn't valid anymore. (Shutdown instance). And then we will have the same problem.


dessalines (approved):

dessalines:
I'll open up an issue for the latter case.

MV-GH:
If you want can make another release after these two fixes.

dessalines:
K, will do.

PR #1300, opened 2023-12-17
Improve Russian translations

mittwerk:


MV-GH (approved):
Thanks!

Those entries that you marked non translatable if they had some translations could you remove them?

mittwerk:
Well, you can take them back, but I don't think it's necessary to translate it into any language.

MV-GH:
I agree that's why I asked if they have already translations that you would remove them.

dessalines (changes requested):
What's going on with all these formatting and dependency changes? Just update the translations only.

MV-GH:
Make a separate PR for those build changes

dessalines (approved):
Sweet, thx

PR #1309, opened 2023-12-22
Woodpecker CI cache fixes

Nutomic:
Same as https://github.com/LemmyNet/lemmy/pull/4276:

- Fail CI run on cache error so that we can actually notice and fix problems
- Increase timeout for cache operations (default is only 3m)
- Disable cache compression (takes unnecessary time)

Also change cache rebuild to run only on main branch, to prevent random attackers from writing into the cache from a PR.

### DIFF
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -76,8 +76,8 @@ steps:
     secrets:
       [MINIO_ENDPOINT, MINIO_WRITE_USER, MINIO_WRITE_PASSWORD, MINIO_BUCKET]
     when:
-      - event: push
-        branch: main
+      - path:
+          include: ["app/build.gradle.kts"]
 
   notify:
     image: alpine:3
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -75,7 +75,7 @@ steps:
         - ".gradle"
     secrets:
       [MINIO_ENDPOINT, MINIO_WRITE_USER, MINIO_WRITE_PASSWORD, MINIO_BUCKET]
-    #when:
+    ##when:
     #  event: cron
 
   notify:
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -22,7 +22,7 @@ steps:
       path-style: true
       backend_operation_timeout: 30m
       compression_level: 0
-      exit_code: true
+      #exit_code: true
       mount:
         - ".gradle"
     secrets:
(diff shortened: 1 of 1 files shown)


.woodpecker.yml
```diff
@@ -65,13 +68,16 @@ steps:
...
+      backend_operation_timeout: 60m
+      compression_level: 0
+      exit_code: true
       mount:
         - ".gradle"
     secrets:
       [MINIO_ENDPOINT, MINIO_WRITE_USER, MINIO_WRITE_PASSWORD, MINIO_BUCKET]
     when:
-      - path:
-          include: ["app/build.gradle.kts"]
```
MV-GH, line 74: Could this be kept? I had added this, because I only want to save the new caches if something had changed.

Nutomic: The problem is that anyone can change the file in a PR, and then potentially get access to secrets to mess with the cache. Maybe a better option would be a cronjob which automatically rebuilds the cache every night or so.

MV-GH: Actually this wouldn't prevent that at all. Anyone can edit this workflow, set whatever event they want and trigger a pr that updates the caches

MV-GH: I don't know how Dessalines Setup the secrets but you can set it up that only the main branch can access the secrets. That is probably a better solution see [this](https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#creating-secrets-for-an-environment)

MV-GH: In Github you can probably do a "hack" where you can set a secret only available (in Environment) if the file hasn't changed and another environment that allows only access to the same secret if approved by a required reviewer.

In Woodpecker it doesn't seem possible to do something similar. https://woodpecker-ci.org/docs/usage/secrets#use-in-pull-requests-events

Nutomic: Thats true. Ive changed it back now.

MV-GH: If this is deemed too much of a security hazard, the caching can be removed, we don't gain that much from it. The biggest improvement was already made by setting the .gradle as ENV. This allowed it to reuse the gradle cache between steps

Nutomic: Its mostly a theoretical problem, at least so far it hasnt been abused in practice. And we would probably notice if someone opened a malicious PR.

MV-GH: Seems it failed to download a depency, network issues prob, try again by pushing a commit

dessalines (approved):

PR #1315, opened 2023-12-25
Expand search functionality

MV-GH:
Did some refactors too. Found repeated actions around posts, comment, created these "helper" constructs to reduce the code duplication.

current progress:

https://github.com/dessalines/jerboa/assets/67873169/849b1712-6e18-47ae-87c4-33e7e570b129

I kept the community select mode, which defaults to community sorts and selections by default. I will expand it to add user, and then we can the search to select users too.


MV-GH:
Changes have been extracted

Issue #1182, opened 2023-08-15
List of moderator actions / moderation abilities


MV-GH:
A top-level issue for moderation abilities:

- [x] locking posts
- [x] feature / Unfeature post in community, local.
- [x] removing / restoring posts
- [x] removing / restoring comments
- [x] banning / suspending users
- [x] registration application queue
- [x] report queue
- [x] Admin view votes
- [x] mod priority - e.g. higher mods can kick lower mods
- [ ] locking entire comment chains (select top level and all sublevels are locked)
- [ ] deleting entire comment chains

dessalines:
These are all done now, except for those last two, but they should be added as back-end features first.

MV-GH:
They can be achieved already on the front end though?

dessalines:
Locking specific comments doesn't exist in the back end, so it'd need an issue.

Deleting comment chains seem *could* be done via a front end loop, but I'd prefer that also be a back end ability where you only need to provide a single parent comment.

Both of these need back end issues, and probably separate ones here too.

PR #1323, opened 2024-01-09
Adding ability to remove and restore comments. #1182

dessalines:
This adds the ability to remove and restore comments. 

It also adds the correct hierarchical logic for the `canMod` function copied from lemmy-ui, which requires a list of admins and mods.


app/src/main/java/com/jerboa/Utils.kt
```diff
@@ -1526,3 +1536,35 @@ fun String.padUrlWithHttps(): String {
         "https://$this"
     }
 }
+
+/**
+ * Determines whether someone can moderate an item. Uses a hierarchy of admins then mods.
+ */
+fun canMod(
```
dessalines, line 1543: This was copied from lemmy-ui.

app/src/main/java/com/jerboa/model/RemoveViewModel.kt
```diff
@@ -0,0 +1,68 @@
...
+import com.jerboa.api.API
+import com.jerboa.api.ApiState
+import com.jerboa.api.toApiState
+import com.jerboa.ui.components.common.apiErrorToast
+import it.vercruysse.lemmyapi.v0x19.datatypes.CommentResponse
+import it.vercruysse.lemmyapi.v0x19.datatypes.CommentView
+import it.vercruysse.lemmyapi.v0x19.datatypes.RemoveComment
+import kotlinx.coroutines.launch
+
+class RemoveViewModel : ViewModel() {
```
dessalines, line 22: Debating whether to have this used for both Posts and Comments, or just Comments.

app/src/main/java/com/jerboa/ui/components/comment/CommentNode.kt
```diff
@@ -171,6 +172,8 @@ fun CommentBodyPreview() {
 
 fun LazyListScope.commentNodeItem(
     node: CommentNode,
+    admins: List<PersonView>,
+    moderators: List<CommunityModeratorView>?,
```
dessalines, line 176: I don't like the fact that we have to inject these lists for every single comment, and I'm sure its a performance hit. But this canMod logic unfortunately lives in the front end, and requires both.

app/src/main/java/com/jerboa/ui/components/comment/CommentNode.kt
```diff
@@ -678,10 +697,18 @@ fun CommentFooterLine(
             onEditCommentClick = onEditCommentClick,
             onDeleteCommentClick = onDeleteCommentClick,
             onReportClick = onReportClick,
+            onRemoveClick = onRemoveClick,
             onBlockCreatorClick = onBlockCreatorClick,
             onCommentLinkClick = onCommentLinkClick,
             onPersonClick = onPersonClick,
             isCreator = account.id == commentView.creator.id,
+            canMod =
```
dessalines, line 705: I'm not sure at what level it makes sense to run this function. Its really only used in the comment dropdown.

app/src/main/java/com/jerboa/ui/components/comment/CommentOptionsDropdown.kt
```diff
@@ -160,6 +171,25 @@ fun CommentOptionsDropdown(
                     onReportClick(commentView)
                 },
             )
+
+            if (canMod) {
+                Divider()
```
dessalines, line 176: Having the moderation-related abilities in another divider.

app/src/main/java/com/jerboa/ui/components/person/PersonProfileActivity.kt
```diff
@@ -644,6 +648,9 @@ fun UserTabs(
                                 }
                                 CommentNodes(
                                     nodes = nodes,
+                                    admins = siteViewModel.admins(),
+                                    // No community moderators available here
+                                    moderators = null,
```
dessalines, line 653: Unfortunate, but lemmy-ui has the same issue.

app/src/main/java/com/jerboa/ui/components/remove/comment/CommentRemoveActivity.kt
```diff
@@ -0,0 +1,95 @@
...
+import com.jerboa.model.AccountViewModel
+import com.jerboa.model.RemoveViewModel
+import com.jerboa.ui.components.common.ActionTopBar
+import com.jerboa.ui.components.common.getCurrentAccount
+import com.jerboa.ui.components.remove.RemoveItemBody
+import it.vercruysse.lemmyapi.v0x19.datatypes.Comment
+
+object CommentRemoveReturn {
+    const val COMMENT_VIEW = "comment-remove::return(comment-view)"
+    const val COMMENT_SEND = "comment-remove::send(comment-view)"
```
dessalines, line 30: These seem redundant, since they serve the same purpose as `CommentEditReturn.COMMENT_XXX` anyway. I spose it makes sense to have them be explicit tho.

app/src/main/java/com/jerboa/ui/components/remove/comment/CommentRemoveActivity.kt
```diff
@@ -0,0 +1,95 @@
...
+    appState: JerboaAppState,
+    accountViewModel: AccountViewModel,
+) {
+    Log.d("jerboa", "got to create comment remove activity")
+
+    val ctx = LocalContext.current
+    val account = getCurrentAccount(accountViewModel = accountViewModel)
+
+    val removeViewModel: RemoveViewModel = viewModel()
+    val comment = appState.getPrevReturn<Comment>(key = CommentRemoveReturn.COMMENT_SEND)
```
dessalines, line 44: These appState returns... I wish there was a better way to do them, because the types seem forced. Its basically a duck-typed key-value store.

dessalines (commented):
I plan on adding more mod-abilities to jerboa in the coming weeks, but this is a good start.

app/src/main/java/com/jerboa/model/RemoveViewModel.kt
MV-GH: If you want it to use both you will probably have to make it into a "helper construct"

app/src/main/java/com/jerboa/ui/components/comment/CommentNode.kt
MV-GH: These need to be `ImmutableList`, all of them. I saw that you used the correct immutablelist at the source but not in the parameters.

*Difference btw ImmutableList vs List, only one of them is the compose compiler sure that it will be immutable*
image


The performance shouldn't be impacted much. I see that it only runs when opening the options menu.

MV-GH: We can avoid these prop drilling. But you won't like the solution either. We can do DI and inject them directly into MoreOptions menu

app/src/main/java/com/jerboa/ui/components/comment/CommentNode.kt
MV-GH: This is alright, just make sure you surround it with a `remember`

app/src/main/java/com/jerboa/ui/components/comment/CommentOptionsDropdown.kt
MV-GH: Yeah that's alright, if the popup becomes to big bc of the mod options then we'll probably have to divide these into sub menus.

app/src/main/java/com/jerboa/ui/components/remove/comment/CommentRemoveActivity.kt
MV-GH: Yeah it aint the best, rather had compile time safety but its the best we can do rn.

MV-GH (changes requested):
I can't test the functionality, not a mod somewhere but I'll take your word on it.

app/src/main/java/com/jerboa/model/RemoveViewModel.kt
dessalines: I'll pry use two different viewModels then.

dessalines:
I've been using voyager.lemmy.ml to test. You can use my test admin account there:

username: `dess_tester_3`

pass: `dess_tester_3dess_tester_3`

app/src/main/java/com/jerboa/ui/components/comment/CommentNode.kt
MV-GH: I can technically simulate this already in the LemmyAPI. If you prefer this?

MV-GH (approved):

PR #1324, opened 2024-01-11
Post removing

dessalines:
Fully working, but should be merged after #1323


app/src/main/res/values/strings.xml
```diff
@@ -218,6 +218,7 @@
     <string name="person_profile_report_person">Report Person</string>
     <string name="pictrs_image_upload_image">Upload Image</string>
     <string name="postListing_deleted">Deleted</string>
+    <string name="postListing_removed">Removed</string>
```
dessalines, line 221: I *really* want to do away with all these "contextual" translations, but it'd potentially be a lot of new work for translators. `Removed` is the same everywhere.

MV-GH: And why can the existing `removed` not be used? 

You can remove existing contextual removed and rename to be generic

dessalines: Thx, will do.

app/src/main/java/com/jerboa/Utils.kt
```diff
@@ -1455,7 +1455,7 @@ fun Context.getInputStream(url: String): InputStream {
             Request.Builder()
                 .url(url)
                 .build(),
-        ).execute().body?.byteStream() ?: throw IOException("Failed to get input stream")
+        ).execute().body.byteStream()
```
dessalines, line 1458: I added this in the last commit, because the compiler warned that the previous checks were pointless. But I can revert if there's something else going on.

app/src/main/java/com/jerboa/ui/components/comment/mentionnode/CommentMentionNode.kt
```diff
@@ -112,6 +118,16 @@ fun CommentMentionNodeFooterLine(
 ) {
     var showMoreOptions by remember { mutableStateOf(false) }
 
+    val canMod =
```
dessalines, line 121: I'd forgotten to add comment removals to comment mention nodes.

app/src/main/res/values/strings.xml
MV-GH: just keep the one for post/comment removed deleted separate as that one can be different across languages

app/src/main/java/com/jerboa/Utils.kt
MV-GH: it was required when I remove the dependency for the custom log interceptor. But I kept it for now, so it isn't needed yet.  But I plan on removing that stuff since auth now is kept in the header. And I plan on moving rather to the LemmyAPI.

app/src/main/java/com/jerboa/ui/components/comment/mentionnode/CommentMentionNode.kt
MV-GH: I am in the progress of merging all those in the existing commentNodes,

app/src/main/java/com/jerboa/ui/components/community/CommunityActivity.kt
```diff
@@ -106,6 +108,14 @@ fun CommunityActivity(
             },
         )
 
+    val moderators =
```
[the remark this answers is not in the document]
MV-GH: I don't follow, but it might be worth it to encase these too in remembers just make the communityRes its key

MV-GH (approved):

app/src/main/java/com/jerboa/ui/components/community/CommunityActivity.kt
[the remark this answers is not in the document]
dessalines: Gotcha, can do.

PR #1325, opened 2024-01-20
Add ability to Ban and Ban from Community for posts and comments.

dessalines:
- Adding appropriate viewModels and activities.
- Adds functions to search for and update lists of posts and comments
- Context: https://github.com/dessalines/jerboa/issues/1182

Edit: Should come after #1324


app/src/main/java/com/jerboa/datatypes/Others.kt
```diff
@@ -187,3 +190,13 @@ fun getLocalizedStringForInboxTab(
         }
     return returnString
 }
+
+/**
+ * A wrapper to store extra community ban info
+ */
+@Serializable
+data class BanFromCommunityData(
```
dessalines, line 198: This is necessary to bundle some info for community bans. site bans don't need this cause `PersonView` has all the data you need.

app/src/main/java/com/jerboa/Utils.kt
```diff
@@ -281,7 +284,8 @@ fun commentsToFlatNodes(comments: List<CommentView>): ImmutableList<CommentNode>
  */
 fun buildCommentsTree(
     comments: List<CommentView>,
-    rootCommentId: Int?, // If it's in CommentView, then we need to know the root comment id
+    // If it's in CommentView, then we need to know the root comment id
```
dessalines, line 287: kotlinter complains about all same-line comments now, so I fixed these up.

app/src/main/java/com/jerboa/Utils.kt
```diff
@@ -1204,6 +1209,66 @@ fun findAndUpdateComment(
     }
 }
 
+fun findAndUpdatePostCreator(
+    posts: List<PostView>,
+    person: Person,
+): List<PostView> {
+    val newPosts = posts.toMutableList()
+    newPosts.replaceAll {
```
dessalines, line 1217: I couldn't find a better way to alter several items for an immutable list. 

I would've liked to use `.filter { }.forEach { }`, but it didn't work since the internal items are immutable.

app/src/main/java/com/jerboa/Utils.kt
```diff
@@ -1568,3 +1633,9 @@ fun canMod(
         false
     }
 }
+
+fun futureDaysToUnixTime(days: Int?): Int? {
```
dessalines, line 1637: Copied from lemmy-ui. The back end requires a unix epoch, but the front end should just use days in its fields.

app/src/main/java/com/jerboa/model/PostViewModel.kt
```diff
@@ -267,6 +268,62 @@ class PostViewModel(val id: Either<PostId, CommentId>) : ViewModel() {
         }
     }
 
+    fun updateBanned(personView: PersonView) {
```
dessalines, line 271: Wasn't sure what to name this one. Its only used for banning a user currently, but could potentially be used for any updates for every person in a post.

app/src/main/java/com/jerboa/ui/components/ban/BanFromCommunityActivity.kt
```diff
@@ -0,0 +1,132 @@
...
+import com.jerboa.ui.components.common.ActionTopBar
+import com.jerboa.ui.components.common.getCurrentAccount
+
+object BanFromCommunityReturn {
+    const val BAN_DATA_VIEW = "ban-from-community::return(ban-data-view)"
+    const val BAN_DATA_SEND = "ban-from-community::send(ban-data-view)"
+}
+
+@Composable
+fun BanFromCommunityActivity(
```
dessalines, line 35: Very similar logic, so unfortunately a lot of copy pasting. But its better to have these activities separate and explicit IMO.

app/src/main/java/com/jerboa/ui/components/ban/BanPerson.kt
```diff
@@ -0,0 +1,79 @@
...
+import androidx.compose.ui.text.input.TextFieldValue
+import com.jerboa.R
+import com.jerboa.db.entity.Account
+import com.jerboa.ui.components.common.CheckboxField
+import com.jerboa.ui.components.common.ExpiresField
+import com.jerboa.ui.components.common.MarkdownTextField
+import com.jerboa.ui.theme.MEDIUM_PADDING
+
+@Composable
+fun BanPersonBody(
```
dessalines, line 22: At least this composable is identical for both bans: local and community.

app/src/main/java/com/jerboa/ui/components/ban/BanPerson.kt
```diff
@@ -0,0 +1,79 @@
...
+            onTextChange = onReasonChange,
+            account = account,
+            modifier = Modifier.fillMaxWidth(),
+            placeholder = stringResource(R.string.type_your_reason),
+        )
+
+        // Only show these fields for a ban, not an unban
+        if (isBan) {
+            if (!permaBan) {
+                ExpiresField(
```
dessalines, line 59: I tried to make this form identical to lemmy-ui: it hides the expires field if you've clicked permaban.

app/src/main/java/com/jerboa/ui/components/comment/CommentNode.kt
```diff
@@ -692,7 +705,7 @@ fun CommentFooterLine(
     var showMoreOptions by remember { mutableStateOf(false) }
 
     val canMod =
-        remember {
+        remember(admins) {
```
dessalines, line 708: Need your help with this one. 

This was an error with the last PR. On first app load, it would incorrectly not should the extra mod actions, because admins hadn't loaded yet. 

This fixes it, but it might cause a lot of recompositions when admins show up? I'm having trouble deciphering when to use derivedState vs remember(key). 

Also, should this remember also include account, IE `remember(account, admins) { ...`

app/src/main/java/com/jerboa/ui/components/common/InputFields.kt
```diff
@@ -641,6 +644,50 @@ fun MyMarkdownText(
...
+@Composable
+fun ExpiresField(
+    value: Int?,
+    onIntChange: (Int?) -> Unit,
+    isValid: Boolean,
+) {
+    OutlinedTextField(
+        modifier = Modifier.fillMaxWidth(),
+        value = value?.toString() ?: "",
+        onValueChange = {
```
dessalines, line 677: Some weird logic here, but it seems to work. Mostly copied from stackoverflow. The expire days needs positive integers, or if its empty, send null.

app/src/main/java/com/jerboa/ui/components/common/PopupItems.kt
```diff
@@ -0,0 +1,61 @@
...
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.res.stringResource
+import com.jerboa.R
+import com.jerboa.communityNameShown
+import com.jerboa.datatypes.BanFromCommunityData
+import com.jerboa.personNameShown
+import it.vercruysse.lemmyapi.v0x19.datatypes.Person
+
+@Composable
+fun BanPersonPopupMenuItem(
```
dessalines, line 15: Externalized these to a common composable, since both post and comment items use them. Should maybe do the same thing for post / comment `Removes`

app/src/main/java/com/jerboa/ui/components/post/composables/PostComposables.kt
```diff
@@ -228,26 +224,6 @@ fun PostCommunitySelector(
     }
 }
 
-@Composable
-fun CheckboxIsNsfw(
```
dessalines, line 232: I externalized this to a common `CheckBoxField` composable.

app/src/main/java/com/jerboa/ui/theme/Color.kt
```diff
@@ -1236,7 +1236,6 @@ fun dracula(): Pair<JerboaColorScheme, JerboaColorScheme> {
     val md_theme_light_inverseOnSurface = Color(0xFFF3F0F4)
     val md_theme_light_inverseSurface = Color(0xFF303034)
     val md_theme_light_inversePrimary = Color(0xFFB9C3FF)
-    val md_theme_light_shadow = Color(0xFF000000)
```
dessalines, line 1239: Compiler complaints.

app/src/main/java/com/jerboa/ui/components/comment/CommentNode.kt
MV-GH: >  I'm having trouble deciphering when to use derivedState vs remember(key).

derivedState is to prevent recompositions when the result is the same. Here i have rather that the computation (the function call itself) is skipped as much as possible too. So I prefer a remember.

> This fixes it, but it might cause a lot of recompositions when admins show up

As long as the key doesn't change it shouldn't cause any recompositions, so as long as the list stays the same it shouldnt be a problem.

> Also, should this remember also include account

if this needs to change when the account changes then the account needs to be included in the key. But I think account change would force it to fully rebuild the compose tree due to navigation. So it wouldn't be necessary

dessalines: Gotcha, thx.

dessalines:
@MV-GH This should probably be the next to merge, since its getting cumbersome to deal with conflicts.

app/src/main/java/com/jerboa/Utils.kt
```diff
@@ -1569,6 +1634,12 @@ fun canMod(
     }
 }
 
+fun futureDaysToUnixTime(days: Int?): Int? {
+    return days?.let {
+        (Date.from(Instant.now().plus(Duration.ofDays(it.toLong()))).time / 1000L).toInt()
```
MV-GH, line 1639: Seems overly complex

Why not `Instant.now().plus(3, DAYS).toEpochMilli()` ?

MV-GH (approved):

app/src/main/java/com/jerboa/Utils.kt
dessalines: Thx, looks like this one works:

`Instant.now().plus(it.toLong(), ChronoUnit.DAYS).epochSecond.toInt()`

MV-GH: There is reason that is a long, this will overflow in 2038

dessalines: 👀 Then we'll also need to change that in the api-client also. Rust uses an `i64` so it should be fine there.

MV-GH: api-client? lemmy-js-client? thats not gonna be a problem since its a float which has a much bigger range and will only start to lose precision in about 300k years

dessalines: https://github.com/MV-GH/LemmyBackwardsCompatibleAPI/blob/master/app/src/commonMain/kotlin/it/vercruysse/lemmyapi/v0x18/datatypes/BanPerson.kt#L11 and the same for `BanPersonFromCommunity`

Those could be `Long` , but if you're auto-generating them, we might need to figure out a smart way to do it, because javascript had too many issues with `BigInt` when we tried to do that the first time.

MV-GH: Hmm that is annoying, I am guessing what ever tool you use to generate those TS types can't annotate its actual type for numbers?

I'll keep a manual list for now

dessalines: I commented over on your recent pr, but imo you might as well just replace every int with long.

PR #1326, opened 2024-01-20
Try removing meltwater cache.

dessalines:
There's a good chance this cache doesn't actually save any time.

### DIFF
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -4,30 +4,6 @@ steps:
     commands:
       - prettier -c "*.md" "*.yml"
 
-  restore-cache:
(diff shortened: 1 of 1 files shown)


MV-GH:
Yeah it doesn't save much, the biggest time i saved was with caching the deps between steps with the .gradle env and preventing the build tools download

dessalines:
Yeah, I verified it doesn't save much time at all. Mostly a waste.

MV-GH:
Can you upgrade the android images see https://github.com/CircleCI-Public/cimg-android/issues/127

dessalines:
Done.

Issue #52, opened 2022-01-21
Swipe upvote / downvote / click and hold


dessalines:

maltfield:
I came here to open a feature request: users to be able to:

1. swipe left on a post/card in the jerboa feed to downvote the post
2. swipe right on a post/card in the jerboa feed to upvote the post

This functionality is what I'd expect with reddit clients, namely [RedReader](https://github.com/QuantumBadger/RedReader)

Unfortunately, when I swipe left on the app today, nothing happens. If I swipe right, it opens the navigation menu.

There's no description for this issue, so I assume we're talking about the same thing?

As for the "click-and-hold", I don't know what that would do. But in RedReader it would open a submenu with the following options that can be tapped:

1. upvote
2. downvote
3. save
4. hide
5. report
6. reply
7. view in external browser
8. go to subreddit
9. block subreddit
10. pin subreddit to main menu
11. unsubscribe from subreddit
12. share
13. share comments
14. copy link
15. user profile
16. properties

I guess jerboa should implement a subset of these

dessalines:
Another possibility that would work:

- Tap once = upvote
- Tap again, downvote
- Tap again, remove vote. 

I'd consider any of these solutions good enough.

loadsas:
I really like the idea (maybe because I'm a redreader user ;) ). There are some issues here about accidently clicking something wrong instead of opening the "thread". This would be a good solution and a better UX (my opinion ;) )

Other issues

https://github.com/dessalines/jerboa/issues/488
https://github.com/dessalines/jerboa/issues/418
https://github.com/dessalines/jerboa/issues/394

Maybe there are more but I'm on holidays and only have a smartphone with me. (too straining)

heeplr:
> ... But in [RedReader](https://github.com/QuantumBadger/RedReader) ...

I'd like to add to this, imho, very important point:
RedReader's UX is extremely popular among users who need accessibility features to a point that Reddit exempted it from the upcoming API ban.

It's UI is **very** slick and can be configured to be ultra clean. It's worth imitating in every aspect.

Snow4DV:
WIP. Liked that feature back in the day in sync. 
Can be easily implemented with swipetodismiss, will refactor it and create the PR

![screen-20240120-125243~2_2](https://github.com/dessalines/jerboa/assets/32912696/399f8213-c705-42d6-93c6-fcb65018ce8e)

maltfield:
@Snow4DV thanks for working on this :)

[...]

Issue #1318, opened 2023-12-29
Make tap/hold actions customizable


MV-GH:
### Pre-Flight checklist

- [X] Did you check to see if this issue already exists?
- [X] This is a single feature request. (Do not put multiple feature requests in one issue)
- [X] This is not a question or discussion. (Use https://lemmy.ml/c/jerboa for that)

### Describe The Feature Request Below

Make the current tap / hold on comments actions customizable. Meaning you can select which actions happens based on a trigger.

In this case the two triggers are: tap, long tap (could be expanded like swipe, .. but that's for another FR)

The possible proposed actions are:
- Do nothing
- (Un)Collapse bar
- (un)collapse comment tree
- Upvote/unvote
- Save
- Share

PR #1327, opened 2024-01-21
Swipe post/comment to upvote/downvote/reply/save

Snow4DV:
This feature will let user swipe to upvote/downvote  posts/comments

Closes feature request #52  

Currently it is WIP (implemented the composable that will wrap post/comment composables)

Thinking about customization - how should it be implemented? Doing it straight-forward will add a lot of stuff to the settings entity (that is stored in room). 

Also probably it would be better to set it up as "swipe to start - upvote, swipe to end - downvote" but this will affect "swipe to go back/to open drawer" gesture

There are multiple ways: 
1) Do a bunch of presets (it will limit the customization but it is the easiest one that won't affect the current settings structure a lot)
2) Create separate object for gestures preferences and convert it using the TypeConverter to json (with gson) when storing in room (4 lists of swipeactiontypes)
4 settings will be added (left/right gesture actions for posts/comments)
3) You tell me :)

298275487-399f8213-c705-42d6-93c6-fcb65018ce8e

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/common/SwipeToAction.kt
+++ b/app/src/main/java/com/jerboa/ui/components/common/SwipeToAction.kt
@@ -1,0 +1,170 @@
+package com.jerboa.ui.components.common
+
+import androidx.compose.animation.animateColor
+import androidx.compose.animation.core.spring
+import androidx.compose.animation.core.updateTransition
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.RowScope
+import androidx.compose.foundation.layout.fillMaxHeight
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.requiredWidth
+import androidx.compose.material3.DismissDirection
+import androidx.compose.material3.DismissState
+import androidx.compose.material3.DismissValue
+import androidx.compose.material3.ExperimentalMaterial3Api
+import androidx.compose.material3.Icon
+import androidx.compose.material3.SwipeToDismiss
+import androidx.compose.material3.rememberDismissState
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableFloatStateOf
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.hapticfeedback.HapticFeedbackType
+import androidx.compose.ui.platform.LocalHapticFeedback
+import androidx.compose.ui.unit.dp
+import com.jerboa.SwipeActionType
+
+@Composable
+@ExperimentalMaterial3Api
+fun SwipeToAction(
+    leftActions: List<SwipeActionType>,
+    rightActions: List<SwipeActionType>,
+    swipeableContent: @Composable RowScope.() -> Unit,
+    swipeState: DismissState,
+) {
+    val haptic = LocalHapticFeedback.current
+
+    val leftActionsRanges =
+        remember(leftActions) { SwipeActionType.getActionToRangeList(leftActions) }
+    val rightActionsRanges =
+        remember(rightActions) { SwipeActionType.getActionToRangeList(rightActions) }
+
+    fun actionByState(state: DismissState): Pair<OpenEndRange<Float>, SwipeActionType>? {
+        return when (state.currentValue) {
+            DismissValue.DismissedToEnd -> {
+                leftActionsRanges.findLast { swipeState.progress in it.first }
+            }
+
+            DismissValue.DismissedToStart -> {
+                rightActionsRanges.findLast { swipeState.progress in it.first }
+            }
+
+            else -> null
+        }
+    }
+
+    val swipeAction = remember(swipeState) { actionByState(swipeState) }
+
+    SwipeToDismiss(
+        directions =
+            remember(leftActions, rightActions) {
+                setOfNotNull(
+                    if (leftActions.isNotEmpty()) DismissDirection.StartToEnd else null,
+                    if (rightActions.isNotEmpty()) DismissDirection.EndToStart else null,
+                )
+            },
+        state = swipeState,
+        background = {
+            val lastSwipeAction = remember { mutableStateOf<SwipeActionType?>(null) }
+            val transition = updateTransition(swipeState, label = "swipe state")
+            val color by transition.animateColor(
+                transitionSpec = {
+                    val currentAction = remember(swipeState) { actionByState(swipeState) }
+                    // vibrates when icon changes
+                    if (lastSwipeAction.value != currentAction?.second) {
+                        haptic.performHapticFeedback(HapticFeedbackType.TextHandleMove)
+                        lastSwipeAction.value = currentAction?.second
+                    }
+                    spring(stiffness = 50f)
+                },
+                label = "swipe color animation",
+                targetValueByState = { state ->
+                    val currentAction = remember(state) { actionByState(swipeState) }
+                    currentAction?.second?.getActionColor() ?: Color.Transparent
+                },
+            )
+            Box(
+                modifier =
+                    Modifier
+                        .fillMaxSize(),
+            ) {
+                Box(
+                    modifier =
+                        Modifier
+                            .fillMaxWidth(if (swipeState.progress != 1f) swipeState.progress else 0f)
+                            .fillMaxHeight()
+                            .background(color = color)
+                            .align(if (swipeState.currentValue == DismissValue.DismissedToStart) Alignment.TopEnd else Alignment.TopStart),
+                    contentAlignment = Alignment.CenterStart,
+                ) {
+                    val tint = Color.White
+                    val modifier =
+                        Modifier
+                            .padding(10.dp)
+                            .requiredWidth(35.dp)
+                            .fillMaxHeight()
+
+                    swipeAction?.second?.getImageVector()?.let { icon ->
+                        Icon(
+                            imageVector = icon,
+                            contentDescription = null,
+                            tint = tint,
+                            modifier = modifier,
+                        )
+                    }
+                }
+            }
+        },
+        dismissContent = { swipeableContent() },
+    )
+}
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun rememberSwipeActionState(
+    leftActions: List<SwipeActionType>,
+    rightActions: List<SwipeActionType>,
+    onAction: (action: SwipeActionType) -> Unit,
+): DismissState {
+    /*
+    This hacky solution is required because confirmValueChange lambda doesn't pass progress state
+     */
+    val leftActionsRanges =
+        remember(leftActions) { SwipeActionType.getActionToRangeList(leftActions) }
+    val rightActionsRanges =
+        remember(rightActions) { SwipeActionType.getActionToRangeList(rightActions) }
+    val progressState = remember { mutableFloatStateOf(1.0f) }
+    val dismissState =
+        rememberDismissState(
+            positionalThreshold = { 0f },
+            confirmValueChange = { dismissValue ->
+                val action =
+                    when (dismissValue) {
+                        DismissValue.DismissedToEnd -> {
+                            leftActionsRanges.findLast { progressState.floatValue in it.first }
+                        }
+
+                        DismissValue.DismissedToStart -> {
+                            rightActionsRanges.findLast { progressState.floatValue in it.first }
+                        }
+
+                        else -> {
+                            null
+                        }
+                    }
+                action?.second?.let { actionType ->
+                    onAction(actionType)
+                }
+                false // do not dismiss
+            },
+        )
+    progressState.floatValue = dismissState.progress
+    return dismissState
+}
--- a/app/src/main/java/com/jerboa/Utils.kt
+++ b/app/src/main/java/com/jerboa/Utils.kt
@@ -27,9 +27,14 @@ import androidx.annotation.RequiresApi
 import androidx.appcompat.app.AppCompatDelegate
 import androidx.browser.customtabs.CustomTabsIntent
 import androidx.compose.foundation.lazy.LazyListState
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.outlined.Bookmark
+import androidx.compose.material.icons.outlined.Comment
 import androidx.compose.material3.DrawerState
+import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.SnackbarDuration
 import androidx.compose.material3.SnackbarHostState
+import androidx.compose.runtime.Composable
 import androidx.compose.runtime.Stable
 import androidx.compose.runtime.mutableStateListOf
 import androidx.compose.runtime.snapshots.SnapshotStateList
@@ -38,6 +43,8 @@ import androidx.compose.ui.draw.drawBehind
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.Path
 import androidx.compose.ui.graphics.drawscope.DrawScope
+import androidx.compose.ui.graphics.vector.ImageVector
+import androidx.compose.ui.res.vectorResource
 import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.TextUnit
 import androidx.compose.ui.unit.dp
@@ -55,6 +62,7 @@ import com.jerboa.db.APP_SETTINGS_DEFAULT
 import com.jerboa.db.entity.AppSettings
 import com.jerboa.ui.components.common.Route
 import com.jerboa.ui.theme.SMALL_PADDING
+import com.jerboa.ui.theme.jerboaColorScheme
 import it.vercruysse.lemmyapi.v0x19.datatypes.*
 import kotlinx.collections.immutable.ImmutableList
 import kotlinx.collections.immutable.toImmutableList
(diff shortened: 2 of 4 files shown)


dessalines:
I wouldn't mind following the standard (I think @aeharding has with voyager):

- Swipe right: Upvote, then downvote
- Swipe left: Reply, then save/bookmark.

MV-GH:
For custom actions I do have something proposed see https://github.com/dessalines/jerboa/issues/1318

> As far as customization, I really don't think we should add or try to maintain custom sets of actions. 

Am I missing something but too it seems we can achieve this rather simple with one enum describing the actions and have in appSettings these additional settings

```kt
val swipeLeftPostAction: Int,
val swipeRightPostAction: Int,
val tapCommentAction: Int,
val longTapCommentAction: Int,
```

Snow4DV:
> For custom actions I do have something proposed see #1318
> 
> > As far as customization, I really don't think we should add or try to maintain custom sets of actions.
> 
> Am I missing something but too it seems we can achieve this rather simple with one enum describing the actions and have in appSettings these additional settings
> 
> ```kotlin
> val swipeLeftPostAction: Int,
> val swipeRightPostAction: Int,
> val tapCommentAction: Int,
> val longTapCommentAction: Int,
> ```

More like 4 enums :) 
one for short left swipe and one for long left swipe, same thing for right ones

Might also be good to add ability to keep all swipes on right side

I think it's not that important feature to make it fully-customizable. I'll just create four presets (all actions on the left - all actions on the right - votes left, reply and save right - votes right, reply and save left. Easy to implement/maintain, will be enough for 99% of users

MV-GH:
I saw it more as one enum with all these actions
image
that has different triggers

Snow4DV:
> that has different triggers

what 

> I saw it more as one enum with all these actions
> image
> that has different triggers

Made a mistake in my previous comment, i meant four variables that would store enum. 

That approach won't let you add four actions to one side (honestly i got used to this one because of sync)

MV-GH:
You mean like trigger that does both upvote and save?

Snow4DV:
> You mean like trigger that does both upvote and save?

That's how I see it: 
Screenshot_20240123-003416

Snow4DV:
> I wouldn't mind following the standard (I think @aeharding has with voyager):
> 
>     * Swipe right: Upvote, then downvote
> 
>     * Swipe left: Reply, then save/bookmark.
> 
> 
> As far as customization, I really don't think we should add or try to maintain custom sets of actions. That would be immensely complicated and cumbersome to maintain. A simple boolean setting for swipe actions, default true, should work.
> 
> For colors: rather than gradients and fading, I'd just do static colors. And I'd much prefer using the already existing [Material v3 colors](https://m3.material.io/styles/color/system/how-the-system-works): primary, secondary, tertiary. Make sure upvote and downvote colors match what those colors already are.



Used these colors for actions: 
```
Upvote -> MaterialTheme.colorScheme.primary
Downvote -> MaterialTheme.colorScheme.tertiary
Reply -> MaterialTheme.colorScheme.onPrimary
Save -> MaterialTheme.colorScheme.onTertiary
```

About colors' transitions: there were no gradients but i used animateColorAsState to achieve smooth transition between two colors. Made it fast enough so it's not notable but improves overall experience

app/src/main/java/com/jerboa/db/AppDB.kt
```diff
@@ -39,10 +39,11 @@ val APP_SETTINGS_DEFAULT =
         showPostLinkPreviews = true,
         postActionbarMode = 0,
         autoPlayGifs = false,
+        swipeToActionPreset = 1,
```
dessalines, line 42: Use your setting Enum, and `.ordinal` .

app/src/main/java/com/jerboa/feat/SwipeToAction.kt
```diff
@@ -0,0 +1,95 @@
...
+enum class SwipeToActionType() {
+    Upvote,
+    Downvote,
+    Reply,
+    Save,
+    ;
+
+    companion object {
+        fun getActionToRangeList(actions: List<SwipeToActionType>): List<Pair<OpenEndRange<Float>, SwipeToActionType>> {
+            val delta = if (actions.size > 2) 0.14f else 0.18f
```
dessalines, line 22: Not sure if here is where the issue is, but it switches from Upvote to Downvote way too quickly. Each action needs a longer swipe range (especially the first).

app/src/main/java/com/jerboa/ui/components/comment/CommentNodes.kt
```diff
@@ -58,6 +58,7 @@ fun CommentNodes(
     showAvatar: Boolean,
     blurNSFW: Int,
     showScores: Boolean,
+    swipeToActionPreset: Int,
```
dessalines, line 61: Don't pass the int, pass a simple settings preset enum around. You should only need to convert to the to / from ordinal at the DB layer.

app/src/main/java/com/jerboa/ui/components/post/PostActivity.kt
```diff
@@ -295,6 +298,17 @@ fun PostActivity(
                                     .testTag("jerboa:comments"),
                         ) {
                             item(key = "${postView.post.id}_listing", "post_listing") {
+                                val instantScores =
```
dessalines, line 301: What's the reason these are getting moved around?

app/src/main/res/values/strings.xml
```diff
@@ -415,8 +415,14 @@
     <string name="lock_post">Lock Post</string>
     <string name="unlock_post">Unlock Post</string>
     <string name="type_your_reason">Type your reason</string>
+    <string name="disabled_swipe_action_preset">Disabled</string>
```
dessalines, line 418: If you want to prefix these, then add that first, IE `swipe_action_preset_disabled`

dessalines (commented):
I'll do a more thorough review of this later, but my biggest concerns are moving instant voting state out of the post listing. You could avoid that by passing that info up and down as needed.

Also try not to use `Int` for that settings preset, when you really want the Enum.

app/src/main/java/com/jerboa/ui/components/comment/CommentNodes.kt
MV-GH: The problem is we kinda don't do that for any of the enums.

Snow4DV: > Don't pass the int, pass a simple settings preset enum around. You should only need to convert to the to / from ordinal at the DB layer.

If we do it in DB layer, it will mean that we'll need to use strings to store enums because AppSettings entity is directly used everywhere. I've seen that Ints are mostly used for that kind of stuff in Jerboa so won't it make it look a little inconsistent?

Or i can simply write a typeconverter

Snow4DV: > Don't pass the int, pass a simple settings preset enum around. You should only need to convert to the to / from ordinal at the DB layer.

Or i can simply do it like this: 
```
HomeActivity(
                            appState = appState,
                            .......
                            .......
                            postActionbarMode = appSettings.postActionbarMode,
                            swipeToActionPreset = SwipeToActionPreset.entries[appSettings.swipeToActionPreset],
                        )
```

MV-GH: * You can also do `appSettings.swipeToActionPreset.toEnum()`
*

app/src/main/java/com/jerboa/ui/components/post/PostActivity.kt
Snow4DV: > What's the reason these are getting moved around?

Honestly that wasn't my best approach, agree to that. Would be better to pass VoteType to child composables

app/src/main/java/com/jerboa/ui/components/comment/CommentNodes.kt
Snow4DV: > toEnum

Oh, that's cool, didn't think too much about this inline+reified approach. Found out it already exists in the code so used it

Snow4DV:
> I'll do a more thorough review of this later, but my biggest concerns are moving instant voting state out of the post listing. You could avoid that by passing that info up and down as needed.
> 
> Also try not to use `Int` for that settings preset, when you really want the Enum.

I've spent a while thinking about a way to keep instantscores down in the composable hierarchy and the only one that comes to my mind (like passing channel to the child composable) will break the states go down - events go up principle. Modifying postview & storing local copy of it with remember(postView) looks even worse to me honestly.

dessalines:
@MV-GH would probably be able to give better guidance than I could... But I believe you could pass down the actions  like `swipeUpVote() -> Unit`, and allow the instant scores state to be altered inside by those actions.

Snow4DV:
Android  moment 
image

I see that instantscores were refactored a little, finally got time, will take a look at it and improve according to new implementation

Snow4DV:
> @MV-GH would probably be able to give better guidance than I could... But I believe you could pass down the actions like `swipeUpVote() -> Unit`, and allow the instant scores state to be altered inside by those actions.

Reimplemented it another way: just nested it inside single post's composable, should be fine, instantscore lays where it should now. Fixed the ranges also, you may have a look when you have time

MV-GH:
I'll review it friday if it hasn't been reviewed by then.

app/src/main/java/com/jerboa/feat/SwipeToAction.kt
```diff
@@ -0,0 +1,103 @@
...
+            }
+        }
+
+        fun getDefaultLeftActions(): List<SwipeToActionType> {
+            return listOf(Reply, Save)
+        }
+
+        fun getDefaultRightActions(): List<SwipeToActionType> {
+            return listOf(Upvote, Downvote)
+        }
```
MV-GH, line 38: These aren't used? if so remove them

app/src/main/java/com/jerboa/feat/SwipeToAction.kt
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    @Composable
+    fun getImageVector(): ImageVector {
+        return when (this) {
+            Upvote -> ImageVector.vectorResource(id = R.drawable.up_outline)
+            Downvote -> ImageVector.vectorResource(id = R.drawable.down_outline)
+            Reply -> Icons.Outlined.Comment
```
MV-GH, line 46: 'Comment: ImageVector' is deprecated. Use the AutoMirrored version at Icons.AutoMirrored.Outlined.Comment

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
```diff
@@ -1076,127 +1088,164 @@ fun PostListing(
...
-                postActionbarMode = postActionbarMode,
-            )
+    val swipeState = rememberSwipeActionState(swipeToActionPreset) {
+        when (it) {
+            SwipeToActionType.Upvote -> {
+                instantScores =
+                    instantScores.update(VoteType.Upvote)
+                onUpvoteClick(postView)
+            }
+            SwipeToActionType.Downvote -> {
```
MV-GH, line 1098: What do we do for when downvote isn't available? Some instances have it disabled

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
```diff
@@ -1076,127 +1088,164 @@ fun PostListing(
...
+                instantScores =
+                    instantScores.update(VoteType.Downvote)
+                onDownvoteClick(postView)
+            }
+            SwipeToActionType.Reply -> {
+                onReplyClick(postView)
```
MV-GH, line 1104: I am not sure why but the reply to post in postlistings actions isnt' working for me at all. It doesn't seem to do anything

app/src/main/java/com/jerboa/ui/components/settings/lookandfeel/LookAndFeelActivity.kt
```diff
@@ -107,6 +107,8 @@ fun LookAndFeelActivity(
     val markAsReadOnScroll = rememberBooleanSettingState(settings.markAsReadOnScroll)
     val autoPlayGifs = rememberBooleanSettingState(settings.autoPlayGifs)
 
+    val swipeToActionPreset = rememberIntSettingState(settings.swipeToActionPreset)
```
MV-GH, line 110: Did you remove the actual settings button to change this preset by accident? Either way it needs to exist

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
MV-GH: The one in postview does work but no in the main feed

MV-GH (changes requested):

MV-GH:
@dessalines also should this be enabled by default? I'll let you decide

app/src/main/java/com/jerboa/db/AppDBMigrations.kt
```diff
@@ -384,6 +384,25 @@ val MIGRATION_26_25 =
...
+                "ALTER TABLE AppSettings ADD COLUMN swipe_to_action_preset INTEGER NOT NULL DEFAULT 1",
+            )
+        }
+    }
+
+val MIGRATION_27_26 =
+    object : Migration(27, 26) {
+        override fun migrate(db: SupportSQLiteDatabase) {
+            db.execSQL(
+                "ALTER TABLE AppSettings DROP COLUMN swipe_to_action_preset",
```
MV-GH, line 401:
```
android.database.sqlite.SQLiteException: near "DROP": syntax error (code 1 SQLITE_ERROR): , while compiling: ALTER TABLE AppSettings DROP COLUMN swipe_to_action_preset
                                                                                                    	at android.database.sqlite.SQLiteConnection.nativePrepareStatement(Native Method)
                                                                                                    	at android.database.sqlite.SQLiteConnection.acquirePreparedStatement(SQLiteConnection.java:1045)
```

It seems that DROP operations aren't supported

@dessalines should we remove all these downgrade migrations? They don't work anyway

Snow4DV: Ooh. It's a surprise to me that SQLite doesn't support drop column in tables. It can be done via creating a separate table without that column (i looked into it and i guess that's how others do it)

I think separate PR should be created for downgrade migrations fix

app/src/main/java/com/jerboa/ui/components/settings/lookandfeel/LookAndFeelActivity.kt
Snow4DV: Yep. Fixed it.

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
Snow4DV: By some reason empty lambda was passed somewhere to postlisting. Fixed it.

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
Snow4DV: Fixed it: Now action will not be shown, if current instance disabled downvotes. It applies to comments & posts

app/src/main/java/com/jerboa/feat/SwipeToAction.kt
Snow4DV: Fixed

app/src/main/java/com/jerboa/feat/SwipeToAction.kt
Snow4DV: Removed.

Snow4DV:
Have a look at resolved comments. Fixed everything & also improved behaviour when account is not logged in (now it doesn't update instanceScore and still shows toast as intended). 

About the latest comment: I think not working migrations should be left now so this could be improved in another pull request. This can be done by creating a separate table without newer column (see [this](https://stackoverflow.com/questions/60012776/drop-column-room-migration-android))

app/src/main/java/com/jerboa/db/AppDBMigrations.kt
MV-GH: yeah probably, but the down migrations aren't really used. So it won't be priority

app/src/main/java/com/jerboa/ui/components/common/SwipeToAction.kt
```diff
@@ -133,16 +140,23 @@ fun SwipeToAction(
 @Composable
 fun rememberSwipeActionState(
     swipeToActionPreset: SwipeToActionPreset,
-    onAction: (action: SwipeToActionType) -> Unit,
+    enableDownVotes: Boolean = true,
```
MV-GH, line 147: Remove this default, its a footgun. We have had many cases before, we forgot to fill it in.

MV-GH (approved):
Other then that one remark its fine

app/src/main/java/com/jerboa/ui/components/common/SwipeToAction.kt
Snow4DV: You're right, fixed it

MV-GH:
for light theme, the white icon doesn't really work.  

https://github.com/dessalines/jerboa/assets/67873169/dd8b5877-5eea-4a95-b9b4-af223902af7f

dessalines:
Use one of the theme background colors, probably surface.

app/src/main/java/com/jerboa/db/AppDBMigrations.kt
dessalines: Yeah its totally fine to remove them. One of the worst android bugs in existence: https://stackoverflow.com/questions/8045249/how-do-i-delete-column-from-sqlite-table-in-android

dessalines (commented):
Thx! Sry about my delay, I should be able to test again tmrw.

Snow4DV:
> Use one of the theme background colors, probably surface.

Surface color is already used by surface, so had to use something else. Had a problem with color collisions (error is tertiary when the theme is dark but reddish when it is light) so i used these colors to make them different and readable on light/dark themes: 
```
@Composable
    fun getActionColor(): Color {
        return when (this) {
            Upvote -> MaterialTheme.colorScheme.secondary
            Downvote -> MaterialTheme.colorScheme.error
            Reply -> MaterialTheme.colorScheme.inversePrimary
            Save -> MaterialTheme.colorScheme.primary
        }
    }
```

Now upvote/downvote colors also match in light theme (didn't match previously)

Snow4DV:
> for light theme, the white icon doesn't really work. You could add as black shim or use the color from the theme
> studio64_4YKa6RD9e5.mp4

I think white icons look good when they're placed on proper accent colors. Should be fine now (fixed theming when using both light/dark themes)

dessalines (commented):
Just tested, and it works really well! The colors look good to me, but we can always tweak them later.

The only issue I found: bookmarking / saving only works to save, but not to **unsave** it, unlike the button version.

You want to handle the conflicts, or should I?

Snow4DV:
> Just tested, and it works really well! The colors look good to me, but we can always tweak them later.
> 
> The only issue I found: bookmarking / saving only works to save, but not to **unsave** it, unlike the button version.
> 
> You want to handle the conflicts, or should I?

Totally missed it, happened due to lambda caching (it kept old instance of postView). Fixed that thing so now everything works correct

I resolved all conflicts so should be fine to merge now

dessalines (approved):
Sweet, thx!

PR #1328, opened 2024-01-21
Adding ability to lock / unlock posts. #1182

dessalines:
Was pretty easy.


app/src/main/java/com/jerboa/model/PersonProfileViewModel.kt
```diff
@@ -180,6 +181,20 @@ class PersonProfileViewModel(personArg: Either<PersonId, String>, savedMode: Boo
         }
     }
 
+    fun lockPost(form: LockPost) {
```
MV-GH, line 184: I have pending changes that deduplicates all these actions. But for now it is good. I'll refactor this later too

MV-GH (approved):

app/src/main/java/com/jerboa/model/PersonProfileViewModel.kt
dessalines: Sweet, thx.

PR #1331, opened 2024-01-22
Adding admin / mod view votes.

dessalines:
This is tested and working now.

https://github.com/dessalines/jerboa/assets/930215/9ead5df2-360f-494f-8fb3-70dc3b7860ef


MV-GH:
image
This is starting to get too long, could you collapse `Feature in` into a submenu?

MV-GH:
If there too many more actions that need to be added, maybe we should explore other options, as all mod actions into a sub menu or add a mod button next to the dot menu which shows all the mod actions

app/src/main/java/com/jerboa/ui/components/post/composables/PostOptionsDropdown.kt
```diff
@@ -428,104 +432,114 @@ fun PostOptionsDropdown(
...
-                        Pair(stringResource(R.string.lock_post), Icons.Outlined.CommentsDisabled)
-                    }
+        // You can do these actions on mods above you
+        if (amMod || amAdmin) {
+            PopupMenuItem(
+                text = stringResource(R.string.view_votes),
+                icon = ImageVector.vectorResource(R.drawable.up_filled),
+                onClick = {
+                    onDismissRequest()
+                    onViewVotesClick(postView.post.id)
```
MV-GH, line 478: This feature is only available in certain lemmy versions, use the FeatureFlag to enable it. `FF.listAdminVotes()`

dessalines:
> If there too many more actions that need to be added, maybe we should explore other options, move all mod actions into a sub menu or add a mod button next to the dot menu which shows all the mod actions. Doesn't need to be addressed in this PR

Very good point, and it makes sense to add a moderation sub-menu. I'll mark as ready for review when I get done with that.

app/src/main/java/com/jerboa/ui/components/viewvotes/post/PostLikesActivity.kt
```diff
@@ -0,0 +1,113 @@
...
+    val postLikesViewModel: PostLikesViewModel = viewModel(factory = PostLikesViewModel.Companion.Factory(postId))
+
+    Scaffold(
+        topBar = {
+            SimpleTopAppBar(
+                text = stringResource(R.string.post_votes),
+                onClickBack = onBack,
+            )
+        },
+        content = { padding ->
```
MV-GH, line 54: I think this padding should be applied to the Box

app/src/main/java/com/jerboa/ui/components/viewvotes/comment/CommentLikesActivity.kt
```diff
@@ -0,0 +1,113 @@
...
+    val commentLikesViewModel: CommentLikesViewModel = viewModel(factory = CommentLikesViewModel.Companion.Factory(commentId))
+
+    Scaffold(
+        topBar = {
+            SimpleTopAppBar(
+                text = stringResource(R.string.comment_votes),
+                onClickBack = onBack,
+            )
+        },
+        content = { padding ->
```
MV-GH, line 54: Same here?

dessalines: Good call.

app/src/main/java/com/jerboa/ui/components/post/composables/PostOptionsDropdown.kt
```diff
@@ -428,43 +435,61 @@ fun PostOptionsDropdown(
...
-                            person = postView.creator,
-                            community = postView.community,
-                            banned = postView.creator_banned_from_community,
-                        ),
-                        onDismissRequest,
-                        onBanFromCommunityClick,
+                // You can do these actions on mods above you
+
+                // These are all amMod || amAdmin
+                if (FeatureFlags(version = API.version.toVersion()).listAdminVotes()) {
```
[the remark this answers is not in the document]
dessalines: Sweet, I'll do that way.

MV-GH (approved):

MV-GH (approved):

PR #1337, opened 2024-01-27
InstantScores refactor + Int to Long changes

MV-GH:
Simplified the InstantScore stuff + convert ints to longs, would be a lot simpler if we used those aliases from the get go


app/src/main/java/com/jerboa/JerboaAppState.kt
```diff
@@ -68,11 +72,11 @@ class JerboaAppState(
         navController.navigate(Route.PRIVATE_MESSAGE_REPLY)
     }
 
-    fun toCommentReport(id: Int) {
+    fun toCommentReport(id: CommentId) {
```
dessalines, line 75: Thx for using these too.

dessalines (approved):
Thx for doing this.

PR #1339, opened 2024-01-30
Adding registration applications queue.

dessalines:
This adds an interface to handle registration applications for admins. Most of this is copied visually from lemmy-ui, but I'm open to improvements / suggestions.

One minor quirk (that also exists for lemmy-ui), is that the button on the bottom app bar only shows up **after** the site is fetched. 

The only way around that would be to persist admin and mod status in the DB, which seems a bit fragile since those can be updated non-locally. 

Demo:

https://github.com/dessalines/jerboa/assets/930215/ffb98cb0-c5da-4d89-a610-2111cc9560dc

#1182


app/src/main/java/com/jerboa/ui/components/registrationapplications/RegistrationApplications.kt
```diff
@@ -0,0 +1,496 @@
...
+                ) {
+                    Text(stringResource(R.string.deny))
+                }
+            }
+        }
+    }
+    HorizontalDivider(modifier = Modifier.padding(bottom = SMALL_PADDING))
+}
+
+@Preview
```
dessalines, line 462: I set these up, with a bunch of sample data too, but unfortunately android studio can't render them (I think because of the markdown view)

app/src/main/java/com/jerboa/model/SiteViewModel.kt
```diff
@@ -114,6 +133,16 @@ class SiteViewModel(private val accountRepository: AccountRepository) : ViewMode
         }
     }
 
+    private fun getUnreadAppCountTotal(): Long? {
```
dessalines, line 136: Made this nullable to denote that non-admins won't have it.

app/src/main/java/com/jerboa/ui/components/common/DropdownMenu.kt
```diff
@@ -280,3 +284,29 @@ fun CenteredPopupMenu(
         }
     }
 }
+
+@Composable
+fun UnreadOrAllOptionsDropDown(
```
dessalines, line 289: Moved this from Inbox.kt to a common dir, since applications can use it also.

app/src/main/java/com/jerboa/ui/components/home/BottomNavActivity.kt
```diff
@@ -55,47 +58,61 @@ import com.jerboa.ui.components.community.list.CommunityListActivity
...
 import com.jerboa.ui.components.person.PersonProfileActivity
+import com.jerboa.ui.components.registrationapplications.RegistrationApplicationsActivity
 import kotlinx.coroutines.launch
 
 enum class NavTab(
     val textId: Int,
     val iconOutlined: ImageVector,
     val iconFilled: ImageVector,
     val contentDescriptionId: Int,
+    val adminOnly: Boolean,
```
dessalines, line 69: When I add the report queue  next, I'll probably also use this.

app/src/main/res/values/strings.xml
```diff
@@ -443,4 +444,13 @@
     <string name="post_votes">Post votes</string>
     <string name="comment_votes">Comment votes</string>
     <string name="moderation">Moderation</string>
+    <string name="registrations">Registrations</string>
```
dessalines, line 447: *Registrations* seems a bit long for the title, but I couldn't think of anything else. In the bottom bar I use *Apps* for short.

app/src/main/java/com/jerboa/ui/components/registrationapplications/RegistrationApplications.kt
```diff
@@ -0,0 +1,496 @@
...
+                is ApiState.Failure -> ApiErrorText(appsRes.msg)
+                is ApiState.Holder -> {
+                    val apps = appsRes.data.registration_applications
+                    LazyColumn(
+                        state = listState,
+                        modifier =
+                            Modifier
+                                .simpleVerticalScrollbar(listState)
+                                .fillMaxSize()
+                                .imePadding(),
```
dessalines, line 246: I'm pretty unsure where this imePadding should live, but it seems to work here.

Its necessary because each registration application form has an expandable `deny_reason` markdown field.

MV-GH (changes requested):
> The only way around that would be to persist admin and mod status in the DB, which seems a bit fragile since those can be updated non-locally.

Yes but we do that already for defaultSortType and defaultListingType. We can add admin and mod status to account too. And they are much less likely to change than those two.

app/src/main/res/values/strings.xml
```diff
@@ -443,4 +444,13 @@
...
+    <string name="applicant">Applicant</string>
+    <string name="approve">Approve</string>
+    <string name="deny">Deny</string>
+    <string name="approved_by">Approved by</string>
+    <string name="denied_by">Denied by</string>
+    <string name="answer">Answer</string>
+    <string name="deny_reason">Deny reason</string>
+    <string name="apps">Apps</string>
```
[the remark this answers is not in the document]
dessalines: K fixed.

app/src/main/java/com/jerboa/ui/components/home/BottomNavActivity.kt
```diff
@@ -180,11 +197,13 @@ fun BottomNavActivity(
             Scaffold(
                 snackbarHost = { JerboaSnackbarHost(snackbarHostState) },
                 bottomBar = {
-                    if (appSettings.showBottomNav) {
+                    if (appSettings.showBottomNav && account.isReady()) {
```
dessalines, line 200: Unfortunately, this means an `anon` account will not show the bottom bar.

I couldn't find a way around this, because even tho I'm persisting `is_admin` to the DB, I still need to wait for it to read the database, otherwise the registration button will pop up in the list.

IE, even if I do `if (appSettings.showBottomNav && (account.isAnon() || account.isReady())`, it means that the bottom bar will show 5 tabs, then *suddenly* the 6th tab will pop up only after the DB `isAdmin` status is loaded.

MV-GH: I putting this on hold rn, until I have some time to look into this. Because this should definitely be possible in a way.

Because this has bigger consequences. It would make those navigation actions things completely inaccessible. Rn its either bottomnav or in the side panel. I'll probably have tomorrow time for this

dessalines: No probs. As a simple fix I could also remove that `account.isReady()` check. Its mainly just a visual annoyance that these extra bottom bar icons suddenly pop up, but that could be tackled at a later time.

MV-GH: Anon accounts are never ready thus isReady check fails for them. (Made it that way because this allowed anon Accounts go to through the checks and display "Not logged in" toasts )

MV-GH (approved):
What do you think of the change?

dessalines:
Just tested, and that still hides the bottom bar for anon accounts.

MV-GH:
Yes you were right, I fixed that.

I also found a different bug, one that has been present for a while, that invalidates the jwt when u swap to anon account. Weird that we never got a bug for this

dessalines:
Thx!

> I also found a different bug, one that has been present for a while, that invalidates the jwt when u swap to anon account.

I've been getting that too, but for some reason only on voyager.lemmy.ml . I tested switching back and forth from Anon to my  lemmy.ml account, and it doesn't seem to be an issue... not sure what's going on with voyager.lemmy.ml

MV-GH:
Weird, your lemmy account jwt token should also have been invalidated.

PR #1345, opened 2024-02-05
Temp workaround for compose-settings height bug.

### LINKED ISSUE #1338: Weirdly stretched settings list items bug
### Jerboa Version

0.0.55 debug / main branch (8b5c6fe)

### Android Version + Phone

Pixel 4 XL API 28/34 (physical device & emulator)

### Describe The Bug

By some reason menu items are stretched like this in the latest debug build
image

### To Reproduce

1. Build app using latest commit ([8b5c6fe](https://github.com/dessalines/jerboa/commit/8b5c6fe24fda700b294781078187a5797eafc42d))
2. Open the app
3. Go to settings

_No response_

dessalines:
- See https://github.com/alorma/Compose-Settings/issues/203
- Fixes #1338

### DIFF
--- a/app/src/main/java/com/jerboa/ui/theme/Sizes.kt
+++ b/app/src/main/java/com/jerboa/ui/theme/Sizes.kt
@@ -22,8 +22,11 @@ val THUMBNAIL_CARET_SIZE = 10.dp
 
 val DRAWER_ITEM_SPACING = 24.dp
 
-// TODO remove all DPs from code, put here.
+// TODO this needs to be removed when Intrinsic height min is fixed
+// See https://github.com/alorma/Compose-Settings/issues/203
+val SETTINGS_MENU_LINK_HEIGHT = 88.dp
 
+// TODO remove all DPs from code, put here.
 const val ICON_THUMBNAIL_SIZE = 96
 const val LARGER_ICON_THUMBNAIL_SIZE = 256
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/jerboa/ui/theme/Sizes.kt
```diff
@@ -22,8 +22,11 @@ val THUMBNAIL_CARET_SIZE = 10.dp
 
 val DRAWER_ITEM_SPACING = 24.dp
 
-// TODO remove all DPs from code, put here.
+// TODO this needs to be removed when Intrinsic height min is fixed
+// See https://github.com/alorma/Compose-Settings/issues/203
+val SETTINGS_MENU_LINK_HEIGHT = 88.dp
```
MV-GH, line 27: How was this height chosen?

Arbitrarily based on what fits? Would this still work when a user uses a different fontsize? That is much bigger?

dessalines: Its from [here](https://github.com/alorma/Compose-Settings/blob/main/compose-settings-ui/src/commonMain/kotlin/com/alorma/compose/settings/ui/internal/SettingsTileScaffold.kt#L28).

No need to get too complicated with it. We should be able to remove this hack with the next compose-settings release.

MV-GH (approved):

PR #1350, opened 2024-02-05
Fixing community pages, used wrong ID_TYPE. Fixes #1348

### LINKED ISSUE #1348: All community pages are broken in `0.0.56`
### Jerboa Version

0.0.56

### Android Version + Phone

all

### Describe The Bug

Every community page displays "Couldn't find community". 

### To Reproduce

1. Click on any community page
2. It displays the error above

NA

dessalines:


app/src/main/java/com/jerboa/ui/components/common/Route.kt
```diff
@@ -62,7 +62,7 @@ object Route {
 
         companion object {
             const val ID = "id"
-            val ID_TYPE = NavType.IntType
+            val ID_TYPE = NavType.LongType
```
dessalines, line 65: Dang compiler not finding this one

MV-GH (approved):

app/src/main/java/com/jerboa/ui/components/common/Route.kt
MV-GH: Yeah it sucks we cant do this type safe

dessalines: Oh well. I'll get another release out tmrw with these two fixes.

PR #1351, opened 2024-02-05
Use edge to edge

### LINKED ISSUE #1184: Adds padding above nav bar on return from imageviewer on Android 9 and below
### Jerboa Version

0.0.44

### Android Version + Phone

Android 9 and below

### Describe The Bug

this
image

### To Reproduce

open a image, go back

_No response_

MV-GH:
Behaviour should be the same as before.

Also fixes #1184 
Caused by that padding


MV-GH:
@dessalines I think to fix this, we might have to do some drastic refactors

I think it caused by the nested bottomNavHost with the drawer, it might have to made part of the root navhost and use state to enable the drawer/bottomnav visiblity

It is probably solveable in other way but I haven't found it yet. I think its applying some padding twice

dessalines:
Seems unavoidable, since that uiController's deprecated.

MV-GH:
Found the real culprit

```kt
             val bottomPadding =
                    if (selectedTab == NavTab.Search && WindowInsets.isImeVisible) {
                        0.dp
                    } else {
                        padding.calculateBottomPadding()
                    }
```

This was being applied to the navhost, which has weird side effects

app/src/main/java/com/jerboa/ui/theme/Theme.kt
```diff
@@ -55,7 +55,7 @@ fun JerboaTheme(
                 ),
             )
         } else {
-            pink()
+            blue()
```
MV-GH, line 58: I don't like the pink theme as default 👯

dessalines: Haha alright.

app/src/main/java/com/jerboa/ui/components/imageviewer/ImageViewerActivity.kt
```diff
@@ -261,6 +246,7 @@ fun ViewerHeader(
                 )
             }
         },
+        // pzv9dPd8iL
```
dessalines, line 249: Typo here.

PR #1354, opened 2024-02-07
Adding dependenciesInfo for f-droid builds. Fixes #1353

### LINKED ISSUE #1353: Question on permissions
My APK scanner got a few additional checks implemented in January, and on today's release of Jerboa reported:

```
! repo/com.jerboa_57.apk declares sensitive permission(s): android.permission.READ_MEDIA_AUDIO
 android.permission.READ_MEDIA_IMAGES android.permission.READ_MEDIA_VIDEO
 android.permission.READ_EXTERNAL_STORAGE
```

My *guess* is this is so local media can be included with posts – but as the app description does not mention it, I thought I'd better make sure before stating so in the permission descriptions:

image

Oh, as for that `DEPENDENCY_INFO_BLOCK`, that's rather easy to get rid of:

```gradle
android {
    dependenciesInfo {
        // Disables dependency metadata when building APKs.
        includeInApk = false
        // Disables dependency metadata when building Android App Bundles.
        includeInBundle = false
    }
}
```

For some background: that BLOB is supposed to be just a binary representation of your app's dependency tree. But as it's encrypted with a public key belonging to Google, only Google can read it – and nobody else can even verify what it really contains.

dessalines:


MV-GH (approved):
I hope this doesn't break play store builds

dessalines:
If it does, we'll try removing that from the bundle only.

PR #1355, opened 2024-02-08
Adding a locally generated changelog.

### LINKED ISSUE #1272: Use `last_version_code_viewed` rather than `changelog_viewed` for showing the changelog
### Pre-Flight checklist

Rather than using the `changelog_viewed` column, and having to do a migration to reset it, we should use a `last_version_code_viewed` column, and update it after viewing the current version. 

This can be copied from ThumbKey.

dessalines:
- Adds a last_version_code_viewed, that gets updated in the DB, and compared against the current version to show the changelog. This means we never have to manually update that column again.
- Add a generate_changelog.sh script that uses git-cliff. It copies the changelog into code assets, which can be done before the release.
- Fixes #1272


app/src/main/java/com/jerboa/db/AppDBMigrations.kt
```diff
@@ -300,15 +300,6 @@ val MIGRATION_21_22 =
         }
     }
 
-val MIGRATION_22_21 =
```
dessalines, line 303: I removed all these down migrations since they were useless.

app/src/main/java/com/jerboa/db/entity/AppSettings.kt
```diff
@@ -24,6 +24,7 @@ data class AppSettings(
         defaultValue = "0",
     )
     val themeColor: Int,
+    // TODO get rid of this column next time you regenerate the app
```
MV-GH, line 27: Why next time? why cant we drop it already?

app/src/main/java/com/jerboa/db/entity/AppSettings.kt
```diff
@@ -124,4 +125,9 @@ data class AppSettings(
         defaultValue = "1",
     )
     val swipeToActionPreset: Int,
+    @ColumnInfo(
+        name = "last_version_code_viewed",
+        defaultValue = "0",
+    )
+    val lastVersionCodeViewed: Int,
```
MV-GH, line 132: I would probably have made a separate table for this. It's not really a app setting. But this is fine

app/src/main/java/com/jerboa/db/repository/AppSettingsRepository.kt
```diff
@@ -40,16 +38,12 @@ class AppSettingsRepository(
...
             try {
-                Log.d("jerboa", "Fetching RELEASES.md ...")
-                // Fetch the markdown text
-                val releasesUrl =
-                    "https://raw.githubusercontent.com/dessalines/jerboa/main/RELEASES.md".toHttpUrl()
-                val req = Request.Builder().url(releasesUrl).build()
-                val res = API.httpClient.newCall(req).execute()
-                _changelog.value = res.body.string()
+                Log.d("jerboa", "Getting RELEASES.md from assets...")
+                val releasesStr = ctx.assets.open("RELEASES.md").bufferedReader().use { it.readText() }
```
MV-GH, line 45: I rather have this logic in the viewmodel. Just pass a stream or string.

I see that we did this before but we shouldn't have

MV-GH (approved):

app/src/main/java/com/jerboa/db/entity/AppSettings.kt
dessalines: android SQLite can't drop columns, you have to regenerate the entire table. Should probably be done at some point, but probably when we go over all the columns and everything we need to refactor.

app/src/main/java/com/jerboa/db/repository/AppSettingsRepository.kt
dessalines: I'll try to refactor that now.

app/src/main/java/com/jerboa/db/repository/AppSettingsRepository.kt
```diff
@@ -30,29 +19,12 @@ class AppSettingsRepository(
...
     }
 
     @WorkerThread
     suspend fun updatePostViewMode(postViewMode: Int) {
         appSettingsDao.updatePostViewMode(postViewMode)
     }
-
-    @WorkerThread
-    suspend fun updateChangelog() {
-        withContext(Dispatchers.IO) {
```
dessalines, line 44: I removed this `withContext` when I moved this to the viewmodel. It didn't seem to have any effect.

app/src/main/java/com/jerboa/db/entity/AppSettings.kt
MV-GH: We have done this before, see when I changed the default of a column

PR #1366, opened 2024-02-10
Remove material1 LocalContentColor.

dessalines:


MV-GH (approved):
Nice, The less dependencies on m1 the better

PR #1370, opened 2024-02-10
Dont show metadata title if it matches the post title.

dessalines:


app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
```diff
@@ -1737,10 +1739,14 @@ fun MetadataCard(post: Post) {
             Column(
                 modifier = Modifier.padding(MEDIUM_PADDING),
             ) {
-                Text(
-                    text = post.embed_title!!,
-                    style = MaterialTheme.typography.titleLarge,
-                )
+                if (post.name !== post.embed_title) {
```
MV-GH, line 1742: Isn't triple equals direct reference comparison? Isn't this not guaranteed to be true? Or did they change this behaviour with Kotlin? Either way I would do `!=`

MV-GH (changes requested):

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
dessalines: Fixed. Didn't know that made a difference for strings.

MV-GH (approved):

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
MV-GH: In Java each (string) object has its own unique ref, `==` is a direct reference comparison and `equals` is the overloaded logical comparison. But Java caches Strings and thus returns sometimes the same ref for the same string. So it works for some cases but not all. In Kotlin `==` is logical comparison (the equals) and `===` is the reference comparison

Issue #1372, opened 2024-02-12
Instead of showing Score, Upvote, and Downvote, show only Score and Upvote Percentage (or sentiment icon)


dessalines:
### Pre-Flight checklist

- [X] Did you check to see if this issue already exists?
- [X] This is a single feature request. (Do not put multiple feature requests in one issue)
- [X] This is not a question or discussion. (Use https://lemmy.ml/c/jerboa for that)

### Describe The Feature Request Below

@MV-GH I'd like your feedback before I start working on this.

Currently, on every post / comment, we show 3 pieces of info:

- A score in the header bar
- Upvote and downvote count in the footer / action bar.

Sample:

![Screenshot_2024-02-12-12-52-32-271-edit_com jerboa_1](https://github.com/dessalines/jerboa/assets/930215/c6793f47-0a47-4a95-8421-9002a2a3116e)

It'd be nice to simplify this (like some other apps I've seen), and show only

- A score and an upvote percentage in the header bar.
  - We could also hide the upvote percentage if its above a certain threshold (IE 80% liked)
  - And instead of a percentage number, we could show one of these [sentiment icons](https://fonts.google.com/icons?icon.query=sentiment)

This would move the pieces of info shown from 3 down to 2 (or 1 if its a well liked item).

MV-GH:
I am open for this, but I would lean on making this a setting at first. Also I prefer percentage (Hide this if its 95%+)

dessalines:
I've decided that this is a pervasive enough setting that it should probably live in the back-end.

I can implement this *before* we add that back-end setting referenced above, by using a dummy setting.

MV-GH:
We are gonna need something dummy regardless for backwards compatibility :/

Lemmy.world is not even on 0.19 yet so the backwards compatibility is still pretty important.

or for Lemmy version which don't implement that feature yet, we can just assume they want the old behaviour and only make the setting available for instances that support the feature

dessalines:
For sure, mainly I just want to hold off on adding a DB migration in jerboa, and instead make this a back-end user setting.

MV-GH:
Isn't this then a setting on Acccount too? like listing etc

dessalines:
Yep, I imagine it would be good to persist there also, just like `default_post_listing_type`, etc.

MV-GH:
Since which lemmy version was postlistingMode added?

PR #1378, opened 2024-02-14
Adding additional vote display modes

dessalines:
- Also removing redundant score counts from comment and post footers.
- Context: #1372

Full: 
Screenshot_2024-02-14-17-58-41-653-edit_com jerboa debug

Score and Upvote Pct (I'd like this to be the new default): 

Screenshot_2024-02-14-17-57-01-796-edit_com jerboa debug

Upvote pct: 

Screenshot_2024-02-14-17-57-39-362-edit_com jerboa debug

Upvote pct in list view: 

Screenshot_2024-02-14-17-57-52-874-edit_com jerboa debug

Hide all:

Screenshot_2024-02-14-18-12-40-282-edit_com jerboa debug


app/src/main/java/com/jerboa/ui/components/common/TimeAgo.kt
```diff
@@ -131,27 +135,126 @@ fun ScoreAndTime(
...
-            Text(
-                text = score.toString(),
-                color = scoreColor(myVote = myVote),
-                fontSize = MaterialTheme.typography.bodyMedium.fontSize.times(1.3),
-            )
+        val upvotePct = upvotePercentStr(
+            upvotes = upvotes,
+            downvotes = downvotes,
+        )
+        when (voteDisplayMode) {
```
dessalines, line 142: Main change is here. All these extra vote display modes get added to the ScoreAndTime composable, which is shown in the header bar of every comment / post.

app/src/main/java/com/jerboa/ui/components/common/VoteHelpers.kt
```diff
@@ -14,10 +14,8 @@ import com.jerboa.ui.theme.muted
 @Composable
 fun VoteGeneric(
     myVote: Int,
-    votes: Long,
```
dessalines, line 17: The footer bars now only show your vote color, with no text score.

app/src/main/java/com/jerboa/model/SiteViewModel.kt
```diff
@@ -211,10 +216,22 @@ class SiteViewModel(private val accountRepository: AccountRepository) : ViewMode
         }
     }
 
-    fun showScores(): Boolean {
+    // TODO this should probably be persisted rather than waited for
+    // For the current default, just use FullScores
+    fun voteDisplayMode(): VoteDisplayMode {
+        val defaultMode = VoteDisplayMode.Full
```
dessalines, line 222: Here's what makes this all temporary. When this actually comes back from the siteRes, then we can remove it.

dessalines:
@MV-GH Once this gets merged, I'll do another release.

app/src/main/java/com/jerboa/feat/Voting.kt
```diff
@@ -28,10 +31,44 @@ data class InstantScores(
...
+fun upvotePercent(
+    upvotes: Long,
+    downvotes: Long,
+): Float {
+    return (upvotes.toFloat() / (upvotes + downvotes))
+}
+
+fun formatPercent(pct: Float): String {
+    val formatted = "%.0f".format(pct * 100F)
+    return "$formatted%"
```
MV-GH, line 60: I think you can do this in one line by just escaping the `%`

app/src/main/java/com/jerboa/model/SiteViewModel.kt
MV-GH: If we don't persist this? Don't we keep resetting it each time the user opens the app again?

app/src/main/java/com/jerboa/ui/components/comment/mentionnode/CommentMentionNode.kt
```diff
@@ -54,16 +55,21 @@ fun CommentMentionNodeHeader(
...
-    showScores: Boolean,
+    voteDisplayMode: VoteDisplayMode,
 ) {
     CommentOrPostNodeHeader(
         creator = personMentionView.creator,
         score = score,
         myVote = myVote,
+        upvotes = upvotes,
+        downvotes = downvotes,
+        voteDisplayMode = voteDisplayMode,
```
MV-GH, line 72: Passing the whole InstantScores basically here? Why not pass the InstantScores itself then?

MV-GH (changes requested):

app/src/main/java/com/jerboa/model/SiteViewModel.kt
dessalines: The question is : Which local user settings should we be saving to our local database?

I think we only really need to save things required *before* the first fetch, so `default_sort_type` . But as for other things that come back with `getSite` ? I'm not sure.

app/src/main/java/com/jerboa/ui/components/comment/mentionnode/CommentMentionNode.kt
dessalines: Mmmk, I'll refactor all these to do instant scores.

app/src/main/java/com/jerboa/feat/Voting.kt
dessalines: That ended up not working `.format(` needs only formatting characters, but I made this more compact now.

dessalines: Ah yeah that did. Originally I tried `\\%` which failed.

app/src/main/java/com/jerboa/model/SiteViewModel.kt
MV-GH: We should discuss this, but not only the context of this setting but like sort options on each screen too.

For backwards compatibility it seems that we will need to save the above setting too unless it not persistenting for older versions is intended.

Also I don't think it is guaranteed that getSite has returned before feed loads? So initially it won't be correct occasionally?

dessalines: > Also I don't think it is guaranteed that getSite has returned before feed loads? So initially it won't be correct occasionally?

Yep, it won't be correct until after GetSite has loaded. In practice I haven't seen a case where the any feed loads *faster* than GetSite tho. 

In that `SiteViewModel`, the current things that *could* be persisted are admins, followList, voteDisplayMode, enableDownvotes, and showAvatars.

If we decide to persist them, we should do that as a later PR, all at once tho. I'm not 100% convinced its worth it... but we do already have everything set up (persisting them would work exactly like `default_sort_type`)

MV-GH: > Yep, it won't be correct until after GetSite has loaded. In practice I haven't seen a case where the any feed loads faster than GetSite tho.

That must be very situational. I just tested 5 times on my phone and all 5 times it did the feed request before the site request. You can check this by checking the log order for the initiation request

image


> admins, followList, voteDisplayMode, enableDownvotes, and showAvatars.

I care less about those are more about things we don't persist at all, like the sort types on each screen (communityview, postview, ...)

Maybe we should devise a much more flexible system that doesn't require so much overhead (as code) for persisting more fields.

dessalines: Sort types on each screen should just use the `default_sort_type` , which is already saved in the DB.

We *could* add a string / JSON field for user settings, but I really don't like putting untyped things in the DB, as that makes migrations and many other things a nightmare.

MV-GH: We can't reuse default sort because it is in a different context+ different enum (sometimes)

For example comment sort options
or sort options in saved/own profile would be different context where you wouldn't want to reuse the same sort as for the feed

dessalines: I'll add default_comment_sort for users and sites eventually: https://github.com/LemmyNet/lemmy/issues/4128

I'm not sure what would be best, but at least for now lets just add more columns to the account table as necessary.

dessalines:
@MV-GH Anything else I need to do here?

MV-GH:
Just persistence, does it now default to the status quo? Else you can go ahead and we'll continue this later

dessalines:
Yes rn it defaults to the status quo. I'll open an issue about persisting more account settings in a sec.

PR #1379, opened 2024-02-15
Fixing view votes crash. Fixes #1357

### LINKED ISSUE #1357: View votes crashes jerboa when there are multiple pages of votes
### Jerboa Version

main

### Android Version + Phone

any

### Describe The Bug

I get the `Conflicting Key` bug, implying that something is going wrong with the appending function

### To Reproduce

1. Click moderation -> view votes on any item
2. If its more than one page, jerboa crashes

See above

dessalines:


app/src/main/java/com/jerboa/model/CommentLikesViewModel.kt
```diff
@@ -58,7 +60,11 @@ class CommentLikesViewModel(val id: CommentId) : ViewModel() {
             likesRes =
                 when (newRes) {
                     is ApiState.Success -> {
-                        val appended = appendData(oldRes.data.comment_likes, newRes.data.comment_likes)
+                        val appended =
+                            getDeduplicateMerge(
+                                oldRes.data.comment_likes,
+                                newRes.data.comment_likes,
+                            ) { it.creator.id }
```
MV-GH, line 67: Is that unique per commentLike post?

MV-GH (approved):

app/src/main/java/com/jerboa/model/CommentLikesViewModel.kt
dessalines: Yep, user can only like an item once.

PR #1381, opened 2024-02-15
Adding mod ability to distinguish comment.

### LINKED ISSUE #287: Add distinguish comment
This would allow [distinguishComment](https://join-lemmy.org/api/classes/LemmyHttp.html#distinguishComment)

IE speak as moderator.

dessalines:
- Also removed shields on PersonProfileLink to only show when the comment is distinguished.
- Fixes #287


app/src/main/java/com/jerboa/model/PersonProfileViewModel.kt
```diff
@@ -268,6 +269,21 @@ class PersonProfileViewModel(personArg: Either<PersonId, String>, savedMode: Boo
...
+        viewModelScope.launch {
+            distinguishCommentRes = ApiState.Loading
+            distinguishCommentRes = API.getInstance().distinguishComment(form).toApiState()
+
+            when (val distinguishRes = distinguishCommentRes) {
+                is ApiState.Success -> {
+                    updateComment(distinguishRes.data.comment_view)
+                }
+
+                else -> {}
```
MV-GH, line 282: Display failure as toast?

dessalines: Check all the ones above this. If we do that we can do it for all of them as part of a cleanup later.

app/src/main/java/com/jerboa/ui/components/person/PersonProfileLink.kt
```diff
@@ -94,20 +93,13 @@ fun PersonProfileLink(
             }
         }
         if (showTags) {
-            if (isModerator) {
+            if (isDistinguished) {
                 Icon(
                     imageVector = Icons.Outlined.Shield,
```
MV-GH, line 98: I don't think we should reuse this icon for a different meaning. Isn't there something like a pin icon that we can use?

dessalines: Would the security or localpolice icons work? https://fonts.google.com/icons?selected=Material+Symbols+Outlined:security:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=shield

MV-GH: Hmm too similar imo, I was looking for a pin like icon but I couldnt find any at first glance

dessalines: Pins are also already used for stickying posts. Distinguish is kind of a *speak officially* thing. 

Its pry okay to use any shield variant, because I removed showing the mod / admin shields by default with this PR.

MV-GH: Ah I was confused what distinguishes means, is it just that the poster has the ability to mark that they are a mod and it should be visible in the comment that they post?

dessalines: Not the poster, but any moderator / admin. They can mark comments where they want to *speak officially* as a moderator.

MV-GH: Yean then it's fine

MV-GH (approved):

PR #1382, opened 2024-02-15
Adding legal / privacy policy info.

### LINKED ISSUE #729: Have link to Legal / Privacy policy.
- [ ] Did you check to see if this issue already exists?
- [ ] Is this only a single feature request? Do not put multiple feature requests in one issue.

Please consider alerting new users that their upvotes and downvotes are publicly federated.  It is just the way that Lemmy works, and not a Jerboa problem per se, but it would be nice for the user experience if this was communcated on the Jerboa end.  

Perhaps the very first time an account votes, a small tool tip could appear with a message similar to:

-  "This is your first time voting!  Please be aware that your votes are publicly associated with your account, and can be seen by other users" 

Or some better message, just as long as it's clearly communicated.

dessalines:
- Fixes #729

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/home/sidebar/SiteSidebarActivity.kt
+++ b/app/src/main/java/com/jerboa/ui/components/home/sidebar/SiteSidebarActivity.kt
@@ -1,9 +1,16 @@
 package com.jerboa.ui.components.home.sidebar
 
 import android.util.Log
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.Policy
 import androidx.compose.material3.ExperimentalMaterial3Api
+import androidx.compose.material3.Icon
+import androidx.compose.material3.IconButton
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/jerboa/ui/components/home/sidebar/SiteSidebarActivity.kt
```diff
@@ -14,24 +21,43 @@ import com.jerboa.ui.components.common.SimpleTopAppBar
...
             }
         }
 
     Scaffold(
         topBar = {
             SimpleTopAppBar(
                 text = title,
-                onBackClick,
+                onClickBack = appState::popBackStack,
+                actions = {
```
dessalines, line 42: Added it to the action bar of the site info page, rather than the top of home. Could go either way.

MV-GH (approved):

PR #1396, opened 2024-02-22
Show full error on post creation failing

MV-GH:
You can choose if you want this or the old the system.

Either way the eventual log viewer will be preferably.

[jerboa-app-debug.zip](https://github.com/dessalines/jerboa/files/14378255/jerboa-app-debug.zip)

@epragalakis

Can you install the above apk and show me which error you are getting


dessalines (approved):
Seems fine, although its probably best to do these errors as toasts, and only have them on the top-level activities / pages.

epragalakis:
Thank you!! I will test it when I find some time and paste the error in the initial thread.

PR #1401, opened 2024-02-24
Fix sidebar showing admin screen to all users

### LINKED ISSUE #1398: Non admin mods are unable to ban people
### Jerboa Version

0.0.59

### Android Version + Phone

N/A

### Describe The Bug

ezgif-3-2c47401774

A user reported that he could not ban people, as a non admin mod. Even though he can in Lemmy UI

The error in the pic does seem to indicate that this comes from the backend though.

### To Reproduce

ban a user a non admin mod

_No response_

MV-GH:
Should fix #1398 

That person was just a mod, and that action was banning users from an instance not from a community.

We don't have that action yet right that you can ban users from just a community?

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/common/AppBars.kt
+++ b/app/src/main/java/com/jerboa/ui/components/common/AppBars.kt
@@ -131,14 +131,7 @@ fun BottomAppBarAll(
     NavigationBar(
         modifier = modifier,
     ) {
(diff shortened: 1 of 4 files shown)


dessalines:
I'm pretty sure I do have banning users from the community, I'll double check it all tho.

dessalines (approved):
Sweet, thx.

Issue #1400, opened 2024-02-24
Add a `Full` vote display mode? And what should be the default?


dessalines:
### Pre-Flight checklist

- [X] Did you check to see if this issue already exists?
- [X] This is a single feature request. (Do not put multiple feature requests in one issue)
- [X] This is not a question or discussion. (Use https://lemmy.ml/c/jerboa for that)

### Describe The Feature Request Below

I have a [PR to add these vote display modes to the lemmy back-end](https://github.com/LemmyNet/lemmy/pull/4450).

But we have two unresolved questions: 

1. Should we add a `Full` mode alongside `Score+Downvote` ?
2. What should be the default until the back end gets merged + deployed? 

For 1, I don't think its necessary, mainly because I'm trying to avoid this mess: ![Screenshot_2024-02-12-12-52-32-271-edit_com jerboa_1](https://github.com/dessalines/jerboa/assets/930215/c6793f47-0a47-4a95-8421-9002a2a3116e) . 

I removed all this distracting extra unecessary info, and put everything in the node header bar. But should it also show the upvotes alongside score? IE `score, upvotes, downvotes, time` . IMO that's too much info, and it isn't necessary.

For 2, I think we should switch the default to `Score+UpvotePct`. It looks the cleanest, and seems to be what most lemmy front-ends are using now.

For reference, here's what these modes look like in jerboa (without full)

Score + Downvote (the current default): 
![Screenshot_2024-02-14-17-58-41-653-edit_com jerboa debug](https://github.com/dessalines/jerboa/assets/930215/5168953a-e0f2-4b62-bc6d-c172bb874807)

Score and Upvote Pct (I'd like this to be the new default): 

![Screenshot_2024-02-14-17-57-01-796-edit_com jerboa debug](https://github.com/dessalines/jerboa/assets/930215/3f630aba-806e-41c5-a8e2-318f2611c645)

Upvote pct: 

![Screenshot_2024-02-14-17-57-39-362-edit_com jerboa debug](https://github.com/dessalines/jerboa/assets/930215/3f4d432d-24b9-46b6-9a6c-8c7159f1ac0f)

Upvote pct in list view: 

![Screenshot_2024-02-14-17-57-52-874-edit_com jerboa debug](https://github.com/dessalines/jerboa/assets/930215/0faac34e-619d-42dc-ae5c-acdd2fbb3828)

Hide all:

[...]

PR #1402, opened 2024-02-25
Adding distinct Full Score, and ScoreAndDownvote vote display modes.

dessalines:
- Also making ScoreAndUpvotePercentage the new default, until these
  get added in the back-end.
- Context https://github.com/dessalines/jerboa/issues/1400

New FullScore mode:

Screenshot_2024-02-24-18-57-12-202-edit_com jerboa debug


MV-GH:
I wouldn't turn `Full` into that, this is just a much visual clutter as before with the arrows. I would just keep `Full` as original arrows design

dessalines:
I don't *really* want to move the score numbers back into the the footer bars, as that's showing the score in three places again. Totally agree that this is cluttered, but some ppl might want all that info. It also only shows for <90% approval.

MV-GH:
> I don't really want to move the score numbers back into the the footer bars, 

What is your rational for this? Code complexity? UX: that is more visually cluttered? Blocks future design changes to the footer?

My rational for the old design, just boils down to that users don't like change to something they are used to. And the fact we can still easily achieve the previous design without much complexity

> as that's showing the score in three places again. 

Whats the third place? Footer, header

dessalines:
Score-related data in 2 distinct spots, besides just one. No other lemmy app does this either.

We also shouldn't be hesitant also to change things just because people get used to the old way. Front end is always like that, ppl are resistant to even the tiniest change, regardless of how much better it is. We will never be able to make anything look better if we use inertia as our benchmark. 

The complexity also goes up now that we have these  modes, IMO we should just keep them in one spot, otherwise it gets a lot of questions like: Should the percent be in the footer, or the header? Should score+downvotes be in both the header and the footer?

dessalines:
I'm also open to other places, or ways to group the score info. One thing I like about jerboa that I don't see other apps doing tho, is that both posts and comments have **score + time** in the top right. Other apps seem to only do that for comments, and treat posts differently.

dessalines:
This is going to get entirely reworked in the back end into a larger set of bools: https://github.com/LemmyNet/lemmy/pull/4450

But I think this is still okay to merge, since the main thing this adds is a switch to score + upvote-pct as default, at least until the next release.

MV-GH (approved):

PR #1404, opened 2024-02-25
Fix that non admin mods can see messages report tab

### LINKED ISSUE #1384: Hide "Messages" section in the reports inbox for non-admins
### Jerboa Version

0.0.59

### Android Version + Phone

Android 14, Google Pixel 8 Pro

### Describe The Bug

The messages section should be hidden for non-admins, but it's shown for everyone.

### To Reproduce

1. Go to Reports inbox as a normal lemmy user
2. See the Messages section

_No response_

MV-GH:
Fix #1384


dessalines (approved):
Nice, thx.

PR #1405, opened 2024-02-25
Fix tab title size. Fixes #1403

### LINKED ISSUE #1403: Tab font size too small
### Jerboa Version

0.0.60

### Android Version + Phone

Android 14, Google Pixel 7

### Describe The Bug

Font size of about, posts and comments tabs in the profile section is considerably smaller than in previous versions. Almost can't see it. 

Applies to tabs across the board, in messages too.

Notified you of this on Lemmy, opening the issue as per your request. Thanks for all the great work. 

### To Reproduce

Just run v0.0.60 and pay attention to tabs font size. 

_No response_

dessalines:


MV-GH:
This not the approach I would take.
This is just a bandaid fix, for which future changes would also need to apply custom changes.

Compose already has appropriately set the correct typography textstyles for their components.

We just treat them as all the same. See the table that I posted. I suggest we fix this structurally, by having the correct defaults. 

9 pixels for `titleSmall` doesn't work and is never appropriate, 9px for `LabelSmall` is.

dessalines:
In that case, would you be able to fix up `Type.kt`, as part of a larger rework?

MV-GH:
Sure, I'll do some more research on this, incase if I can find a cleaner solution else I am probably going with a lot of custom defaults

dessalines:
Thx, I'll close this one then.

PR #1410, opened 2024-02-25
Fix super and sub script not being properly rendered

### LINKED ISSUE #1341: Sub and super script in markdown doesn't work
### Jerboa Version

0.0.55

### Android Version + Phone

n/a

### Describe The Bug

The sub and sup markdown actions don't work, it puts the symbols but the markdown renders them as normal.

### To Reproduce

press the sub/sup button, preview the changes

_No response_

MV-GH:
image

Fix #1341

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/common/MarkdownHelper.kt
+++ b/app/src/main/java/com/jerboa/ui/components/common/MarkdownHelper.kt
@@ -32,6 +32,7 @@ import com.jerboa.util.markwon.BetterLinkMovementMethod
 import com.jerboa.util.markwon.ForceHttpsPlugin
 import com.jerboa.util.markwon.MarkwonLemmyLinkPlugin
 import com.jerboa.util.markwon.MarkwonSpoilerPlugin
+import com.jerboa.util.markwon.ScriptRewriteSupportPlugin
 import io.noties.markwon.AbstractMarkwonPlugin
 import io.noties.markwon.Markwon
 import io.noties.markwon.MarkwonConfiguration
@@ -92,6 +93,7 @@ object MarkdownHelper {
                 .usePlugin(ForceHttpsPlugin())
                 // email urls interfere with lemmy links
                 .usePlugin(LinkifyPlugin.create(Linkify.WEB_URLS))
+                .usePlugin(ScriptRewriteSupportPlugin())
                 .usePlugin(MarkwonLemmyLinkPlugin())
                 .usePlugin(MarkwonSpoilerPlugin(true))
                 .usePlugin(StrikethroughPlugin.create())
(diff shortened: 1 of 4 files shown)


MV-GH:
image

This doesn't work in Lemmy UI

So I assume ~ and ^ are markdown specs set by lemmy?

Too bad there never was a formal syntax spec for markdown.

Is this something that could be considered added to Lemmy UI or does Jerboa need to support ~ and ^ too?

That will be more complex as Markwon is kinda of enigma

dessalines:
Yep, lemmy uses a few things that aren't in the commonmark spec (spoilers, sub/sup). Mostly we follow the standards for extensions used by the markdown-it extensions: https://github.com/markdown-it/markdown-it?tab=readme-ov-file#syntax-extensions

app/src/main/java/com/jerboa/ui/components/common/InputFields.kt
```diff
@@ -584,7 +584,8 @@ fun MarkdownHelperBar(
         IconButton(
             onClick = {
                 simpleMarkdownSurround(
-                    "~",
+                    startText = "<sub>",
```
dessalines, line 587: This isn't the right way to do it, we have to follow the conventions [here](https://join-lemmy.org/docs/users/02-media.html?highlight=markdown#text) (but it looks like I need to fix super and subscript there also) , so that all lemmy clients can support it. 

This needs to be fixed in the markdown renderer.

The best way to test it to use lemmy-ui, and make sure the rendered, and view source, match both on jerboa and lemmy-ui.

dessalines: I'm fairly sure also that these already exist in [eternity](https://codeberg.org/Bazsalanszky/Eternity), and could be taken from there.

MV-GH: I checked eternity only supports superscript and its slightly different syntax too. its just `^`

image

Their execution is very complex. 

Imo I think we can achieve something better with a plugin that just rewrites ^^ and ~~ to their html tags equivalent. But from user perspective it will look like its `^^`. That way we wont need to maintain all this super complex logic that seems to depend on other custom markwon nodes.

dessalines: Totally agree. There's a simple example of altering the markdown string before processing here: https://noties.io/Markwon/docs/v4/core/plugins.html#process-markdown

It'd be much easier to do regex to replace `^...^` with `<sup>...</sup>`.

MV-GH:
image

Works pretty great, and much simpler

dessalines (approved):
Perfect, this is so much better.

PR #1412, opened 2024-02-26
Fix admin actions visibility for non admin mods

### LINKED ISSUE #1408: Non admin mod can access Admin actions through mod menu
### Jerboa Version

0.0.59

### Android Version + Phone

n/a

### Describe The Bug

Non admin mod can access Admin actions through mod menu

Context #1398 

Heard back from that user, they explained that they actually got to that screen through moderation menu. So it must mean we failed to hide that action.

### To Reproduce

Open mod menu on a non admin mod account

I suppose we forgot to filter here either.

_No response_

### LINKED ISSUE #1411: Non admin mods can't see moderation menu from the main feed
### Jerboa Version

0.0.60

### Android Version + Phone

n/a

### Describe The Bug

I looked into it, for moderators we give null in the feed because there is no list available

and then we do this

```
    val amAdmin =
        remember(admins) {
            amAdmin(
                admins = admins,
                myId = account.id,
            )
        }

    val amMod =
        remember {
            amMod(
                moderators = moderators,
                myId = account.id,
            )
        }
```

So both false, thus they cant see the menu

But since we cache those values in the account can't we just use those ?

### To Reproduce

..

_No response_

MV-GH:
Fix #1411 
Fix #1408 

- Fixed some admin actions being shown to non admin mods
- Did more deep research into the non admin can't ban user from community. I couldn't reproduce this at all. *User was certain though it was a community ban and not a instance ban* 

- Added a simulateModerators only for post options popup menu, it is not completely correct but this way non admin mods which probably make up most of the mods can actually moderate from the main feed. In case of the edge case moderator moderates a higher mod, the call will fail. So not that bad

- Also fixed the weird movement of the post when opening the morePostOptions menu (regression)
see https://stackoverflow.com/questions/77010371/prevent-popup-from-adding-padding-in-a-column-with-arrangement-spacedbylarge-p

### DIFF
--- a/app/src/main/java/com/jerboa/feat/ModActions.kt
+++ b/app/src/main/java/com/jerboa/feat/ModActions.kt
@@ -1,6 +1,10 @@
 package com.jerboa.feat
 
-import it.vercruysse.lemmyapi.v0x19.datatypes.CommunityModeratorView
+import android.content.Context
+import com.jerboa.MainActivity
+import com.jerboa.db.entity.Account
+import com.jerboa.findActivity
+import it.vercruysse.lemmyapi.v0x19.datatypes.CommunityId
 import it.vercruysse.lemmyapi.v0x19.datatypes.PersonId
 import it.vercruysse.lemmyapi.v0x19.datatypes.PersonView
 import java.time.Instant
@@ -12,29 +16,25 @@ import java.time.temporal.ChronoUnit
 fun canMod(
     creatorId: PersonId,
     admins: List<PersonView>?,
-    moderators: List<CommunityModeratorView>?,
-    myId: PersonId?,
+    moderators: List<PersonId>?,
+    myId: PersonId,
     onSelf: Boolean = false,
 ): Boolean {
-    return if (myId !== null) {
-        // You can do moderator actions only on the mods added after you.
-        val adminIds = admins?.map { a -> a.person.id }.orEmpty()
-        val modIds = moderators?.map { m -> m.moderator.id }.orEmpty()
+    // You can do moderator actions only on the mods added after you.
+    val adminIds = admins?.map { a -> a.person.id }.orEmpty()
+    val modIds = moderators.orEmpty()
 
-        val adminsThenMods = adminIds.toMutableList()
-        adminsThenMods.addAll(modIds)
+    val adminsThenMods = adminIds.toMutableList()
+    adminsThenMods.addAll(modIds)
 
-        val myIndex = adminsThenMods.indexOf(myId)
-        if (myIndex == -1) {
-            false
-        } else {
-            // onSelf +1 on mod actions not for yourself, IE ban, remove, etc
-            val subList = adminsThenMods.subList(0, myIndex.plus(if (onSelf) 0 else 1))
-
-            !subList.contains(creatorId)
-        }
-    } else {
+    val myIndex = adminsThenMods.indexOf(myId)
+    return if (myIndex == -1) {
         false
+    } else {
+        // onSelf +1 on mod actions not for yourself, IE ban, remove, etc
+        val subList = adminsThenMods.subList(0, myIndex.plus(if (onSelf) 0 else 1))
+
+        !subList.contains(creatorId)
     }
 }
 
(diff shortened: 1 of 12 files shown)


app/src/main/java/com/jerboa/feat/ModActions.kt
```diff
@@ -12,29 +16,25 @@ import java.time.temporal.ChronoUnit
...
-        if (myIndex == -1) {
-            false
-        } else {
-            // onSelf +1 on mod actions not for yourself, IE ban, remove, etc
-            val subList = adminsThenMods.subList(0, myIndex.plus(if (onSelf) 0 else 1))
-
-            !subList.contains(creatorId)
-        }
-    } else {
+    // You can do moderator actions only on the mods added after you.
```
MV-GH, line 23: Does the moderator hierarchy not apply to admins? i.e. Any admin can moderate any admin?

dessalines: It does apply to other admins. Check L28 and L35, it adds admins, then mods for the ordered hierarchy.

app/src/main/java/com/jerboa/feat/ModActions.kt
```diff
@@ -45,15 +45,30 @@ fun futureDaysToUnixTime(days: Long?): Long? {
...
-    myId: PersonId?,
-): Boolean {
-    return admins?.map { it.person.id }?.contains(myId) ?: false
+/**
+ * In screens with posts from different communities we don't have access to moderators of those communities
+ * So that means that non admin mods can't moderate those posts from that screen
+ *
+ * So this is QoL were we simulate the mods of the community
+ * It is not completely accurate as it doesn't take into account the hierarchy of mods
+ */
```
dessalines, line 60: This really shows that the `canMod` logic needs to be added on the back-end: https://github.com/LemmyNet/lemmy/issues/4365

Once I eventually get some time to work on that, we'd be able to remove the mod list from almost every response, and code like this.

app/src/main/java/com/jerboa/ui/components/post/PostActivity.kt
```diff
@@ -309,7 +309,7 @@ fun PostActivity(
                     is ApiState.Failure -> ApiErrorText(postRes.msg, padding)
                     is ApiState.Success -> {
                         val postView = postRes.data.post_view
-                        val moderators = postRes.data.moderators
+                        val moderators = remember(postRes) { postRes.data.moderators.map { it.moderator.id } }
```
dessalines, line 312: You have a remember for moderators both in the postlisting, and here. One can probably be removed.

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
```diff
@@ -1645,9 +1644,8 @@ fun PostListingCard(
                 .padding(vertical = MEDIUM_PADDING)
                 .clickable { onPostClick(postView) }
                 .testTag("jerboa:post"),
-        verticalArrangement = Arrangement.spacedBy(MEDIUM_PADDING),
```
dessalines, line 1648: Why remove this one, and add the spacers?

MV-GH: See StackOverflow link

app/src/main/java/com/jerboa/feat/ModActions.kt
MV-GH: One thing I forgot, how does the `self` option play into this? Which mod actions can you do on yourself? like lockpost? How would that work with canMod from server side?

Currently we don't do that correctly either. canMod should probably return which actions can be done?

app/src/main/java/com/jerboa/feat/ModActions.kt
MV-GH: It's not about the code, but rather does Lemmy also have the hierarchy for admin

app/src/main/java/com/jerboa/ui/components/post/PostActivity.kt
MV-GH: Those remember different things?

app/src/main/java/com/jerboa/feat/ModActions.kt
dessalines: Yep admins also have a hierarchy.

app/src/main/java/com/jerboa/feat/ModActions.kt
dessalines: I just looked through the lemmy-ui code, and its only `distinguish` that you can do to yourself, and anyone lower than you. We could probably change that to just `isModOrAdmin && myComment`

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
dessalines: Ah I see, that's a weird one.

dessalines (approved):

app/src/main/java/com/jerboa/feat/ModActions.kt
MV-GH: Only distinguish? Can top admin not pin his own comment? Or lock his own posts?

PR #1419, opened 2024-03-02
Fix git cliff

dessalines:


MV-GH (approved):
The CI pipeline got cancelled cuz a a different one started, it should only cancel a pipeline for a older commit in the same branch

PR #1420, opened 2024-03-02
Working on upgrading compose-settings.

dessalines:


app/src/main/java/com/jerboa/ui/components/settings/account/AccountSettings.kt
```diff
@@ -33,10 +32,10 @@ import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.text.input.KeyboardCapitalization
 import androidx.compose.ui.text.input.KeyboardType
 import androidx.compose.ui.text.input.TextFieldValue
-import com.alorma.compose.settings.storage.base.rememberBooleanSettingState
-import com.alorma.compose.settings.storage.base.rememberIntSettingState
+import com.alorma.compose.settings.storage.disk.rememberBooleanSettingState
```
MV-GH, line 35: Does this persist the settings for us?

This shouldn't be done for account right? These settings are different per account

dessalines: We really don't need to remember anything, since the remembering is done in the DB. But I'm not sure exactly how to integrate that with alorma as inputs. I'll see if it can be done without these.

MV-GH: The previous remembers aren't for persisting, just how state works in compose https://developer.android.com/jetpack/compose/state

dessalines: alorma's compose setting I mean, supposedly its saveable and restoring to disk.

PR #1421, opened 2024-03-03
Add non mirrored only votes swipe action

### LINKED ISSUE #1416: Swipe-to-vote gestures are backwards ("swipe right" to upvote)
### Jerboa Version

0.059

### Android Version + Phone

Android

### Describe The Bug

The UX for the swipe gesture is backwards, currently:
1. Swipe to the right is downvote and
2. Swipe to the left is upvote

The standard UX is:
1. Swipe to the right for things you like
2. Swipe to the left for things you don't like

For reference, see this article on How To Geek that describes "swipe right" vs "swipe left"
 * https://www.howtogeek.com/702931/what-do-swipe-left-and-swipe-right-mean/

Moreover, I expect that you'll find most reddit apps have "swipe right for upvote" and "swipe left for downvote". I can only speak to RedReader, which is widely recognized as the most user-accessible reddit app (so much so that it was given an exemption from the API changes so [RedReader could continue to use the reddit API for free](https://old.reddit.com/r/RedReader/comments/145du4j/update_4_redreader_granted_noncommercial/) -- because their UX is so good). And, yes, **RedReader has "swipe right for upvote" and "swipe left for downvote"**

The above-described UX is so commonly ingrained in our culture that it's become common slang. Consider the entry in Urban Dictionary for "swipe right"

 * https://www.urbandictionary.com/define.php?term=Swipe+right

> **Swipe Right**
> (verb) A phrase used to describe your acceptance of something.
> ...
> [example] "These burritos are so good! I'm glad we swiped right on Chipotle."

### Solution

The solution to this is to swap the directions of the votes on Jerboa, such that:

1. Swipe to the right is upvote and
2. Swipe to the left is downvote

### To Reproduce

1. Download Jerboa v0.0.59 from F-Droid
2. Tap on the hamburger menu
3. Tap on `Settings`
4. Tap on `Look and feel`
5. Tap on the `Preset for swipe gesture` drop-down menu
6. Tap on `Only votes`
7. Click the back button
8. Click the back button
9. Attempt to "swipe right" on something to upvote it
10. Post gets a downvote
11. Get confused

_No response_

MV-GH:
- Add non mirrored only votes swipe action
- Decrease sensitivity slightly, not sure how to reduce the vertical sensitivity though
- Also changed default to disabled (This feature is not stable enough)

Fix #1416


app/src/main/java/com/jerboa/db/AppDB.kt
```diff
@@ -41,11 +41,11 @@ val APP_SETTINGS_DEFAULT =
         showPostLinkPreviews = true,
         postActionBarMode = 0,
         autoPlayGifs = false,
-        swipeToActionPreset = SwipeToActionPreset.DEFAULT.ordinal,
+        swipeToActionPreset = SwipeToActionPreset.DISABLED.ordinal,
```
dessalines, line 44: I agree, disabled default is probably best.

dessalines (approved):
Looks good, although I think one more issue with that ticket, is that every place we use the word `swipe right`, it actually means `swipe left` . But feel free to merge when you see fit.

MV-GH:
@Snow4DV Was it intended to work as the direction mention in the text you had written it. Or was the text wrong? I think most users want these actions to be mirrored?

Snow4DV:
> @Snow4DV Was it intended to work as the direction mention in the text you had written it. Or was the text wrong? I think most users want these actions to be mirrored?

I guess I just made whole jerboa community aware of me never using tinder lol :DD 

Definitely a mistake on my side - good fix!


And I too agree on disabling it by default. Default behavior of swipetodismiss turned to be pretty bad in terms of nested scrolling. I tried improving it but it basically means rewriting it from scratch and I'm too low on time atm (hope google would fix swipetodismiss rather sooner than later)

app/src/main/java/com/jerboa/feat/SwipeToAction.kt
Snow4DV, line 1: That doesn't really fix the issue, I tried

MV-GH: Yeah it doesn't ;/

Maybe messing with the dampner options?

Snow4DV: > Yeah it doesn't ;/
> 
> Maybe messing with the dampner options?

I also tried this workaround: https://stackoverflow.com/a/76604939 (change velocity threshold of anchoreddraggable) modifying stuff using java's reflections and it behaves the same sadly :( 

I'll try digging deeper when i have time

dessalines (approved):

PR #1423, opened 2024-03-04
More horizontal line fixes.

dessalines:
I removed a few more pointless horizontal lines, and it does look much cleaner IMO:

https://github.com/dessalines/jerboa/assets/930215/34379b59-6e30-4bf2-a78e-fdd291c7383b


MV-GH:
Hmm i think the divide between comments and post in postview could still use a horizontal divider?

MV-GH:
Boost doesn't go that far but other G apps do seem to go in this direction too. Does still feel unnatural to me but I'll probably get used to it.

dessalines:
> Hmm i think the divide between comments and post in postview could still use a horizontal divider?

I'll add this shortly.

MV-GH (approved):

PR #1428, opened 2024-03-05
Add first frame video support for comment videos

MV-GH:
Coil has support to show first frames of videos.

Now this doesn't work for the image viewer. Not sure why.
Also discovered that the `components` override each in the newBuilder() thing but setting it for gifImage  loader which is used by the imageviewer doesn't solve it. (Just displays black instead of failure)

Add logging for that login action, but i doubt that it errors. Else the delete action wouldn't succeed


dessalines (approved):
:+1:

PR #1442, opened 2024-03-10
Fix migration changing the the app settings

### LINKED ISSUE #1440: Screenshots prevented by security settings
### Jerboa Version

0.0.61

### Android Version + Phone

Samsung Galaxy A50

### Describe The Bug

When attempting to take a screenshot anywhere in the app (tested on main screen, comments, viewing media, and settings menu), it is prevented, with a system message banner saying it's due to security settings.

### To Reproduce

1. Open the app
2. Attempt to take a screenshot anywhere

_No response_

MV-GH:
This is really annoying, the order of these column definitions is very important.

This caused half the settings to be switched with two columns lower.

This fixes that

It is actually miracle that this didnt cause any crashes (well if it did we would have found it earlier). I don't see how to add a test for this to prevent this from happening again.

We will have to be very careful with doing big changes to appSettings again

Fixes #1440

I have verified this with a pre 30 db version app and  31 has the exact same settings.

### DIFF
--- a/app/schemas/com.jerboa.db.AppDB/31.json
+++ b/app/schemas/com.jerboa.db.AppDB/31.json
@@ -1,0 +1,279 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 31,
+    "identityHash": "9ba87658605a1a57d0a4bebb42d35828",
+    "entities": [
+      {
+        "tableName": "Account",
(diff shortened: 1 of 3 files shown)


dessalines (approved):
Yup it is, android makes DB migrations an absolute nightmare. 

I also tested this and it works.

PR #1443, opened 2024-03-10
Fix issues caused by edge to edge

### LINKED ISSUE #1436: Nav buttons in post view lets comments draw behind them
### Jerboa Version

0.0.61

### Android Version + Phone

android 11

### Describe The Bug

The comments appear behind the navigation systembar

Introduced in the edge to edge pr

### To Reproduce

open a post with comments

_No response_

### LINKED ISSUE #1437: Weird padding above keybaord
### Jerboa Version

0.0.61

### Android Version + Phone

android 11

### Describe The Bug

Everywhere where you use the keyboard its eems to add a bit more padding.

Seems very similar to the bug I had  with edge to edge where it drew extra padding at the bottom everywhere

Are we applying some padding somewhere?

Either way this is most likely introduced in the edge to edge change

### To Reproduce

reply somewhere, extend body length until it hides behind the keyboard, then you can see the extra padding when u move the body up and down.

image

_No response_

### LINKED ISSUE #1439: Replies tab scrolling issue
### Jerboa Version

0.0.61

### Android Version + Phone

Android 13; Samsung S20FE 5g(SM-G781B/DS)

### Describe The Bug

With multiple new replies, scrolling doesn't go further than the top of the screen. I'm assuming that this issue will effect replies long enough to be longer than one full screen.

### To Reproduce

1. Goto Replies
2. Have more than one reply or a reply with a very large body
3. Try to scroll
4. ???

_No response_

### LINKED ISSUE #1441: Hiding text on navigation systembar clips it to unreasonibly small size.
### Jerboa Version

0.0.61

### Android Version + Phone

Android 11

### Describe The Bug

https://github.com/dessalines/jerboa/assets/67873169/af8d9b04-6816-438f-8b9f-63bcd55ba861

### To Reproduce

 have disable text in navigation bar disabled
 
 navigation bar looks weirdly small clippped now

_No response_

MV-GH:
I guess we didn't need edge to edge. Causes too much problems atm

Fixes #1436 
Fixes #1441 
Fixes #1439
Fixes #1437

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/community/list/CommunityListActivity.kt
+++ b/app/src/main/java/com/jerboa/ui/components/community/list/CommunityListActivity.kt
@@ -1,8 +1,12 @@
 package com.jerboa.ui.components.community.list
 
 import android.util.Log
+import androidx.compose.foundation.layout.PaddingValues
+import androidx.compose.foundation.layout.WindowInsets
+import androidx.compose.foundation.layout.consumeWindowInsets
 import androidx.compose.foundation.layout.imePadding
 import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.systemBarsPadding
 import androidx.compose.material3.DrawerState
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Scaffold
@@ -45,6 +49,7 @@ fun CommunityListActivity(
     blurNSFW: BlurNSFW,
     drawerState: DrawerState,
     showAvatar: Boolean,
+    padding: PaddingValues? = null,
 ) {
     Log.d("jerboa", "got to community list activity")
 
@@ -55,8 +60,20 @@ fun CommunityListActivity(
 
     val scope = rememberCoroutineScope()
 
+    val baseModifier = if (padding == null) {
+        Modifier
+    } else {
+        // https://issuetracker.google.com/issues/249727298
+        // Else it also applies the padding above the ime (keyboard)
+        Modifier.padding(padding)
+            .consumeWindowInsets(padding)
+            .systemBarsPadding()
+    }
+
     Surface(color = MaterialTheme.colorScheme.background) {
         Scaffold(
+            contentWindowInsets = WindowInsets(0, 0, 0, 0),
+            modifier = baseModifier,
             topBar = {
                 CommunityListHeader(
                     openDrawer = {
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/jerboa/ui/components/home/BottomNavActivity.kt
```diff
@@ -238,7 +236,7 @@ fun BottomNavActivity(
                         )
                     }
                 },
-            ) { _ ->
+            ) { padding ->
```
MV-GH, line 239: This change was need to propagate the padding of the navigation bar (ours) else content hides behind it. We can't use the previous approach that applies it to the navhost, that causes weird issues

app/src/main/java/com/jerboa/ui/components/home/BottomNavActivity.kt
```diff
@@ -308,6 +311,7 @@ fun BottomNavActivity(
                             siteViewModel = siteViewModel,
                             drawerState = drawerState,
                             blurNSFW = appSettings.blurNSFW.toEnum(),
+                            padding = padding,
```
MV-GH, line 314: Bit annoying, at some point I'll see if I can lift this all out of this nested navhost and just have a single one

app/src/main/res/values/themes.xml
```diff
@@ -2,6 +2,7 @@
 <resources>
 
     <style name="Theme.Jerboa" parent="Theme.AppCompat.NoActionBar">
+        <item name="android:windowTranslucentStatus">true</item>
```
MV-GH, line 5: If i don't do this, it applies the status bar padding twice, this could have also been due to the nested scaffold.

app/src/main/java/com/jerboa/ui/components/community/list/CommunityListActivity.kt
```diff
@@ -55,8 +60,20 @@ fun CommunityListActivity(
 
     val scope = rememberCoroutineScope()
 
+    val baseModifier = if (padding == null) {
+        Modifier
+    } else {
+        // https://issuetracker.google.com/issues/249727298
```
MV-GH, line 66: When using edge to edge, it also applies the navigationbar padding above the IME

dessalines (approved):
Tested, and everything seems to work well, thx.

PR #1446, opened 2024-03-11
Fix migrations 1

dessalines:
Okay, I strongly typed all our default values, and double checked the column order (2 were in the wrong place iirc).

It'd be good if you could test this also, just to be sure it doesn't break existing installs.

### DIFF
--- a/app/src/main/java/com/jerboa/db/AppDB.kt
+++ b/app/src/main/java/com/jerboa/db/AppDB.kt
@@ -11,41 +11,69 @@ import com.jerboa.db.dao.AccountDao
 import com.jerboa.db.dao.AppSettingsDao
 import com.jerboa.db.entity.Account
 import com.jerboa.db.entity.AppSettings
-import com.jerboa.feat.SwipeToActionPreset
+import com.jerboa.toBool
 import com.jerboa.ui.theme.DEFAULT_FONT_SIZE
 import java.util.concurrent.Executors
+
+// Warning: Be careful about changing any of these defaults, as it will mean you'll need to
+// regenerate the database
+// Unfortunately can't use enum ordinals here, because Room's compile time annotations
+// don't allow it
+//
+// Use Int for Bools
+const val DEFAULT_THEME = 0
+const val DEFAULT_THEME_COLOR = 0
+const val DEFAULT_LAST_VERSION_CODE_VIEWED = 0
+const val DEFAULT_POST_VIEW_MODE = 0
+const val DEFAULT_POST_NAVIGATION_GESTURE_MODE = 0
+const val DEFAULT_SHOW_BOTTOM_NAV = 1
+const val DEFAULT_SHOW_COLLAPSED_COMMENT_CONTENT = 0
+const val DEFAULT_SHOW_COMMENT_ACTION_BAR_BY_DEFAULT = 1
+const val DEFAULT_SHOW_VOTING_ARROWS_IN_LIST_VIEW = 1
+const val DEFAULT_SHOW_PARENT_COMMENT_NAVIGATION_BUTTONS = 0
+const val DEFAULT_NAVIGATE_PARENT_COMMENTS_WITH_VOLUME_BUTTONS = 0
+const val DEFAULT_USE_CUSTOM_TABS = 1
+const val DEFAULT_USE_PRIVATE_TABS = 0
+const val DEFAULT_SECURE_WINDOW = 0
+const val DEFAULT_BLUR_NSFW = 1
+const val DEFAULT_SHOW_TEXT_DESCRIPTIONS_IN_NAVBAR = 1
+const val DEFAULT_BACK_CONFIRMATION_MODE = 1
+const val DEFAULT_MARK_AS_READ_ON_SCROLL = 0
+const val DEFAULT_SHOW_POST_LINK_PREVIEWS = 1
+const val DEFAULT_POST_ACTION_BAR_MODE = 0
+const val DEFAULT_AUTO_PLAY_GIFS = 0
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/jerboa/db/AppDB.kt
```diff
@@ -11,41 +11,69 @@ import com.jerboa.db.dao.AccountDao
...
 import com.jerboa.ui.theme.DEFAULT_FONT_SIZE
 import java.util.concurrent.Executors
 
+// Warning: Be careful about changing any of these defaults, as it will mean you'll need to
+// regenerate the database
+// Unfortunately can't use enum ordinals here, because Room's compile time annotations
+// don't allow it
+//
+// Use Int for Bools
+const val DEFAULT_THEME = 0
```
dessalines, line 24: I really wanted to be able to use the Enums and ordinals here, but Room makes you use constants for their `currentValue` annotations.

app/src/main/java/com/jerboa/ui/components/settings/lookandfeel/LookAndFeelActivity.kt
```diff
@@ -116,7 +116,6 @@ fun LookAndFeelActivity(
         appSettingsViewModel.update(
             AppSettings(
                 id = 1,
-                viewedChangelog = settings.viewedChangelog,
```
dessalines, line 119: Dropped this column, as its superceded by `last_changelog_viewed`

MV-GH:
There is a test in AndroidTest that goes through the migrations to ensure it doesn't fail on migrating between versions but I will test it regardless.

app/src/main/java/com/jerboa/db/AppDBMigrations.kt
```diff
@@ -469,6 +470,81 @@ val MIGRATION_30_31 =
...
+val MIGRATION_31_32 =
+    object : Migration(31, 32) {
+        override fun migrate(db: SupportSQLiteDatabase) {
+            // Fix wrong ordering and defaults in previous migration
+
+            db.execSQL(
+                """
+                   CREATE TABLE IF NOT EXISTS AppSettingsBackup (
+                      `id` INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
+                      `font_size` INTEGER NOT NULL DEFAULT ${DEFAULT_FONT_SIZE},
```
MV-GH, line 482: You can't do something like this, the migrations need to stay exactly the same.

If we make migration version 33, say we change a default.

We change the constant and now we have also changed migration of 32.

Room will rightly throw a error that the checksum of migration 32 checksum mismatches. Because that means that some users their db schema 32 is different. That can be very problematic

MV-GH:
Which columns where in the wrong place?

app/src/main/java/com/jerboa/db/AppDBMigrations.kt
```diff
@@ -469,6 +470,81 @@ val MIGRATION_30_31 =
...
+                      `swipe_to_action_preset`,
+                      `last_version_code_viewed`
+             FROM AppSettings
+            """,
+            )
+            db.execSQL("DROP TABLE AppSettings")
+            db.execSQL("ALTER TABLE AppSettingsBackup RENAME to AppSettings")
+
+            // Reset a few messups to default
+            db.execSQL("UPDATE AppSettings SET post_actionbar_mode = $DEFAULT_POST_ACTION_BAR_MODE")
```
MV-GH, line 544: There are probably more but only the enums matter, bc if they get assigned a value bigger than their enum size (due swapping between two enums which are bigger and that user having the biggest value) it would cause a NPE when we select the enum.

MV-GH: Imo this is enough unless you want to reset more enums.

app/src/main/java/com/jerboa/db/AppDBMigrations.kt
```diff
@@ -469,6 +470,81 @@ val MIGRATION_30_31 =
...
+                      `last_version_code_viewed` INTEGER NOT NULL DEFAULT $DEFAULT_LAST_VERSION_CODE_VIEWED
+                    )
+                """.trimIndent(),
+            )
+
+            // Need to select explicitly, because mark_changelog_viewed was dropped
+            db.execSQL(
+                """
+            INSERT INTO AppSettingsBackup SELECT
+                      `id`,
```
MV-GH, line 513: I assume you don't want to correct the existing columns which have the wrong values assigned?

Because 

postNavigationGestureMode -> post_navigation_gesture_mode -> show_collapsed_comment_content -> show_collapsed_comment_content
(For a 29 to 32)

If did want to do this you will have to delay the removal of that old column. (or do it in two steps: two queries)

And have this insert be based on the final correct order. (Manually mapping the correct columns is a headache took me 10m to do the example correct)

app/src/main/java/com/jerboa/db/AppDBMigrations.kt
dessalines: Alright, can change them.

app/src/main/java/com/jerboa/db/AppDBMigrations.kt
dessalines: No I didn't bother, and just made them match the exact order of `AppSettings` .

dessalines:
iirc the two out of place, were `show_collapsed_comment_content`, and `post_action_bar_mode` . I'll fix it so that the migration doesn't use these defaults tho.

MV-GH (changes requested):
Linter, looks good else

dessalines:
K, I'll get a release out after its merged.

PR #1452, opened 2024-03-12
Rework fontsize, better defaults

### LINKED ISSUE #1403: Tab font size too small

MV-GH:
|                | Compose default in SP | Jerboa default in Px | Jerboa default in SP now |
|----------------|-----------------------|----------------------|----------------------|
| headlineLarge  | 32                    | 16                   | 22* |
| headlineMedium | 28                    | 13                   | 16* |
| headlineSmall  | 24                    | 9                    | 14* |
| titleLarge     | 22                    | 16                   |  22                    | 
| titleMedium    | 16                    | 13                   | 16                    |
| titleSmall     | 14                    | 9                    | 14                    |
| bodyLarge      | 16                    | 16                   | 16                    
| bodyMedium     | 14                    | 13                   | 14                    |
| bodySmall      | 12                    | 9                    | 12                    |
| labelLarge     | 14                    | 16                   | 14                    |
| labelMedium    | 12                    | 13                   | 12                    | 
| labelSmall     | 11                    | 9                    | 11                    |
| displayLarge   | 57                    | x                    | x |
| displayMedium  | 45                    | x                    | x |
| displaySmall   | 36                    | x                    | x |

I made it scale with the ratio of the bodyLarge, meaning that if you set font size to 14 bodyLarge becomes 14.
And the other styles resize at the same ratio (14/16 = 87.5% size)

Fixes #1403 

I am not done yet. But I recommend checking out if you see anything standing out. I made it so that most fields stay the same.


app/src/main/java/com/jerboa/ui/theme/Type.kt
```diff
@@ -1,76 +1,61 @@
...
-const val FONT_MEDIUM_ADD = -3
-const val FONT_LARGE_ADD = 0
-const val LINE_HEIGHT_MULTIPLIER = 1.5
+const val COMPOSE_BODY_LARGE_DEFAULT = 16
+val BASE_TYPOGRAPHY = Typography()
 
-// Set of Material typography styles to start with
-// Example from here: https://egeniq.com/blog/dynamic-font-sizes-with-jetpack-compose/
 fun generateTypography(baseFontSize: Int): Typography {
+    val customRatio = baseFontSize.toFloat() / COMPOSE_BODY_LARGE_DEFAULT
```
dessalines, line 9: This approach does seem much simpler, thx.

MV-GH:
Somethings I don't get

How it looked the old way:

image

this is twice titleMedium somehow `new` has more weight??

but using the new system

titleMedium is bigger but (even when only 14 vs original 13)
Both have the same font weight

image

MV-GH:
Made it look like this

image


The documented changes are:
- HeaderTitle
- Tabs
- PostAndCommunityContextHeader
- SimpleTopBar (Indifferent to this one was 16 now is 22, kept compose default)

Everything else should be the exact same as before

Opinions?

app/src/main/java/com/jerboa/ui/components/common/AppBars.kt
```diff
@@ -811,3 +813,35 @@ fun ActionTopBar(
         },
     )
 }
+
+@Composable
+fun DualHeaderTitle(
```
dessalines, line 818: Thx for extracting this.

dessalines (approved):
Just tested this, and it fixed the primary issue, and sizing works much better now. Thx!

PR #1465, opened 2024-04-01
Suggestion: redesign the PostListingList Composable

rodrigo-fm:
I thought that removing the username, moving the community name to the top and adding the community icon preview would make it cleaner and prettier. Let me know your thoughts on this.

**Before**

drawing

**After**

drawing

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/common/MarkdownHelper.kt
+++ b/app/src/main/java/com/jerboa/ui/components/common/MarkdownHelper.kt
@@ -252,7 +252,7 @@ object MarkdownHelper {
                 )
             },
             update = { textView ->
-                previewMarkwon!!.setMarkdown(textView, markdown)
+                previewMarkwon?.setMarkdown(textView, markdown)
             },
             modifier = modifier,
         )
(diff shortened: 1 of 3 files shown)


MV-GH:
Showing those community avatars should already be an setting that the user can control.

I believe this change is only regarding the list view mode, now my opinion:

List view mode is a compact mode where it only shows what matters for a post, the title. This should be first, second should be the meta data about the post. Your redesign makes it more consistent with the other modes but it reduces the clearity, (Not sure how to put this in words but before it was much easier/faster to read the titles)

For some context. Jerboa is heavily influenced by `Boost for Reddit` in all its design aspects. We believe (me and Dessalines) that it has the best UX out of all Reddit clients. And this includes this compact mode which follows BfR.

So I do prefer `before` but nothing is stopping is us from adding more post view modes but it should be a bigger change than just this.

MV-GH:
The setting already exists though its `show avatar`

dessalines:
I also lean *a tiny bit* towards the before, but mainly because I'm partial to boost. 

As far as icons: I believe I made that `PostListingList` just ignore your `showAvatar` setting in order to make it more compact, but I'd be totally fine with that being added back in.

As far as the other two changes: 

- Removing the creator name: I agree it looks cleaner on mobile, but a lot of people would consider that vital info.
- Moving the community above the title:
  - This could be solved with a setting that applies to all post views?
  - Or just creating a new post listing view mode.

Its up to you whether you want to:

- Create a new post view mode, and add your new one. (I like this one the best I think).
- Or create settings that apply to all post view modes.

PR #1466, opened 2024-04-02
Fixing a few other top app bar title sizes.

dessalines:
This fixes a few other top app bar titles by using `titleSmall`


MV-GH:
I believe it should be `titleMedium` instead. Previously it was was titleLarge which was 16sp and titleMedium now is 16sp and titleSmall is 14sp

MV-GH (approved):

PR #1473, opened 2024-04-07
Fix Saved comments header does nothing

### LINKED ISSUE #1469: Saved comments, empty space in header does nothing
### Jerboa Version

0.0.65

### Android Version + Phone

n/a

### Describe The Bug

Header in the saved comments, empty space does nothing. Expect it to open the post. This would be inline with other headers.

### To Reproduce

go to save comment
press on the header empty space (title does behave as expected)

https://github.com/MV-GH/LemmyBackwardsCompatibleAPI/assets/67873169/62d7f6d4-be1e-4f71-94dd-6954e090379d

_No response_

MV-GH:
Fix #1469 

Clickable should always be before padding.  Else it won't apply the clickable to the padding. And we will have tiny gaps that aren't clickable

Before

image
after

image

I have also looked into setting short/long click on base comment column instead individually on header context, header, body and footer. But seems to have some different behaviour. Which I will need to delve deeper into another time.


dessalines (approved):
Makes sense, thx.

PR #1477, opened 2024-04-09
Add tr locales

mikropsoft:


MV-GH:
Thanks for the translation, you will also need to update `locales_config.xml` see CONTRIBUTING.md

MV-GH (changes requested):
Missing local entry

PR #1483, opened 2024-04-24
Creates a screen to allow the unblocking of users, communities and instances

rodrigo-fm:
This is my biggest PR on jerboa so far, users now have a dedicated screen to unblock anything they want, here's a video demonstrating:

https://github.com/LemmyNet/jerboa/assets/56604262/328bffc5-111d-44f9-9fce-3230e5f29c05

P.S: I also updated the AGP version


app/build.gradle.kts
```diff
@@ -139,7 +139,7 @@ dependencies {
     implementation("io.coil-kt:coil-svg:2.6.0")
     implementation("io.coil-kt:coil-video:2.6.0")
     // Allows for proper subsampling of large images
-    implementation("me.saket.telephoto:zoomable-image-coil:0.8.0")
+    implementation("me.saket.telephoto:zoomable-image-coil:0.9.0")
```
MV-GH, line 142: There is a 10 release

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,7 +1,7 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 distributionSha256Sum=3e1af3ae886920c3ac87f7a91f816c0c7c436f276a6eefdb3da152100fef72ae
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.4-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.6-bin.zip
```
MV-GH, line 4: U will have to update the checksum above or the CI will fail

MV-GH (changes requested):
Thanks!

I'll give it a proper review later. But can you fix the conflicts. Diff is hard to read atm

That delete icon, is it a M3 icon? I would like to keep the same style across the app

app/src/main/java/com/jerboa/ui/components/blocks/BlocksActivity.kt
```diff
@@ -0,0 +1,260 @@
...
+        Text(name, modifier = Modifier.weight(1f))
+        TextButton(onClick = onUnblock, colors= ButtonDefaults.buttonColors(Color.Transparent)) {
+            when (apiState) {
+                ApiState.Loading -> CircularProgressIndicator(
+                    modifier = Modifier.size(UNBLOCK_BUTTON_SIZE.dp),
+                    color = Color.Gray,
+                )
+                is ApiState.Success -> onSuccessfulUnblock()
+                else -> Text(
+                    text = "X",
```
MV-GH, line 88: I would use an icon

app/src/main/java/com/jerboa/ui/components/blocks/BlocksActivity.kt
```diff
@@ -0,0 +1,260 @@
...
+                )
+            }
+        }
+    }
+}
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun BlocksActivity(
+    siteViewModel: SiteViewModel,
```
MV-GH, line 103: I would make a viewModel for this screen. Keep the UI and data logic separate.

MV-GH (changes requested):

app/src/main/java/com/jerboa/ui/components/blocks/BlocksActivity.kt
```diff
@@ -0,0 +1,260 @@
...
+                        item { Spacer(modifier = Modifier.padding(vertical = 8.dp)) }
+                        item { Title(stringResource(R.string.blocked_communities)) }
+                        item { Spacer(modifier = Modifier.padding(vertical = 4.dp)) }
+
+                        if (communityBlocks.isNullOrEmpty()) item {
+                            Text(stringResource(R.string.you_have_no_blocked_communities))
+                        } else {
+                            items(
+                                communityBlocks,
+                                key = { ++key },
```
MV-GH, line 185: Just use null, by default it will then use the index

gradle/wrapper/gradle-wrapper.properties
rodrigo-fm: I downgraded for now since I started to have build errors, I'll have to solve it later.

dessalines (commented):
As @MV-GH mentioned this needs to be reworked to follow the conventions of all the other examples: ViewModel, Activity, and read-only composables in their own file. IMO it makes sense to add the block responses to the siteViewModel, that makes sense, so that they can update those values without having to refetch the site.

Start with an existing somewhat similar example like `ReportsActivity.kt`, copy-paste the entire folder, then change as necessary.

app/src/main/java/com/jerboa/ui/components/blocks/BlocksActivity.kt
dessalines: Absolutely, but I'd also add the block actions to the SiteViewModel rather than creating a new one, so that it can update the blocks. We may also need to change / fix some other block actions, such as `CommunityViewModel.blockCommunity`, so that it can stay in sync with `SiteViewModel.my_user.community_blocks`... etc.

app/src/main/java/com/jerboa/Utils.kt
```diff
@@ -1046,13 +1046,56 @@ fun showBlockPersonToast(
...
+            Toast.makeText(
+                ctx,
+                ctx.getText(R.string.user_block_toast_failure),
+                Toast.LENGTH_SHORT,
+            ).show()
+        }
+    }
+}
+
+fun showBlockInstanceToast(
```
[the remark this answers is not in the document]
MV-GH: I rather we put this in  `feat/Block.kt`, I have been slowly moving stuff out of `Utils` its been used as a dump.

app/src/main/java/com/jerboa/ui/components/blocks/BlocksActivity.kt
```diff
@@ -0,0 +1,275 @@
...
+import it.vercruysse.lemmyapi.v0x19.datatypes.BlockCommunityResponse
+import it.vercruysse.lemmyapi.v0x19.datatypes.BlockInstance
+import it.vercruysse.lemmyapi.v0x19.datatypes.BlockInstanceResponse
+import it.vercruysse.lemmyapi.v0x19.datatypes.BlockPerson
+import it.vercruysse.lemmyapi.v0x19.datatypes.BlockPersonResponse
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+
+private const val UNBLOCK_BUTTON_SIZE = 18
```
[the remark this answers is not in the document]
MV-GH: For buttons we do hardcode this.

But use a preset from here: 
https://github.com/LemmyNet/jerboa/blob/main/app/src/main/java/com/jerboa/ui/theme/Sizes.kt

app/src/main/java/com/jerboa/ui/components/blocks/BlocksActivity.kt
```diff
@@ -0,0 +1,275 @@
...
+    onSuccessfulUnblock: () -> Unit,
+) {
+    Row(verticalAlignment = Alignment.CenterVertically) {
+        icon?.let {
+            CircularIcon(
+                icon = it,
+                contentDescription = "",
+                size = 26.dp,
+            )
+            Spacer(modifier = Modifier.padding(horizontal = 4.dp))
```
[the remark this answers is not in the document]
MV-GH: Use the `SMALL_PADDING` preset constant

https://github.com/LemmyNet/jerboa/blob/main/app/src/main/java/com/jerboa/ui/theme/Sizes.kt

app/src/main/java/com/jerboa/ui/components/blocks/BlocksActivity.kt
```diff
@@ -0,0 +1,275 @@
...
+            when (apiState) {
+                ApiState.Loading -> CircularProgressIndicator(
+                    modifier = Modifier.size(UNBLOCK_BUTTON_SIZE.dp),
+                    color = Color.Gray,
+                )
+                is ApiState.Success -> onSuccessfulUnblock()
+                else -> Icon(
+                    imageVector = Icons.Rounded.Close,
+                    modifier = Modifier.size(UNBLOCK_BUTTON_SIZE.dp),
+                    tint = Color.Red,
```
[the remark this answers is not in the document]
rodrigo-fm: I used ``Material.colorScheme.error`` as you suggested, but the "X" icon is blue instead of red, are my app settings wrong?

image

app/src/main/java/com/jerboa/ui/components/blocks/BlocksActivity.kt
```diff
@@ -0,0 +1,275 @@
...
+    name: String,
+    onUnblock: () -> Unit,
+    onSuccessfulUnblock: () -> Unit,
+) {
+    Row(verticalAlignment = Alignment.CenterVertically) {
+        icon?.let {
+            CircularIcon(
+                icon = it,
+                contentDescription = "",
+                size = 26.dp,
```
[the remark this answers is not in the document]
rodrigo-fm: I created a new size called SMALL_ICON_SIZE = 26.dp, but the naming doesn't seem intuitive since ICON_SIZE is equal to 20.dp. Do you think I should switch their names?

app/src/main/java/com/jerboa/ui/components/blocks/BlocksActivity.kt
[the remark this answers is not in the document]
MV-GH: No it depends on your selected theme. Light + blue theme will have that effect l

app/src/main/java/com/jerboa/ui/components/blocks/BlocksActivity.kt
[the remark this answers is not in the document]
MV-GH: Is it not possible to use the existing one of 28 dp?

We do all these presets for conformity. So that every page has the same colors, sizes, structure, ..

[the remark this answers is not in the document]
rodrigo-fm: @MV-GH I tried to search for a constant with 28.dp, couldn't find it

image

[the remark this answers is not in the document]
MV-GH: Ah mb, sure you can switch the naming

app/src/main/java/com/jerboa/ui/components/blocks/BlocksActivity.kt
rodrigo-fm: @dessalines I don't think I understand the idea of adding the block actions to the SiteViewModel. I had to create an apiState for each element on the list to avoid the same state being displayed to every element (this would cause the loading state to be displayed on each element, for instance).

MV-GH: Add what you can to the blockviewmodel. I don't understand  what you would need a apistate per block

rodrigo-fm: @MV-GH you suggest a list of ApiStates inside the BlockViewModel? That's interesting

MV-GH: Well or a BlockListViewModel and have a viewmodel per entry.

MV-GH: See these guidelines on why we do this.


https://developer.android.com/develop/ui/compose/architecture#udf
https://android.googlesource.com/platform/frameworks/support/+/androidx-main/compose/docs/compose-api-guidelines.md#prefer-stateless-and-controlled-composable-functions

dessalines:
Sry, been really busy lately. I'll try to get to this sometime this week.

app/src/main/java/com/jerboa/ui/components/blocks/BlockedPersonViewModel.kt
```diff
@@ -0,0 +1,52 @@
...
+import com.jerboa.api.toApiState
+import com.jerboa.feat.showBlockPersonToast
+import it.vercruysse.lemmyapi.v0x19.datatypes.BlockPerson
+import it.vercruysse.lemmyapi.v0x19.datatypes.BlockPersonResponse
+import it.vercruysse.lemmyapi.v0x19.datatypes.PersonId
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+
+class BlockedPersonViewModel(
```
MV-GH, line 22: Didn't have to make a separate Blocked Person/Community/Instance  viewmodel

app/src/main/java/com/jerboa/ui/components/blocks/BlocksActivity.kt
```diff
@@ -0,0 +1,242 @@
...
+                ApiState.Empty -> ApiEmptyText()
+                ApiState.Loading -> LoadingBar(padding)
+
+                is ApiState.Failure -> ApiErrorText(siteRes.msg)
+
+                is ApiState.Success -> {
+                    val personBlocks = siteRes.data.my_user?.person_blocks?.toMutableStateList()
+                    val communityBlocks =
+                        siteRes.data.my_user?.community_blocks?.toMutableStateList()
+                    val instanceBlocks = siteRes.data.my_user?.instance_blocks?.toMutableStateList()
```
MV-GH, line 131: Needs to be in the viewmodel. 

image

The `onSuccessfulUnblock` needs to be removed from the UI. Instead the ViewModel updates the SnapShotList (Block event). The UI will react to this change and update. *Dont use a list, use mutableStateList and only expose SnapShotList (Its what I am doing with changes to the feed, atm we create a new list there constantly but causes the entire feed to recompose each time)*

MV-GH (changes requested):

app/src/main/java/com/jerboa/ui/components/blocks/BlocksActivity.kt
```diff
@@ -0,0 +1,242 @@
...
+                                    onUnblock = { viewModel.blockCommunity(false, context) },
+                                    onSuccessfulUnblock = {
+                                        communityBlocks.removeIf { it.community.id == blockedCommunity.id }
+                                    },
+                                )
+                            }
+                        }
+
+                        item { Spacer(modifier = Modifier.padding(vertical = MEDIUM_PADDING)) }
+                        item { Title(stringResource(R.string.blocked_instances)) }
```
MV-GH, line 197: If you want https://developer.android.com/develop/ui/compose/lists#sticky-headers

Might look interesting for the subtitles

app/src/main/java/com/jerboa/ui/components/blocks/BlocksActivity.kt
rodrigo-fm: This would be a different viewmodel? The ones I created so far are viewmodels for each list element.

MV-GH: Yes that was what I had in mind. But if you have a solution that works with viewmodel per element you can do that too. But I think it would be a lot simpler if you used just one viewmodel.

rodrigo-fm: Agree, a BlocksViewModel sounda good. I'll see later what I can do about the unique viewmodel that represents a list item

app/src/main/java/com/jerboa/ui/components/blocks/BlockedCommunityViewModel.kt
dessalines, line 1: None of our other components have viewmodels in the same folder as the UI elements / Activities. There is a dedicated models folder for this.

app/src/main/java/com/jerboa/ui/components/blocks/BlockedPersonViewModel.kt
dessalines: Agree, you can just use a single blockviewmodel, and have the functions pass in the id, and the block.

app/src/main/java/com/jerboa/ui/components/blocks/BlocksActivity.kt
```diff
@@ -275,3 +228,15 @@ fun BlocksActivity(
         },
     )
 }
+
+@Composable
```
dessalines, line 232: These aren't necessary. Look at how all other viewmodels are instantied within activities.

app/src/main/java/com/jerboa/ui/components/blocks/BlockedCommunityViewModel.kt
```diff
@@ -0,0 +1,52 @@
...
+import com.jerboa.feat.showBlockCommunityToast
+import it.vercruysse.lemmyapi.v0x19.datatypes.BlockCommunity
+import it.vercruysse.lemmyapi.v0x19.datatypes.BlockCommunityResponse
+import it.vercruysse.lemmyapi.v0x19.datatypes.CommunityId
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+
+class BlockedCommunityViewModel(
+    private val communityId: CommunityId,
```
dessalines, line 23: You only need these input params for the viewmodels, if an `init` function is required (for example, to initially fetch a person's profile). Those aren't necessary here. Just pass in the IDs to the block functions, and remove the factories.

app/src/main/java/com/jerboa/ui/components/blocks/BlocksActivity.kt
```diff
@@ -49,18 +40,13 @@ import com.jerboa.ui.components.common.LoadingBar
...
-import it.vercruysse.lemmyapi.v0x19.datatypes.BlockPersonResponse
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.launch
-import kotlinx.coroutines.withContext
+import it.vercruysse.lemmyapi.v0x19.datatypes.CommunityId
+import it.vercruysse.lemmyapi.v0x19.datatypes.Instance
+import it.vercruysse.lemmyapi.v0x19.datatypes.PersonId
 
 @Composable
 private fun BlockedElementListItem(
```
dessalines, line 52: Please move this to its own `Blocks.kt` file. Our standard has been *ideally* for the Activity to only contain the top-level screen / activity, which works with the viewmodels and does all the actions. Then another kotlin file for all the inner composables. 

Look again at a few existing folders here.

dessalines (changes requested):

rodrigo-fm:
Hey, sorry for being away for weeks, had other things to take care of. I'll be marking the comments addressed by the commits as resolved.

MV-GH:
No problem, but in the meantime I was working on this feature too. I'll post my PR soon.

dessalines:
Closing in favor of #1545

PR #1485, opened 2024-05-01
Use Compose BOM

### LINKED ISSUE #1484: Use compose BOM
### Pre-Flight checklist

https://developer.android.com/develop/ui/compose/setup#using-the-bom

MV-GH:
Fixes #1484

You can find the BOM here
https://developer.android.com/develop/ui/compose/bom/bom-mapping

Also bumps gradle to 8.7


dessalines (approved):
Nice, thx. I'll have to add this to my other projects.

PR #1502, opened 2024-05-15
Set specific versions for runtime-livedata.

dessalines:


MV-GH:
Why was this done? I had to remove them
had this issue
https://stackoverflow.com/questions/78360992/java-lang-illegalstateexception-compositionlocal-locallifecycleowner-not-presen

MV-GH:
Actually even with that change I am still experiencing it in different parts of the app. Opening the imageviewer will result in crash

`  java.lang.IllegalStateException: CompositionLocal LocalLifecycleOwner not present`

We can't release until we fix this.

dessalines:
Seems like using https://github.com/chrisbanes/compose-bom is the best short-term fix, so we don't have to make any code changes.

Issue #1514, opened 2024-05-20
Several elements aren't labelled clearly for screen readers.


dessalines:
### Pre-Flight checklist

- [X] Did you check to see if this issue already exists?
- [X] This is a single feature request. (Do not put multiple feature requests in one issue)
- [X] This is not a question or discussion. (Use https://lemmy.ml/c/jerboa for that)

### Describe The Feature Request Below

Several elements are not labeled in a way that describes their topic or purpose.  

- The amount of upvotes is not labeled as such, a screen reader will just read this number without saying what this number means. 
- The upvote buttons are read as ‘downvoted’. 
- The community icon states “Go to community” but does not state which community. 
- Whenever a post is selected (see image), the first thing that is read is the amount of upvotes and the time it was posted, this does not describe the post. Only after this information is read the title of the posts is stated. The screen then says: “Double tap to activate”, this does not convey what will happen: when the user double-taps the screen they are referred to the comment-section of this post.

PR #1516, opened 2024-05-20
Fix content description for upvotes

MV-GH:
For #1514

> The upvote buttons are read as ‘downvoted’.

I had noticed this before but I had fixed it wrongly.


dessalines (approved):
Nice, thx.

PR #1523, opened 2024-05-22
chore(deps): update kotlin monorepo to v2 (major)

MV-GH:
The 2.0.0 kotlin compiler has been released already. It just moved now to a plugin

https://developer.android.com/jetpack/androidx/releases/compose-kotlin

### DIFF
--- a/app/benchmark-rules.pro
+++ b/app/benchmark-rules.pro
@@ -4,3 +4,4 @@
 # wrong symbols would be generated. The generated Baseline Profile will be properly applied when generated
 # without obfuscation and your app is being obfuscated.
 -dontobfuscate
+-dontwarn org.slf4j.impl.StaticLoggerBinder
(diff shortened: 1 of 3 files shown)


dessalines (approved):
Nice, thx. How do you learn about these gradle changes for compose btw?

dessalines (approved):

MV-GH:
Mostly from the noise in the kotlin compose slack channel

Also beware that kotlin compiler 2 brings some changes to recompositions. Strong skipping is now enabled by default. This is good but might cause some issues if we have something wrongly configured. So I would at least wait a bit to do a release.

https://medium.com/androiddevelopers/jetpack-compose-strong-skipping-mode-explained-cbdb2aa4b900

Important changes
>   - Composables with unstable parameters can be skipped.
>   - Unstable parameters are compared for equality via instance equality (===)
>  - Stable parameters continue to be compared for equality with Object.equals()
>   - All lambdas in composable functions are automatically remembered. This means you will no longer have to wrap lambdas in remember to ensure a composable that uses a lambda, skips. 
 
 That last one is very good, many of my recompositions changes were that, (my PR is getting pretty far but will have to change now)

dessalines:
Sweet.

PR #1527, opened 2024-05-27
Performance refactors for Posts feed

MV-GH:
Refactor to use `SnapshotStateList` instead of `List`

Before what we did was create a new list each time we change something in that list. Since compose checks on equality the list is now different and fully recomposes the entire list. Instead use `SnapshotStateList` which emits events for only the changes made to the list.

Many changes have made to the postlisting so that when a vote changes it doesn't recompose the entire postlisting but just the votes. This was only successful for postlisting as the card modes use postview to deeply into the "onAction" lambdas which trigger recompose as the postview has different equality.

It no longer recomposes the feed when adding/updating postviews

~~The collapsible header also causes recompositions as it changes the innerpadding which is applied to the lazy column. Not sure if anything can be done for that. (But only when it moves, so when u scroll down once its up, no recompositions happen)~~ (This has been solved)

I am also going to refactor the comments. (I'll do that in a separate PR)


app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
```diff
@@ -736,6 +743,33 @@ fun PostFooterLine(
     }
 }
 
+@Composable
```
dessalines, line 746: 👍

compose_compiler_config.conf
```diff
@@ -7,4 +7,7 @@ it.vercruysse.lemmyapi.*.datatypes.**
 arrow.core.Either
 
 // Consider List, etc stable, see https://github.com/LemmyNet/jerboa/issues/1332
-kotlin.collections.*
\ No newline at end of file
+kotlin.collections.*
+
+// todo CHEck if still needed PostOptionsDropdown
```
dessalines, line 12: typo, also all caps TODO

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
```diff
@@ -128,7 +128,10 @@ import kotlinx.coroutines.CoroutineScope
 
 @Composable
 fun PostHeaderLine(
-    postView: PostView,
+    community: Community,
```
dessalines, line 131: Whats the reason to break out postView? I couldn't find why this was necessary below.

app/src/main/java/com/jerboa/feed/PaginationController.kt
```diff
@@ -0,0 +1,18 @@
+package com.jerboa.feed
+
+import it.vercruysse.lemmyapi.v0x19.datatypes.PaginationCursor
+
+class PaginationController(
+    var page: Long = 1,
```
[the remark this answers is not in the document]
MV-GH: When we stop supporting 0.18. We can remove it but atm we do still support it. Beehaw is still on 0.18. And it doesn't add any complexity to keep supporting it.

app/src/main/java/com/jerboa/model/PostsViewModel.kt
```diff
@@ -15,38 +15,37 @@ import com.jerboa.api.toApiState
...
         private set
     var sortType by mutableStateOf(SortType.Active)
         private set
     var listingType by mutableStateOf(ListingType.Local)
         private set
 
+    private val pageController = PaginationController()
+    private val postController = PostController()
+
+    private val postsList = postController.feed
```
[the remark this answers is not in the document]
MV-GH: > Doesn't this need to be public to get the benefits of the snapshotstate list

`SnapshotStateList` implements List. Its type does not need to be explicitly `SnapshotStateList`

To benefit from it (This can be verified by observing the the recompositions. Else the post simply wouldnt update since the list did not change. And since we declared it stable. Compose wouldn't see any change and thus not recompose but since its actually a SnapShotStateList it emits a event just for that item and the post does recomposition)

> It seems safer to keep one copy of this list, rather than two.

I'll remove postsList and update PostController to only expose List

I don't want it to be modified other than by the Controller itself.

app/src/main/java/com/jerboa/ui/components/home/HomeActivity.kt
```diff
@@ -313,20 +322,22 @@ fun MainPostListingsContent(
...
-                    scope,
-                    siteViewModel,
-                ) {
-                    homeViewModel.savePost(
-                        SavePost(
-                            post_id = postView.post.id,
-                            save = !postView.saved,
-                        ),
-                    )
+            onSaveClick = remember {
```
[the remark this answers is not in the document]
MV-GH: These remembers can indeed be removed I forgot to do so when I enabled Strong Skipping. But before these were required to avoid being constantly recomposed.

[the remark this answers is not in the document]
MV-GH: Also these function don't need to be keyed with postview. They are pure, and receive postview as argument thus do not need to be recreated when the postview changes. It will receive the new postview as argument.

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
MV-GH: We used pass the entire postview into this composable but we use only select parts from it.

For example we have `pv1 = PostView(...)`, we update our vote on it `pv2 = pv1.copy(my_vote=1)` 
Now `pv1 ~= pv2` This composable now recomposes as they are not equal (Stable are compared on equality)
But this Composable doesn't use `my_vote` But it still has to unnecessary recompose  because it arguments are different. Now Since `Post`, `Community` etc stayed the same `pv1.Post == pv2.Post` This composable arguments did not change and thus didn't need to recompose.

Btw Strong skipping made it so that unstable are compared on reference. If their ref is the same they won't recompose. Before all unstable would always recompose. This also means for example that list (which is still considered unstable) will no longer recompose. (We have marked it as stable, because we treat as such)  

If you watch the recompositions (in the layout inspector) for postlisting viewmode then you will see that only the buttons recompose when you upvote or downvote. As they are the only composables that use `my_vote`.

app/src/main/java/com/jerboa/feed/FeedController.kt
```diff
@@ -0,0 +1,76 @@
+package com.jerboa.feed
+
+import android.util.Log
+import androidx.compose.runtime.mutableStateListOf
+import androidx.compose.runtime.snapshots.SnapshotStateList
+
+open class FeedController<T> {
```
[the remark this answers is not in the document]
MV-GH: I was thinking that too but all our "update" methods must use this safe update method. I do not want to be able accidentally use the builtin update that will throw exceptions for items that it doesn't find. All the exposed methods currently are safe, none will throw exceptions. And I do not allow access to SnapShotState directly where one accidentally could do `snap[i] =`

The few extra copies are worth it imo.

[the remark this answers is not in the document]
dessalines: Would it be easier to just use inheritance with an override, [described here](https://kotlinlang.org/docs/inheritance.html).

IE something like 

```kotlin
open class JerboaSnapshotStateList(): SnapshotStateList {
  final override fun update(...) {
```

We can't be the only ones dealing with this problem also, I'd be surprised if there's not already a library that handles this case.

If those aren't possible, then its no problem, just seems like a lot of unecessary code.

app/src/main/java/com/jerboa/model/PostsViewModel.kt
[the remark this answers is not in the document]
dessalines: Sounds good.

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
dessalines: Ahh mmk, makes sense.

app/src/main/java/com/jerboa/model/PostsViewModel.kt
```diff
@@ -18,40 +18,37 @@ import com.jerboa.api.toApiState
...
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.launch
 
 open class PostsViewModel(protected val accountRepository: AccountRepository) : ViewModel() {
-    var postsRes: ApiState<GetPostsResponse> by mutableStateOf(ApiState.Empty)
-        private set
-    private var page by mutableLongStateOf(1)
-    protected var pageCursor: PaginationCursor? by mutableStateOf(null)
+    val lazyListState = LazyListState()
+    var postsRes: ApiState<List<PostView>> by mutableStateOf(ApiState.Empty)
```
dessalines, line 42: This one still needs to be got rid of then I think. Sometimes below you are updating the postController, and sometimes the postsRes. Lets try to keep only one list.

app/src/main/java/com/jerboa/ui/components/common/TextBadge.kt
```diff
@@ -52,13 +53,13 @@ fun TextBadge(
...
  * Displays activitypub items (communities, users), with a smaller @instance shown
  */
-@OptIn(ExperimentalFoundationApi::class)
 @Composable
 fun ItemAndInstanceTitle(
     modifier: Modifier = Modifier,
     title: String,
     actorId: String?,
     local: Boolean,
+    onClick: (() -> Unit)?,
```
dessalines, line 62: There's no need to make these optional IMO. If its called as `onClick = {}`, it at least forces us to look at each case where its called.

app/src/main/java/com/jerboa/feed/FeedController.kt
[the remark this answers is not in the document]
MV-GH: >  just seems like a lot of unecessary code.

Its four lines
```
    fun add(item: T) = items.add(item)
    fun remove(item: T) = items.remove(item)
    fun clear() = items.clear()
    fun addAll(newItems: List<T>) = items.addAll(newItems)
```

I didn't do inheritance because then I would expose a lot different methods that I don't want either. Plus then they would still be able to `snap[]` in a unsafe matter.

app/src/main/java/com/jerboa/ui/components/home/HomeActivity.kt
```diff
@@ -228,70 +235,72 @@ fun MainPostListingsContent(
...
                 .zIndex(100f),
         )
         // Can't be in ApiState.Loading, because of infinite scrolling
         if (homeViewModel.postsRes.isLoading()) {
-            LoadingBar(padding = padding)
+            LoadingBar()
         }
 
-        val posts =
+        val posts: List<PostView> = remember(homeViewModel.postsRes) {
```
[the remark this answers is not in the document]
dessalines: bump on this one too.

app/src/main/java/com/jerboa/ui/components/common/TextBadge.kt
MV-GH: No it is important this is optional, We can't add a clickable modifier for these. Else it wouldn't propagate the click to he parent clickable modifier. This empty onClick handler would capture the click

app/src/main/java/com/jerboa/feed/FeedController.kt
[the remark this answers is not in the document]
dessalines: mmmk, that's no problem then.

app/src/main/java/com/jerboa/model/PostsViewModel.kt
MV-GH: We need to update both PostRes and the PostController. PostRes still controls the api states, ex: loading/appending show loading bar. PostController controls the list in the postRes

app/src/main/java/com/jerboa/ui/components/common/TextBadge.kt
dessalines: Ah I see.

app/src/main/java/com/jerboa/model/PostsViewModel.kt
MV-GH: We store one list, the other is just a reference to it.

Doing `ApiState<SnapshotStateList<PostView>>` would me we expose the `SnapshotStateList` everywhere so everywhere we could update the apistate from. 

`List<PostView>` is immutable. So can't be manipulated.

app/src/main/java/com/jerboa/ui/components/home/HomeActivity.kt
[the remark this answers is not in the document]
MV-GH: I'll remove the remember here but it will generate the same one anyway

I could do ApiState<Unit> and expose the posts as `List` then I would be able to directly use it.
I am neutral about that, but I didn't think you would prefer it.

app/src/main/java/com/jerboa/model/PostsViewModel.kt
dessalines: L75 has `postController.addAll(it.posts)` , adding all the items individually to a separate list.

L96-97 Also add items to both lists, twice.

dessalines: I think this was the point of the `private set` also, to ensure that its source data can only change in the model.

I need to read more about snapshotstatelist, because this is a problem that should be able to be solved without keeping two lists.

MV-GH: L75 is the initPosts, on startup we load the first page. So we add all the items.

On L96-99 is appendPosts (load next page). so we add all the new posts

Ig they are kinda the same and I could just use appendPosts but this is how we do it everywhere. (init then append for next pages)

What am I missing?

MV-GH: `private set` doesn't prevent us from manipulating the inner state of apiState just prevent us from reassigning it to another.

dessalines: The two lists are the `SnapshotStateList` in FeedController, and the `List` wrapped in the ApiState in `postsRes`.

> Doing ApiState<SnapshotStateList<PostView>> would mean we expose the SnapshotStateList everywhere, so everywhere we could update the posts from.

I spose this is all a workaround due to the fact that we can't make `SnapshotStateList` externally safe, although to me it'd be logically cleaner to just do `ApiState<SnapshotStateList<PostView>>`.

Did you test to make sure things like `postController.findAndUpdateCreator(personView.person)` for banning / unbanning work correctly? To me it seems like these aren't going to work correctly, since externally only `postsRes` is exposed, not your new snapshotstatelist.

Also for example on L97, you're doing `postsRes = ApiState.Success(postController.feed)`, which to me seems like its going to replace the list anyway, so you'd lose all the benefits of the snapshotstatelist.

MV-GH: There is only one list, `SnapshotStateList ` in FeedController. I expose this as `List` in the feed controller. All operations are done on this (SnapShotState)List. During appending I change the `ApiState` to  appending ` ApiState.Appending(oldRes.data)` So that the loading bar, etc trigggers but its still the same list.

then when I receive the newPosts from the API. I add the posts to the `SnapShotStateList`
`postController.addAll(newRes.data.posts)` then I change ApiState back to Success ` postsRes = ApiState.Success(postController.feed)` so that the loading bar goes away.

I could do `postsRes = ApiState.Success(oldRes.data)` same thing. There is only ever one list and all contain a reference to it.


> Did you test to make sure things like (...)

Yes it works see vid below. Only `postRes` is indeed exposed and `postRes` exposes that `SnapShotStateList` as `List`

dessalines: Alright, these cases are a bit confusing to me, but that's my bad.

dessalines (approved):

PR #1528, opened 2024-05-30
Adding alt_text and custom_thumbnail to post form. Fixes #1513

### LINKED ISSUE #1513: Support alt-text field for post images.
### Pre-Flight checklist

All images that aren’t decorative need a text alternative, so that people using screen readers can hear what the image or icon means. This might be hard to implement as there is much user generated content. Perhaps you could provide an option to your users to add text descriptions to their uploads, and encourage them to do so.

dessalines:
This adds the recently added `alt_text` and `custom_thumbnail` fields to creating and editing posts: 

- The `alt_text` field is only shown after you've added a url that's an image, matching lemmy-ui's behavior.
- The `postView.post.alt_text` is now used as the `contentDescription` fields for pictrs images.
- There might be some issues with `custom_thumbnail`, but that's a back-end problem still.

Example:

https://github.com/LemmyNet/jerboa/assets/930215/06b065cc-207c-48ba-baa8-dbe599f7792b


MV-GH (approved):

app/src/main/java/com/jerboa/ui/components/post/composables/PostComposables.kt
```diff
@@ -141,9 +146,36 @@ fun CreateEditPostBody(
             PictrsUrlImage(
                 url = url,
                 blur = false,
+                contentDescription = altText,
+            )
+
+            /**
+             * Alt text field. Only show if its a url image.
+             */
+            OutlinedTextField(
```
MV-GH, line 155: Alt text is probably a recent feature? You might wanna add a feature flag check to hide these for older Lemmy Versions

dessalines: Ah forgot about that, good call.

dessalines: K I added them for hiding posts, alt_text, and custom_thumbnails, but all are using the `FF.hidePost` since `FF.altText` doesn't exist (Its nbd to me).

MV-GH: Yeah I kinda forgot about altText when I generated the Types. I'll add one but yeah it doesn't really matter

dessalines: Cool, ya its not vital for this anyway.

PR #1529, opened 2024-05-31
Adding LocalUserVoteDisplayMode for saving, and displaying votes.

### LINKED ISSUE #1400: Add a `Full` vote display mode? And what should be the default?

dessalines:
- This uses the newly added field to LocalUserView above, to display votes according to your preferences.
- Its also backward compatible, so if that setting is missing, it'll use Upvote + Downvote by default.
- Fixes #1400 

Samples:

Screenshot_20240530_203452_AVNC_1


MV-GH:
I should have time Wednesday to review this

app/src/main/java/com/jerboa/ui/components/settings/account/AccountSettings.kt
```diff
@@ -316,6 +322,24 @@ fun SettingsForm(
                     Text(text = stringResource(R.string.account_settings_show_scores))
                 },
             )
+            SwitchPreference(
```
MV-GH, line 325: These settings are only added with 0.19.4? They should be behind a FeatureFlag?

MV-GH (changes requested):

app/src/main/java/com/jerboa/model/SiteViewModel.kt
```diff
@@ -218,23 +224,13 @@ class SiteViewModel(private val accountRepository: AccountRepository) : ViewMode
...
-                res.data.my_user?.let { mui ->
-                    if (mui.local_user_view.local_user.show_scores) {
-                        defaultMode
-                    } else {
-                        VoteDisplayMode.HideAll
-                    }
-                } ?: run {
-                    defaultMode
-                }
+                res.data.my_user?.local_user_view?.local_user_vote_display_mode ?: LocalUserVoteDisplayMode.default()
```
dessalines, line 231: This one should be okay, because the default in your API is null.

dessalines (commented):
Good call, I updated.

app/src/main/java/com/jerboa/ui/components/settings/account/AccountSettings.kt
dessalines: Fixed. The other ones should be okay, because you set it by default to null in your API addition.

MV-GH (approved):

PR #1530, opened 2024-05-31
Use chris banes compose-bom.

dessalines:
context: https://github.com/LemmyNet/jerboa/pull/1502#issuecomment-2137935525

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -110,7 +110,18 @@ composeCompiler {
 }
 
 dependencies {
-    implementation("androidx.appcompat:appcompat:1.6.1")
+    // Unfortunately, ui tooling, and the markdown thing, still brings in the other material2 dependencies
+    // This is the "official" composeBom, but it breaks the imageviewer until 1.7 is released. See:
+    // https://github.com/LemmyNet/jerboa/pull/1502#issuecomment-2137935525
+    // val composeBom = platform("androidx.compose:compose-bom:2024.05.00")
+
+    val composeBom = platform("dev.chrisbanes.compose:compose-bom:2024.05.00-alpha03")
+    implementation(composeBom)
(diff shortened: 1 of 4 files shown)


app/build.gradle.kts
```diff
@@ -110,7 +110,18 @@ composeCompiler {
...
 dependencies {
-    implementation("androidx.appcompat:appcompat:1.6.1")
+    // Unfortunately, ui tooling, and the markdown thing, still brings in the other material2 dependencies
+    // This is the "official" composeBom, but it breaks the imageviewer until 1.7 is released. See:
+    // https://github.com/LemmyNet/jerboa/pull/1502#issuecomment-2137935525
+    // val composeBom = platform("androidx.compose:compose-bom:2024.05.00")
+
+    val composeBom = platform("dev.chrisbanes.compose:compose-bom:2024.05.00-alpha03")
+    implementation(composeBom)
+    implementation("androidx.activity:activity-compose")
```
dessalines, line 120: I'm assuming only the androidx. namespace things can use this composeBom. 

I tried to do it with all of them at first, but some still required explicit versioning.

MV-GH: My understanding was that only the libraries listed where were covered
https://developer.android.com/develop/ui/compose/bom/bom-mapping

But that is not case. So I am not sure, you can probably look at the dependencies of the BOM to find out. Which libs are covered.

MV-GH (approved):
I will check one more thing once this is merged

since material3 alpha brings some behavioral changes to ExposedDrownMenu
image

This version also brings improvements to pullToRefresh which is a M3 version of pullRefresh we currently use. This is probably one of the last m1 we use.

dessalines:
According to [their setup instructions](https://developer.android.com/develop/ui/compose/setup#setup-compose), it looks like I got em all.

PR #1534, opened 2024-05-31
Fix #1533 use M3 PullToRefresh

### LINKED ISSUE #1533: Replace M1 pullRefresh with M3 PullToRefresh
### Pre-Flight checklist

https://developer.android.com/reference/kotlin/androidx/compose/material3/pulltorefresh/package-summary

MV-GH:
I like the API of the new one, much simpler

Fix #1533

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/inbox/InboxActivity.kt
+++ b/app/src/main/java/com/jerboa/ui/components/inbox/InboxActivity.kt
@@ -3,8 +3,6 @@ package com.jerboa.ui.components.inbox
 import android.content.Context
 import android.util.Log
 import androidx.annotation.StringRes
-import androidx.compose.foundation.ExperimentalFoundationApi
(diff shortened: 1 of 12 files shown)


app/src/main/java/com/jerboa/ui/components/inbox/InboxActivity.kt
```diff
@@ -690,44 +653,26 @@ fun InboxTabs(
                         }
                     }
 
-                    val loading = inboxViewModel.messagesRes.isLoading()
-                    val refreshing = inboxViewModel.mentionsRes.isRefreshing()
```
MV-GH, line 694: This was incorrect before

dessalines: Nice, thx.

dessalines (approved):
Sooo much cleaner, thx for this.

PR #1537, opened 2024-06-01
Refactor deprecated M1 SwipeToDismiss, use M3 SwipeToDismissBox

### LINKED ISSUE #1479: Replace Swipeable with AnchoredDraggable
### Pre-Flight checklist

This is the last part of the deprecated code, since upgrading to the latest Compose

It is also on the blockers for removing material  #1433

https://developer.android.com/develop/ui/compose/touch-input/pointer-input/migrate-swipeable

MV-GH:
Use M3 replacement.

Whats strange is that they named it "Swipe left" but dragging to right is swiping right? So I have corrected the naming to fit the behaviour.

This was the last accompanist dependency. Also the last M1 component.

Fixes #1479

https://github.com/LemmyNet/jerboa/assets/67873169/0a1e92d2-a1ff-4d0c-b9d7-01a4e519598d

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -132,12 +132,6 @@ dependencies {
     implementation("io.noties.markwon:image-coil:4.6.2")
     implementation("io.noties.markwon:linkify:4.6.2")
 
-    // Accompanist
-    implementation("com.google.accompanist:accompanist-pager:0.34.0")
-    implementation("com.google.accompanist:accompanist-pager-indicators:0.34.0")
-    implementation("com.google.accompanist:accompanist-flowlayout:0.34.0")
-    implementation("com.google.accompanist:accompanist-navigation-animation:0.34.0")
-
     // LiveData
     implementation("androidx.compose.runtime:runtime-livedata")
     implementation("androidx.lifecycle:lifecycle-runtime-compose")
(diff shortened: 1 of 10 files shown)


app/build.gradle.kts
```diff
@@ -132,12 +132,6 @@ dependencies {
     implementation("io.noties.markwon:image-coil:4.6.2")
     implementation("io.noties.markwon:linkify:4.6.2")
 
-    // Accompanist
-    implementation("com.google.accompanist:accompanist-pager:0.34.0")
```
dessalines, line 136: 🥳

app/build.gradle.kts
```diff
@@ -184,7 +178,7 @@ dependencies {
     implementation("androidx.compose.material:material-icons-extended")
 
     implementation("org.ocpsoft.prettytime:prettytime:5.0.8.Final")
-    implementation("androidx.navigation:navigation-compose")
+    implementation("androidx.navigation:navigation-compose:2.7.7")
```
dessalines, line 181: This version should come automatically with bom, no?

MV-GH: It did not for me, must ve come from accompanist

dessalines: Gotcha.

PR #1542, opened 2024-06-09
Reduce recompositions due to TopAppBar

MV-GH:
When you swipe up and down the top app bar moves. This updates the padding value. All composable which take this as param constantly recompose because the padding constantly changes as the header takes up more/less space. We can avoid that by applying it before the main composables. I had already done this for the feed. But we can do this easily already for the other screens.


dessalines (approved):
Hopefully some of the newer jetpack compose tutorials should really tell you about that: to not pass in padding values as it will cause a ton of these recompositions.

PR #1544, opened 2024-06-09
Restore legacy 'Show Scores' option

MV-GH:
The legacy 'Show Scores' option was removed, this restores it.

This also fixes missing scores in listing view mode mode for pre 0.19.4 instances


MV-GH:
This also means show_scores was reset to false for pre 0.19.4 versions. As it used the wrong option. If users saved their settings in the meantime

app/src/main/java/com/jerboa/ui/components/common/TimeAgo.kt
```diff
@@ -144,52 +146,58 @@ fun ScoreAndTime(
             downvotes = instantScores.downvotes,
         )
 
+        // If the show_scores is disabled, and we are the instance is pre 0.19.4, we fallback to legacy behaviour
+        val version = API.getInstanceOrNull()?.version
+        val legacyScoresHidden = version != null && version < Version(0, 19, 4) && !voteDisplayMode.score
```
dessalines, line 151: You can probably just use the `!api.FF.hidePost()` like in the other files, to denote that its before `0.19.4`

app/src/main/java/com/jerboa/ui/components/common/TimeAgo.kt
```diff
@@ -144,52 +146,58 @@ fun ScoreAndTime(
...
-                    Icons.Outlined.ArrowDownward,
-                    stringResource(id = R.string.downvoted),
-                ),
-            )
-        }
-        // Only show this spacer if at least one of the fields is enabled
-        if (voteDisplayMode.score || voteDisplayMode.upvotes || voteDisplayMode.downvotes || voteDisplayMode.upvote_percentage) {
-            DotSpacer(style = MaterialTheme.typography.labelMedium)
+        if (!legacyScoresHidden) {
+            if (voteDisplayMode.score && !hideScore) {
```
dessalines, line 159: You can just move !legacyScoresHidden into this if statement.

app/src/main/java/com/jerboa/model/SiteViewModel.kt
```diff
@@ -228,7 +228,8 @@ class SiteViewModel(private val accountRepository: AccountRepository) : ViewMode
     fun voteDisplayMode(): LocalUserVoteDisplayMode {
         return when (val res = siteRes) {
             is ApiState.Success ->
-                res.data.my_user?.local_user_view?.local_user_vote_display_mode ?: LocalUserVoteDisplayMode.default()
+                res.data.my_user?.local_user_view?.local_user_vote_display_mode
+                    ?: LocalUserVoteDisplayMode.default(res.data.my_user?.local_user_view?.local_user?.show_scores)
```
dessalines, line 232: This works. Eventually that field will go away.

app/src/main/java/com/jerboa/ui/components/common/TimeAgo.kt
MV-GH: Ah yes

app/src/main/java/com/jerboa/model/SiteViewModel.kt
MV-GH: 0.20 In Lemmy, yes. But It will always exist in 0.19. And always in some form in the LemmyApi datatypes. So it can be mapped to from older versions. I'll probably have a legacy namespace for that.

app/src/main/java/com/jerboa/ui/components/common/TimeAgo.kt
MV-GH: I don't think so? It's previous behaviour was to hide all "voting types" , total, upvotes/downvotes. Current is to hide only scores. So to match previous behaviour this is necessary

dessalines (approved):

PR #1545, opened 2024-06-10
Add blocks screen

MV-GH:
Thanks @rodrigo-fm , I have largely based this PR from yours.

https://github.com/LemmyNet/jerboa/assets/67873169/60934659-8541-40ef-b22c-a6773db1a293


rodrigo-fm:
I ran the app and blocked one community and one user. Then I navigated to the BlocksActivity, but I couldn't see the user and the community I blocked, I had to drag the screen to trigger the reload, only then the screen showed the user and the community I blocked.

I think you can run the update event when the user opens the BlocksActivity to avoid the need to drag the screen to update. Other solution is to update the blocked users/community/instance lists when a block occurs. I think I used the first solution when I worked on this.

rodrigo-fm:
Btw thank you very much for continuing the blocks screen work! :)

MV-GH:
I didn't want to do that initially because it would make it load first each time they open the screen. But I guess this a common usecase

app/src/main/java/com/jerboa/ui/components/post/composables/PostOptionsDropdown.kt
```diff
@@ -442,7 +442,7 @@ fun PostOptionsDropdown(
                                         ),
                                     )
                                 withContext(Dispatchers.Main) {
-                                    showBlockCommunityToast(resp, instance, ctx)
+                                    showBlockInstanceToast(resp, instance, ctx)
```
dessalines, line 445: Good catch

app/src/main/java/com/jerboa/ui/components/settings/block/BlocksScreen.kt
```diff
@@ -0,0 +1,234 @@
...
+        verticalAlignment = Alignment.CenterVertically,
+    ) {
+        ItemAndInstanceTitle(
+            title = name,
+            actorId = actor,
+            local = false,
+            onClick = null,
+        )
+
+        IconButton(
```
dessalines, line 198: I like this one a lot, there might be other places we can use this style.

dessalines (approved):
I tested and this works great. 

Lots of thanks to @rodrigo-fm for laying this foundation!

PR #1549, opened 2024-06-12
Increase fontsize votes slightly

MV-GH:
I think the fontsize of  voting types are a bit too small. We can slightly increase it and it will still be capable of showing all the voting types. All though it will be very rare that someone has enabled all of them. I believe its better to cater to the common case which will be showing two voting types.

After: (14 SP)
image

Before: (12 SP)

image


dessalines (approved):
Seems fine to me.

PR #1551, opened 2024-06-13
Improve baseline documentation + update baseline profiles

MV-GH:


CONTRIBUTING.md
```diff
@@ -120,3 +122,21 @@ If you add a migration to the DB, test it with this gradle task
...
+
+You can generate the latest baseline profiles by executing the following gradle task.
+It will generate the baseline profiles and the startup profiles.
+
+The baseline profile code is located in the Benchmarks module.
+
+Be warned that this task will take a long time to run (+30m) and can fail.
+
+You will also need to set the variant of `app` to `benchmarkRelease`,
+have no device connected and it should start a emulator with the correct settings.
```
dessalines, line 136: Can be on the same line.

CONTRIBUTING.md
```diff
@@ -120,3 +122,21 @@ If you add a migration to the DB, test it with this gradle task
 ```shell
 ./gradlew app:connectAndroidTest
 ```
+
+## Updating the Baseline profiles
+
+You can generate the latest baseline profiles by executing the following gradle task.
+It will generate the baseline profiles and the startup profiles.
```
dessalines, line 129: Should be on the same line, unless it's a new paragraph.

MV-GH: The advantages of Markdown is that you can write your sentences line by line but they will be properly merged in the rendered document.

image

dessalines: I've always found that weird bc it makes viewing the file in an editor very different from how it loovs rendered. Up to you I spose.

dessalines (approved):

PR #1555, opened 2024-06-18
Removing renovate schedule.

dessalines:
I have 3 CI runners now, so it seems pointless to wait for updates.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,5 +1,4 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "extends": ["config:recommended"],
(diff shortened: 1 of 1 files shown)


MV-GH (approved):
I thought this was more so that you wouldn't get spammed with updated. Just once a month it bundles everything.

dessalines:
That turned out to be less annoying than I'd thought, since it's only a click of a button to merge them.

MV-GH:
Can you add rule to the changelog so that renovate bump deps is only shown once and not per PR? Else the changelog will pretty much just become version bumps

dessalines:
Yep I think I can do that.

PR #1573, opened 2024-06-27
Update lemmy api to beta

MV-GH:
I use "universal" types now, nolonger I expose a specific version as API. I can completely control its shape. So when we upgrade to Lemmy 0.20. The imports wont have to change, and I ll be able to add custom fields for backwards compability.

The datatypes now also implement serializable and Parcelable.

We can go back to the previous way of storing state on the navigation. Instead of the current json serialization.


dessalines (approved):
Nice, should be much easier to support then.

PR #1578, opened 2024-07-05
Fix multiple spoiler blocks displaying incorrectly (#1577)

### LINKED ISSUE #1577: Closing and opening multiple spoiler tags works incorrectly
### Jerboa Version

0.0.69

### Android Version + Phone

Android 11 / Redmi Note 8T

### Describe The Bug

When a post has multiple spoiler tags, opening them in order will not render correctly. It renders correctly if opened bottom to top.

This is flipped for closing, where closing them in order works, and bottom to top does not.

When I open the first one and then the second one:

When I open the second one and then the first one:

Same happens in the post preview while posting.

As seen in this post: https://lemmy.ml/post/17579992

### To Reproduce

1. Open post or comment that has multiple spoiler tags ([https://lemmy.ml/post/17579992](https://lemmy.ml/post/17579992))
2. Open first spoiler tag
3. Open second spoiler tag
4. Spoiler tags are rendered incorrectly until post is refreshed

_No response_

Tyoda:
- The spans are now overwriting the correct text even if the text has changed since the first parsing
- Fixes #1577

### DIFF
--- a/app/src/main/java/com/jerboa/util/markwon/MarkwonSpoilerPlugin.kt
+++ b/app/src/main/java/com/jerboa/util/markwon/MarkwonSpoilerPlugin.kt
@@ -113,17 +113,19 @@ class MarkwonSpoilerPlugin(
                                 textView.cancelPendingInputEvents()
                                 open = !open
 
+                                val spoilerStartCurrent = spanned.getSpanStart(spoilerTitle)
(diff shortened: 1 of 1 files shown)


MV-GH (approved):
Seems to work well, thanks!

PR #1579, opened 2024-07-08
Fix show back button when clicking on tagged user

Tyoda:
Previously a non-working menu button would show in the top left corner when going to a user profile through a user tag (e.g. @Tyoda@lemm.ee). Now a working go back button is shown instead.


app/src/main/java/com/jerboa/MainActivity.kt
```diff
@@ -362,6 +362,7 @@ class MainActivity : AppCompatActivity() {
                             blurNSFW = appSettings.blurNSFW.toEnum(),
                             showPostLinkPreviews = appSettings.showPostLinkPreviews,
                             drawerState = drawerState,
+                            onBack = appState::popBackStack,
```
MV-GH, line 365: Thanks, but can you remove the default `null` from it. We have noticed these defaults have been footguns. They caused issues like this.

MV-GH (changes requested):

app/src/main/java/com/jerboa/MainActivity.kt
Tyoda: Not sure what you mean by that, sorry. There is no `null` involved.

Tyoda: Unless you mean from the constructor of PersonProfileScreen, but I didn't write that. Could remove it of course.

MV-GH: Yes, thats what I mean

Tyoda: Removed it from both `onBack` and `padding`, and made them null where they were that by default before.

dessalines (approved):

PR #1595, opened 2024-07-15
Fixing build script.

dessalines:


dessalines:
Just making sure this does a clean before building.

PR #1607, opened 2024-07-31
Adding edge to edge support

dessalines:
This was pretty easy to add, and mostly works well.

Some quirks mainly have to do with us not using proper material3 color scheming in a few places, but I'll start on that soon.


MV-GH:
I have attempted this before, and it didn't go well. Did you check this on every Android version, especially the old ones. Had very weird behaviour. see https://github.com/LemmyNet/jerboa/pull/1443

dessalines:
I just tested on my device (android 14). I have a few other apps that are just using it in the [recommended way](https://developer.android.com/develop/ui/views/layout/edge-to-edge), and ppl don't seem to be having issues with it.

I suppose I could try to test a few others out.

MV-GH:
I'll check it this weekend on various verions.

MV-GH:
On android 10 and lower you have these issues

image

image


The bottom app bar color doesn't match
image
image


Top bar icons stay dark
image

So yeah still plenty  of issues

dessalines:
Dang. I'll try to investigate why the bottomappbar is doing that.

dessalines:
Okay I installed an android 10 emulator, and fixed the insets.

I couldn't replicate the icon color issue you're getting tho.

Screenshot_20240805_183512_AVNC

The bottomappbar color I think is unavoidable for lower android version, but its still usable.

app/src/main/java/com/jerboa/ui/components/home/BottomNavScreen.kt
```diff
@@ -294,16 +294,6 @@ fun BottomNavScreen(
                         )
                     }
 
-                    composable(route = NavTab.RegistrationApplications.name) {
```
MV-GH, line 297: Is this intended to be removed?

dessalines: Ya check L287 , its an exact dupe.

dessalines:
Anything else I need to do with this one @MV-GH ?

MV-GH:
I'll test it more this weekend

dessalines:
bump on this one also @MV-GH

MV-GH:
I'll test it rn

MV-GH:
Can't find anything wrong. But users will definitely remark the jumping images

MV-GH (approved):

dessalines:
I'm fairly sure this has to do with the `coil` version bumps, not this. I'll test downgrading those.

MV-GH:
On Android 11

light theme looks like it has the wrong colors for the nagivation bar + statusbar

image

dessalines:
The only thing I can think of is to put version checks on `enableEdgeToEdge()`. But since we're using this in the officially-documented way, we should also probably open up an issue on the google issue tracker.

MV-GH:
> The only thing I can think of is to put version checks on enableEdgeToEdge().

I would only do this as a last resort.

PR #1613, opened 2024-08-06
Deduplicate posts in the feed

MV-GH:
Before the FeedController we had deduplication logic. I presumed this wasn't necessary anymore with pagination cursor. But it seems that is not the case. I have added support for it.


dessalines (approved):
Nice, thx.

PR #1614, opened 2024-08-06
Some material 3 cleanups.

### LINKED ISSUE #1608: Fix material 3 colors to be more standard, stop using `onSurface` for everything.
### Pre-Flight checklist

I built another android app recently, and found that the material color recomendations are much more standardized than when I started building jerboa.

https://m3.material.io/

This is a placeholder issue to remind me to go over those docs again, as well as every component. Specifically:

- We are overriding way too many colors, and often using `onSurface`, where we shouldn't be. 
- I'd still like to keep the layout as *flat* as possible, but there are a few cases where the default card coloring should be used.

dessalines:
- Adding a fade to the post body preview.
- Getting rid of muted, using colorScheme.outlined instead.
- Making label-type text use label rather than body style.
- Getting rid of some unecessary surface and onSurface colors.
- Fixes #1608

Example:

Screenshot_2024-08-06-15-10-57-43_c9298fda48802cf3e2d17cb10a48759f

### DIFF
--- a/app/src/main/java/com/jerboa/Utils.kt
+++ b/app/src/main/java/com/jerboa/Utils.kt
@@ -474,7 +474,7 @@ fun communityNameShown(community: Community): String =
 
 fun hostName(url: String): String? =
     try {
-        URL(url).host
+        URL(url).host.replace("www.", "")
     } catch (e: MalformedURLException) {
         null
     }
--- a/app/src/main/java/com/jerboa/Utils.kt
+++ b/app/src/main/java/com/jerboa/Utils.kt
@@ -474,10 +474,15 @@ fun communityNameShown(community: Community): String =
 
 fun hostName(url: String): String? =
     try {
-        URL(url).host.replace("www.", "")
+        URL(url).host
     } catch (e: MalformedURLException) {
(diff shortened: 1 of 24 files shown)


app/src/main/java/com/jerboa/Utils.kt
```diff
@@ -474,7 +474,7 @@ fun communityNameShown(community: Community): String =
 
 fun hostName(url: String): String? =
     try {
-        URL(url).host
```
MV-GH, line 477: Whats this for?

dessalines: Its to show the hostname of the site for any link posts. The only thing changed here is to remove the `www.`, which a lot of links have and wasted screen space.

MV-GH: Create a different function for that. And only use for the "Shown" purposes. But right now it is also used functionality wise to compare between hosts.

dessalines: Done

MV-GH (approved):

PR #1615, opened 2024-08-07
Move time ago and score to a flowrow layout

dessalines:
A few reasons for doing this:

- The `SpaceBetween` / `SpaceAround` layouts aren't going to work well when we start adding larging screen size support.
- Reading through material design docs, the title should always be on top, with attribution and extra subtitles below.
- Made the community icon on post listings smaller to look better.

What it looks like now:

Screenshot_2024-08-06-21-35-57-01_c9298fda48802cf3e2d17cb10a48759f

Screenshot_2024-08-06-21-16-33-78_c9298fda48802cf3e2d17cb10a48759f

Comes after #1614


dessalines:
I've been using this for a few days, and I'm liking it a lot.

app/src/main/java/com/jerboa/ui/components/post/PostListing.kt
dessalines, line 1: I split the card and list into different files.

app/src/main/java/com/jerboa/ui/components/comment/reply/CommentReply.kt
```diff
@@ -286,3 +286,28 @@ fun PostReply(
         )
     }
 }
+
+@Composable
+fun PostNodeHeader(
```
dessalines, line 291: Moved this as its only used here.

app/src/main/java/com/jerboa/ui/components/common/AppBars.kt
```diff
@@ -273,36 +272,40 @@ fun CommentOrPostNodeHeader(
...
-
-            PersonProfileLink(
-                person = creator,
-                onClick = { onPersonClick(creator.id) },
-                showTags = true,
-                isPostCreator = isPostCreator,
-                isDistinguished = isDistinguished,
-                isCommunityBanned = isCommunityBanned,
-                showAvatar = showAvatar,
+        val centerMod = Modifier.align(Alignment.CenterVertically)
```
dessalines, line 275: This centerMod is necessary on flowRow unfortunately. I wish it could do some cross-axis alignment, but it doesn't seem to work.

app/src/main/java/com/jerboa/ui/components/reports/PostReportItem.kt
```diff
@@ -86,42 +87,43 @@ fun PostReportItem(
...
-                voteDisplayMode = voteDisplayMode,
                 fullBody = false,
             )
         }
 
         //  Title + metadata
         PostBody(
-            post = postView.post,
-            read = postView.read,
+            postView = postView,
```
MV-GH, line 103: These were all "optimizations"

MV-GH: Any field that changes in postView will cause it recompose instead only if the relevant field changed

MV-GH:
I am honestly not such a fan of it. Maybe its worth here using a post view mode for the old view.

With current layout, it "forces" you to read community + votes each time. If you go like vertical: title -> content, you see the community  + votes then the content.

it also seems to be behaving weirdly? With the community + user at the bottom? Or is that intended because I don't like that at all

image

vs old

image

Much better at using the "screen real estate", takes up less vertical space

Can directly see title -> content, want to see, user/community votes look up

app/src/main/java/com/jerboa/ui/components/reports/PostReportItem.kt
dessalines: Erps, my bad, I'll go through and fix some of these.

dessalines:
The community line below the image post is a weirdness I couldn't figure out a great way around. 

Part of the problem is that the title block wraps the post subtitle in a single row, but the community info is below that block. You can test this by looking at any link-type post (or going to small card mode to force images into a smaller thumbnail).

I *could* move the community info into that title block, or move the post subtitle below that title block. Both options aren't ideal, because they end up with wasted screen real estate, but I'll mess around with them.

A few possibilities I could use your input on:

- I'm open to moving the community line above the post subtitle, or moving the subtitle out of that title block.
- I'm open to any variation of ordering the hostname, vote score, and time ago. As long as its a left-starting flow layout.
- I'm open to removing the hostname, although I do think its pretty useful to see that.

MV-GH:
> The smaller community / attribution line, below the title (as per material design standards)

Is that really? If i look up sample M3 apps, they don't follow that at all.


image


Title -> content. Meta data around it

app/src/main/java/com/jerboa/ui/components/reports/PostReportItem.kt
dessalines: Actually I'm not sure if its worth it. A lot of these use like 5+ fields from `PostView`, and if that's the case, it just seems like we should pass in the full thing anyway. 

IE, is there any field on PostView that changes frequently enough to justify separating these out? I don't think there is, as instantScores is the main one, but that's separated.

MV-GH:
Also another thing but not relevant to this PR, Jerboa should not consume postviewmode settting from the backend.

This setting is very UI specific. And tying all UI to the same mode doesn't really work. These modes will look very different depending on the UI and device. Where I would want to use different modes. For example on browser in lemmy UI I would take a more spacious mode like card. But on the phone I would prefer list mode to see more posts. And on the tablet I might want a "tablet mode" where I can see posts on left and comments on the right. But now I would have to constantly switch between the same account.

I understand the need for this setting in Lemmy UI as users would want their settings to persist between browsers. But Jerboa doesn't need it.

dessalines:
> Title -> content. Meta data around it

Good point. I'll also try to go back to replicating Boost a little:

Screenshot_2024-08-09-18-07-42-90_044f1e467fa7f6fc20a65e6d6e356754

Could you copy-paste your PostViewMode comment to a new issue? I have some thoughts on that but it'd be better to put them on its own issue.

dessalines:
Does this look okay?

Screenshot_2024-08-09-19-04-27-73_c9298fda48802cf3e2d17cb10a48759f

MV-GH:
This looks better. 

But personally I would like title directly against content. Have time top right and votes in bottom bar?

dessalines:
I'll give that a try.

dessalines:
Okay, I've moved the time into the header, and the score and voting back into the action bar. Here are some examples:

Showing full info:

Screenshot_2024-08-10-15-29-45-04_c9298fda48802cf3e2d17cb10a48759f

Some voting examples:

Screenshot_2024-08-10-15-31-24-02_c9298fda48802cf3e2d17cb10a48759f

Screenshot_2024-08-10-15-31-53-77_c9298fda48802cf3e2d17cb10a48759f

dessalines:
K this is ready to be looked at again.

MV-GH:
Great stuff

I believe we show the full title of the community? Maybe its better to just use the shorter "name"? And use the title in the opened post, where we have more space to work with?

image


Interesting behaviour

image

image

If its easier for you, it would definitely be okay with time being part of the flow row there. And it thus not always being in the top right.

dessalines:
It uses the shorter name only if its local. In lemmy-ui (and I think here?) we used to conditionally not show the `@instance.tld` based on various circumstances, but it just got too confusing and unpredictable.

I tried messing around with a few non-flow layouts there, but it always ends up looking strange. So ya I'd say lets just go with this for now.

If you'd like to mess around with it, its `PostListingCard.kt` `L669` and `L675`

MV-GH:
> It uses the shorter name only if its local. 

With that I meant using `name` instead of `title`

Yeah sure. I have been testing this out and images sometimes "jump". Have you noticed that?

dessalines:
@MV-GH lmk when you can get to reviewing these 2 PRs, I'd like to do a release this week, as I'll be out of town next week.

MV-GH:
Yeah sure.

MV-GH (approved):
I haven't really tested. But I ll run master after this locally and see if anything pops up.

dessalines:
Cool.

PR #1632, opened 2024-08-20
Make sup/subscript more strict

### LINKED ISSUE #1628: Markdown rending subtext
### Jerboa Version

0.0.71

### Android Version + Phone

N/a

### Describe The Bug

image

image

This should not be formatted as subtext. But it is following the current spec in Jerboa. Which is any text between a tilde that is uninterrupted by a linebreak becomes subscript.

There doesn't seem to be a commonmark spec to follow?

What does Lemmy UI do?

### To Reproduce

- this text

```
Been driving an EV for 6 years now. Tesla model X (range ~ 260kms) first, now a model Y LR (range ~ 480kms)
```

_No response_

MV-GH:
Fixes #1628


dessalines (approved):
Sry for me being slow, lots of vacation lately.

PR #1650, opened 2024-09-03
Fixes wrong height Navigation bar when descriptions where hidden.

### LINKED ISSUE #1648: Buttons navbar gone with .72 update
Might be just me but I lost the navbar with home, search, notifications. If you squint enough in the screenshot it looks like it is simply cut off but a tiny bit left, so you can see. I tried tapping on the third of a button that I get to see, but they don't work, or some do, but super hard to tap on. Not enough button. 😁

MV-GH:
Since we switched to edgeToEdge. The height now starts from the bottom. (Before the System navigation bar instead of after) This small tweak makes it apply like before.

Fixes #1648

image

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -1,5 +1,8 @@
 @file:Suppress("UnstableApiUsage")
 
+import org.jetbrains.kotlin.compose.compiler.gradle.ComposeFeatureFlag
+
+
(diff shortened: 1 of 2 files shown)


dessalines (approved):
Thx. I'll get a release out with these two fixes.

Issue #1649, opened 2024-09-03
Status/navigation system bars colors don't match with application


MV-GH:
### Jerboa Version

0.0.72

### Android Version + Phone

Android 14

### Describe The Bug

I think this issue has been going for a while for the navigation bar.

![image](https://github.com/user-attachments/assets/77b60a0c-2319-4a01-b7f4-811993270e77)

This code responsible for the navigation bar.
```
    // Check for preview mode
    if (LocalContext.current is Activity) {
        val window = (LocalContext.current as Activity).window
        val colorScheme = MaterialTheme.colorScheme

        DisposableEffect(Unit) {
            window.navigationBarColor = colorScheme.surfaceColorAtElevation(3.dp).toArgb()

            onDispose {
                window.navigationBarColor = colorScheme.background.toArgb()
            }
        }
    }
```
    
And we used to have a translucent status bar, which made the status bar color match with the background. But that was removed in #1607 

### To Reproduce

Start application

MV-GH:
In Jerboa 0.0.60 on Android 12 this is how it used to look. We need to go back to that.

https://github.com/user-attachments/assets/556bf048-82e9-4f40-91ed-5407881ebd1a

MV-GH:
Its good enough for now.

PR #1651, opened 2024-09-04
Fixes mostly systembars colors not matching with application

MV-GH:
The statusbar is not as perfect as before as it would change with the elevation color change, now not anymore.

These changes are needed:

```
insets.isAppearanceLightStatusBars = isLight
insets.isAppearanceLightNavigationBars = isLight
```

This is needed so that if you don't have system set as ThemeMode it will still set the correct (inverse) coloring of the statusbar.    

```
window.statusBarColor = Color.Transparent.toArgb()
```

Needed to mimic translucentStatus. Not as good but pretty close. see #1607 removal

The navigationbar seemed to use a different container color now. So this solves that.

image

#1649


dessalines (approved):
Sweet, thx.

PR #1659, opened 2024-09-06
Fix legacy score behaviour

### LINKED ISSUE #1626: Legacy show scores behaviour is not correct
### Jerboa Version

rework ui

### Android Version + Phone

m/a

### Describe The Bug

If legacy show scores is false it must hide all scores. Which it doesn't do atm

### To Reproduce

1 -> use instance that is pre 0.19.4
2 -> disable show_scores
3 -> still see scores

_No response_

MV-GH:
Fixes #1626

I map the legacy behaviour into the new viewMode makes it much easier to reason about. And we can remove the legacy checks.


dessalines (approved):
Nice, thx.

PR #1685, opened 2024-09-30
Adding the ability to export / import the database.

dessalines:
- Uses the room-db-export-import library I recently made.


app/src/main/res/values/strings.xml
```diff
@@ -496,4 +496,9 @@
     <string name="torrent_help">Lemmy can use a technology called BitTorrent to view or stream media. [Click here](%1$s) to learn how.</string>
     <string name="by">by</string>
     <string name="upvote_percentage">Upvote Percentage</string>
+    <string name="backup_database">Backup Database</string>
```
MV-GH, line 499: Other settings titles have only the first letter capitalized.

MV-GH:
Database restore/save is only usable if the app is on the same version?

i don't exactly like its placement in Settings.

MV-GH:
Yeah sure

MV-GH (approved):

PR #1695, opened 2024-10-06
Update Norwegian Nynorsk translation

huftis:


MV-GH (approved):

app/src/main/res/values-nn/strings.xml
```diff
@@ -358,4 +373,130 @@
...
+  <string name="unable_to_retrieve">Klarte ikkje henta</string>
+  <string name="blocks">Blokkering</string>
+  <string name="unblock">Fjern blokkering</string>
+  <string name="blocked_users">Blokkerte brukarar</string>
+  <string name="blocked_communities">Blokkerte fellesskap</string>
+  <string name="blocked_instances">Blokkerte instansar</string>
+  <string name="you_have_no_blocked_users">Du har ikkje nokon blokkerte brukarar</string>
+  <string name="you_have_no_blocked_communities">Du har ikkje nokon blokkerte fellesskap</string>
+  <string name="you_have_no_blocked_instances">Du har ikkje nokon blokkerte instansar</string>
+  <string name="torrent_help">Lemmy can use a technology called BitTorrent to view or stream media. [Meir informasjon](%1$s)</string>
```
MV-GH, line 493: Btw this is partially still in English

huftis: Good catch! This was a (partial) copy/paste error (a similar string is used in Lemmy). Now fixed.

app/src/main/res/values-nn/strings.xml
```diff
@@ -358,4 +373,130 @@
...
+  <string name="remove_content">Fjern innhald</string>
+  <string name="permaban">Utesteng for godt</string>
+  <string name="days_until_expiration">Utgår (dagar)</string>
+  <string name="feature_in_community">Framhev i fellesskapet</string>
+  <string name="unfeature_in_community">Fjern framheving frå fellesskapet</string>
+  <string name="feature_in_local">Framhev lokalt</string>
+  <string name="unfeature_in_local">Fjern framheving lokalt</string>
+  <string name="swipe_to_action_presets">Innstilling for sveipe­rørsle</string>
+  <string name="swipe_action_preset_only_votes">Berre røyster</string>
+  <string name="swipe_action_preset_only_votes_mirrored">Only votes (spegla)</string>
```
MV-GH, line 440: Here too I believe

huftis: You’re right. Thanks for the thorough review! Now fixed.

Issue #1712, opened 2024-11-01
Redesign for blocks screen


rodrigo-fm:
### Pre-Flight checklist

- [X] Did you check to see if this issue already exists?
- [X] This is a single feature request. (Do not put multiple feature requests in one issue)
- [X] This is not a question or discussion. (Use https://lemmy.ml/c/jerboa for that)

### Describe The Feature Request Below

I think the blocks screen could use a redesign. I was using the app lately and saw the profile screen with these 3 tabs on the top:

![image](https://github.com/user-attachments/assets/41c6627f-3625-4d2f-9295-41e6d95b95ea)

Currently, the blocks screen displays all three lists on the same screen, I think that using tabs will make it more consistent with the general app's design and can help with the user experience (even though I do not have much knowledge about what makes a good user experience).

![image](https://github.com/user-attachments/assets/799bc6e0-1689-4d38-b0ae-6ba22b2a2d24)
*Current blocks screen design*

I also think about displaying the count of blocked elements on each tab, the tabs texts would look something like this:

Users (3) | Communities (2) | Instances (0)

I can also display no number instead of (0) if you think it's better.

Let me know if you guys like this suggestion, I can code it myself. I believe that this change won't require too much refactoring since it's just about rearranging the displayed information.

dessalines:
Agree, that sounds like a good idea. Ping us either here or in the jerboa dev chat if you need any help.

rodrigo-fm:
> Agree, that sounds like a good idea. Ping us either here or in the jerboa dev chat if you need any help.

Thanks, @dessalines. I'll try to open the PR throughout the next week.

PR #1718, opened 2024-11-09
Redesign for blocks screen

rodrigo-fm:
Pull request for #1712 

Here's the new blocks screen:
 
Screenshot_20241108_231047

image

image

### DIFF
--- a/app/src/main/java/com/jerboa/datatypes/Others.kt
+++ b/app/src/main/java/com/jerboa/datatypes/Others.kt
@@ -16,6 +16,7 @@ import androidx.compose.ui.graphics.vector.ImageVector
 import com.jerboa.R
 import com.jerboa.UnreadOrAll
 import com.jerboa.ui.components.person.UserTab
+import com.jerboa.ui.components.settings.block.BlocksTab
 import it.vercruysse.lemmyapi.datatypes.Community
 import it.vercruysse.lemmyapi.datatypes.Person
 import it.vercruysse.lemmyapi.datatypes.Post
@@ -119,6 +120,19 @@ fun getLocalizedStringForUserTab(
         UserTab.About -> ctx.getString(R.string.person_profile_screen_about)
         UserTab.Posts -> ctx.getString(R.string.person_profile_screen_posts)
         UserTab.Comments -> ctx.getString(R.string.person_profile_screen_comments)
+    }
+
+/**
+ * Returns localized Strings for BlocksTab Enum
+ */
+fun getLocalizedStringForBlocksTab(
+    ctx: Context,
+    tab: BlocksTab,
+): String =
+    when (tab) {
+        BlocksTab.Instances -> ctx.getString(R.string.blocked_instances)
+        BlocksTab.Communities -> ctx.getString(R.string.blocked_communities)
+        BlocksTab.Users -> ctx.getString(R.string.blocked_users)
     }
 
 /**
--- a/app/src/main/java/com/jerboa/ui/components/settings/block/BlocksScreen.kt
+++ b/app/src/main/java/com/jerboa/ui/components/settings/block/BlocksScreen.kt
@@ -4,9 +4,11 @@ import android.util.Log
 import androidx.annotation.StringRes
 import androidx.compose.foundation.ExperimentalFoundationApi
 import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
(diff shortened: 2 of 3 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -487,9 +487,9 @@
     <string name="unable_to_retrieve">Unable to retrieve</string>
     <string name="blocks">Blocks</string>
     <string name="unblock">Unblock</string>
-    <string name="blocked_users">Blocked users</string>
-    <string name="blocked_communities">Blocked communities</string>
-    <string name="blocked_instances">Blocked instances</string>
+    <string name="blocked_users">Users</string>
```
MV-GH, line 490: You changed english translation but now it no longer correct. Name doesn't correspond and existing translations either.

Put this back, if not used remove including translations. Check if existing entries exist. If not create ones.

app/src/main/java/com/jerboa/datatypes/Others.kt
```diff
@@ -121,6 +122,19 @@ fun getLocalizedStringForUserTab(
         UserTab.Comments -> ctx.getString(R.string.person_profile_screen_comments)
     }
 
+/**
+ * Returns localized Strings for BlocksTab Enum
+ */
+fun getLocalizedStringForBlocksTab(
+    ctx: Context,
+    tab: BlocksTab,
```
MV-GH, line 130: Make the R.string a property of the enum.

Don't need a switch then

app/src/main/java/com/jerboa/ui/components/settings/block/BlocksScreen.kt
```diff
@@ -86,7 +96,10 @@ fun BlocksScreen(
                     is ApiState.Failure -> ApiErrorText(res.msg)
                     is ApiState.Holder -> {
                         res.data.my_user?.let {
-                            BlockList(it)
+                            BlockList(
+                                userInfo = it,
+                                scope = rememberCoroutineScope()
```
MV-GH, line 101: What do you gain from passing it here Instead of just defining it in the composable?

MV-GH (changes requested):

app/src/main/java/com/jerboa/datatypes/Others.kt
rodrigo-fm: Done, thanks for the tip

app/src/main/java/com/jerboa/ui/components/settings/block/BlocksScreen.kt
rodrigo-fm: I think I gain nothing. I copied the UserTabs from the PersonProfileScreen.kt, I saw that it expects the coroutine scope as an argument, but I removed it from the BlockList arguments, thanks.

app/src/main/res/values/strings.xml
rodrigo-fm: What do you mean by "name doesn't correspond"? Should I change the string res name from "blocked_users" to "users"? These resources are only used in the blocks screen.

The problem here is that the translations are now incorrect when compared to the english text?

rodrigo-fm:
I also centered the empty list message, now it'll show like this:

image

app/src/main/res/values/strings.xml
MV-GH: Yes, you only changed the English translation. I don't expect you fix the other languages, so instead create new entry just for Users, Communities, Instances. The blocked translations wont have any purpose anymore so you can then remove them including their translations

rodrigo-fm: Done. I removed the instances_xxx string resources and created new resources for this screen

rodrigo-fm:
@MV-GH I feel that the PR seems ready to merge, but could you wait until I add the blocked elements count? So that we have this layout for the tabs as proposed in the issue? Users (3) | Communities (2) | Instances (0)

MV-GH:
Sure you mark it draft

dessalines:
Looks good so far! Ping me or request a review when you're ready for that.

rodrigo-fm:
@MV-GH @dessalines I think I'll leave the PR as it is. I tried to add the blocks count to the tabs, but I wasn't able to display it in a horizontally scrollable interface, since the text was too constrained and caused a line break. I tried the ScrollableTabRow composable for this, but it wasn't using the screen's full width, which gave the tabs this incomplete line below them, causing them to have a weird design.

In the end I found this link showing how to reduce the internal padding from the Tabs inside the TabRow https://stackoverflow.com/questions/68949292/jetpack-compose-reducing-internal-padding-of-tabs, but I think it'll be better to leave it to another pull request, so that this one doesn't get too big or overcomplicated for an initially simple idea.

My goal with removing the padding is to avoid the scenario below:

image

MV-GH (approved):

MV-GH:
For counter changes, not so sure removing padding will be better. For the navigation bar we had to resort to making the text smaller.

But its not perfect, depending on language and phone size it can still fail.

PR #1725, opened 2024-11-20
Adding an image upload button for custom thumbnails.

dessalines:
This adds an image upload button for the custom thumbnail field:

Screenshot_20241120_143807_Jerboa (Debug)

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/post/composables/PostComposables.kt
+++ b/app/src/main/java/com/jerboa/ui/components/post/composables/PostComposables.kt
@@ -56,14 +56,16 @@ fun CreateEditPostBody(
     url: String,
     urlField: InputField,
     onUrlChange: (url: String) -> Unit,
+    sharedImage: Uri? = null,
+    onImagePicked: (image: Uri) -> Unit,
+    isUploadingImage: Boolean = false,
     altText: String,
     onAltTextChange: (altText: String) -> Unit,
     customThumbnailField: InputField,
     customThumbnail: String,
     onCustomThumbnailChange: (customThumbnail: String) -> Unit,
-    sharedImage: Uri? = null,
-    onImagePicked: (image: Uri) -> Unit,
-    isUploadingImage: Boolean = false,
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/jerboa/ui/components/post/edit/PostEditScreen.kt
```diff
@@ -131,6 +128,22 @@ fun PostEditScreen(
...
                 isUploadingImage = isUploadingImage,
+                altText = altText,
+                onAltTextChange = { altText = it },
+                customThumbnailField = customThumbnailField,
+                customThumbnail = customThumbnail,
+                onCustomThumbnailChange = { customThumbnail = it },
+                isUploadingCustomThumbnailImage = isUploadingCustomThumbnailImage,
+                onCustomThumbnailImagePicked = { uri ->
+                    if (!account.isAnon() && uri != Uri.EMPTY) {
+                        val imageIs = imageInputStreamFromUri(ctx, uri)
```
MV-GH, line 139: I see that we do this in a lot of places but does this actually work on API level 26-28 ?

app/src/main/java/com/jerboa/ui/components/post/composables/PostComposables.kt
```diff
@@ -178,6 +180,22 @@ fun CreateEditPostBody(
                 keyboardOptions = KeyboardOptions(keyboardType = KeyboardType.Uri),
                 modifier = Modifier.fillMaxWidth(),
             )
+
+            /**
+             * Pick and upload an image, show a preview if possible
+             */
+            PickImage(
```
MV-GH, line 187: I see that we use this in a lot of places but does this actually work on API level 26-28 ?

MV-GH (approved):

app/src/main/java/com/jerboa/ui/components/post/composables/PostComposables.kt
dessalines: Not sure, but I don't see any API warnings for versions. Somebody would have to open an issue if there are.

PR #1801, opened 2025-04-17
Fixes: Webp images using image_proxy can't be saved in Android 10+

### LINKED ISSUE #1800: Can't download or share pictures from some links
### Jerboa Version

0.0.77

### Android Version + Phone

Android 15, Pixel 7

### Describe The Bug

There are some links that seem to refuse to be downloaded, and when you use the "share" button to share the file it tries to send a file called "image_proxy" instead of the actual image file.

Here's an example of a link that caused this:

https://lemmy.ml/api/v3/image_proxy?url=https%3A%2F%2Flemmy.world%2Fpictrs%2Fimage%2Fb78339bf-95ab-4a61-b9ab-bb67696b2a4d.webp

### To Reproduce

1. Tap and hold an image
2. Press the "Save Image" button
3. Observe a toast message that says "Saving Image..." followed shortly by "Failed to save media!"

You get the same issue if you tap the image to full-screem it and click the on-screen "download" icon.

_No response_

MV-GH:
Worked on API Level 28 and below, uses a different API.

It fails to read MIME type from url. Normally thats fine, it will infer later when it gets inserted. For most images that okay, but it seems for webp images (or atleast those from pictrs) it infers the mime as "application/octet-stream" and then it throws exception because `Images.Media.EXTERNAL_CONTENT_URI` doesn't allow that MIME to be inserted there.

I looked at what some other clients do, some dont set MIME at all, meaning those images would also fail. Some always set `image/jpeg`. I added fallback for that here.

Not sure what the future holds for image_proxy endpoint? If kept we might want to better support it, properly extract extension and filename from it.

Fixes #1800 

allDevices was fix for deprecation (just rename)

---
This whole file handling is quite messy in Android. Basically forced to give wrong MIME.

### DIFF
--- a/app/src/main/java/com/jerboa/Utils.kt
+++ b/app/src/main/java/com/jerboa/Utils.kt
@@ -864,10 +864,16 @@ fun saveMediaQ(
     displayName: String,
     mediaType: PostType,
 ): Uri {
+    val mimeTypeWithFallback = mimeType ?: when (mediaType) {
+        PostType.Image -> "image/jpeg"
+        PostType.Video -> "video/mp4"
+        PostType.Link -> "application/octet-stream"
+    }
+
     val values =
         ContentValues().apply {
             put(MediaStore.MediaColumns.DISPLAY_NAME, displayName)
-            put(MediaStore.MediaColumns.MIME_TYPE, mimeType)
+            put(MediaStore.MediaColumns.MIME_TYPE, mimeTypeWithFallback)
             put(MediaStore.MediaColumns.RELATIVE_PATH, mediaType.toMediaDir() + "/Jerboa")
         }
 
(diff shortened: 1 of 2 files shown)


MV-GH:
I don't really like this solution as now it prevents inferring MIME type which was used to set file exentension when saving. Instead just saves the webp as jpeg. For imageviewers it doesn't rlly matter as they read content anyway. I'll check when i have some time if there is anything wrong with webp image.

But I don't see another solution.

dessalines (approved):
Not much we can do about that, its a pictrs proxy issue.

MV-GH:
This change doesn't change anything for share action. As that works somewhat, depends on how the app handles the received extensionless image file.

Ideally we would support the image_proxy endpoint better to extract necessary information that is needed. (image name + extension)
But I believe the image_proxy endpoint is on shaky grounds. Unless I am mistaken? So not sure if we should commit to something that could be removed. 

It just fixes the download action from failing when trying to save extenionless webp images as it seems to be wrongly inferring the MIME type for it.

It shouldn't be downloading/sharing the wrong image. So if you see that behaviour, could post a separate issue with a reproducer.

dessalines:
> Ideally we would support the image_proxy endpoint better to extract necessary information that is needed. (image name + extension)

That's the best option, because that isn't going anywhere, and should be stable. You should be able to assume the mime type from the existing extension.

MV-GH:
Alright I look into it

PR #1813, opened 2025-05-01
Add Lemmy Donation Dialog

### LINKED ISSUE #1794: Add Donation Dialog
### Pre-Flight checklist

- [x] This is a single feature request. (Do not put multiple feature requests in one issue)
- [x] This is not a question or discussion. (Use https://lemmy.ml/c/jerboa for that)

I am one of the Lemmy maintainers and work on the project fulltime together with Dessalines. Our work is funded by donations, but these are gradually going down and don't even cover a single dev salary now (see [join-lemmy.org](https://join-lemmy.org/donate)). That's why we added a new donation dialog in 0.19.11 which is shown once a year to every user:

Image

---

Many people use Lemmy exclusively through apps, so we would greatly appreciate if you could add such a dialog to your app too. The logic is relatively simple:
- From the `/api/v3/site` response, check `my_user.local_user_view.local_user.last_donation_notification`
- If the date is more than one year ago, display a dialog like the one above with buttons Donate, Close
- When Donate is clicked:
	- Open `https://join-lemmy.org/donate`
	- Close dialog
	- Call `POST /api/v3/user/donation_dialog_shown` to hide dialog until next year
- When Close is clicked also call the `donation_dialog_shown` endpoint

To test this functionality with a 0.19.11 instance, run the SQL query `update local_user set last_donation_notification = '2024-04-07 09:05:06';` which shows the dialog for all local users. You can reuse the [code](https://github.com/LemmyNet/lemmy-ui/blob/release/v0.19/src/shared/components/home/donation_dialog.tsx) and [strings](https://github.com/LemmyNet/lemmy-translations/pull/161/files) used in lemmy-ui.

Thanks for your consideration!

iByteABit256:
Fixes #1794

Added a Donation Dialog in MainActivity which shows the donation message in case the user:

- is logged in
- uses a server running a version which supports the donation dialog feature
- hasn't been shown a donation dialog for at least a year

The dialog has a `Donate` button, a `No Thanks` button and the ability to dismiss it by clicking outside.
If the user clicks `Donate` they get navigated to the donation page, and in all of the cases a POST request is then sent to the `donation_dialog_shown` endpoint to update that the user has seen it, then the dialog closes.

Also bumped the `lemmy-api` version to 0.4.1.


app/src/main/java/com/jerboa/model/LocalUserViewModel.kt
```diff
@@ -0,0 +1,36 @@
...
+import androidx.lifecycle.viewModelScope
+import com.jerboa.api.API
+import kotlinx.coroutines.launch
+
+class LocalUserViewModel : ViewModel() {
+    fun markDonationNotificationShown(onComplete: () -> Unit) {
+        if (API.getInstanceOrNull()?.FF?.markDonationDialogShown() == true) {
+            viewModelScope.launch {
+                val result = runCatching {
+                    API.getInstance().markDonationDialogShown()
```
MV-GH, line 14: This is unnecessary

Every endpoint does this already.

see the LemmyApi internals
```kt
suspend inline fun <reified R> HttpClient.postResult(
    urlString: String,
    builder: HttpRequestBuilder.() -> Unit = {},
): Result<R> = runCatching { post(urlString, builder).body() }
```

app/src/main/java/com/jerboa/ui/components/home/donation/DonationDialog.kt
```diff
@@ -0,0 +1,139 @@
...
+                    showDonationDialog = false
+                }
+            },
+        )
+    }
+}
+
+@OptIn(ExperimentalComposeUiApi::class)
+@Composable
+fun DonationNotificationDialog(
```
MV-GH, line 101: Can you make this private, wasn't clear to me that was only used for the preview

app/src/main/java/com/jerboa/ui/components/home/donation/DonationDialog.kt
```diff
@@ -0,0 +1,139 @@
...
+    onClick: () -> Unit,
+    onDismiss: () -> Unit,
+) {
+    MarkdownHelper.init(LocalContext.current)
```
MV-GH, line 105: This is for preview only use.

Else it will override init by the main.

app/src/main/java/com/jerboa/ui/components/home/donation/DonationDialog.kt
```diff
@@ -0,0 +1,139 @@
...
+    Nutomic and Dessalines,
+    
+    Lemmy Developers
+
+    """.trimIndent()
+
+@OptIn(ExperimentalComposeUiApi::class)
+@Composable
+fun ShowDonationNotification(siteViewModel: SiteViewModel) {
+    var showDonationDialog by remember { mutableStateOf(false) }
```
MV-GH, line 56: Use `rememberSaveable` else many changes can reset this to false. Like screen rotation.

app/src/main/java/com/jerboa/ui/components/home/donation/DonationDialog.kt
```diff
@@ -0,0 +1,139 @@
...
+                    ?.last_donation_notification
+                    ?: return@LaunchedEffect
+                if (lastDonationNotification.isEmpty()) return@LaunchedEffect
+
+                try {
+                    val lastDonationTime = OffsetDateTime.parse(lastDonationNotification, DateTimeFormatter.ISO_DATE_TIME)
+                    val oneYearAgo = OffsetDateTime.now().minusYears(1)
+                    showDonationDialog = lastDonationTime.isBefore(oneYearAgo)
+                } catch (e: Exception) {
+                    Log.d("ShowDonationNotification", "could not parse datetime", e)
```
MV-GH, line 74: Log.e

app/src/main/java/com/jerboa/model/LocalUserViewModel.kt
```diff
@@ -0,0 +1,36 @@
+package com.jerboa.model
+
+import android.util.Log
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
+import com.jerboa.api.API
+import kotlinx.coroutines.launch
+
+class LocalUserViewModel : ViewModel() {
```
MV-GH, line 9: Why did you make a ViewModel just for this?

iByteABit256: I wanted to keep the same project structure as I see with the other API calls, should I just call it directly from the component instead?

MV-GH: Yes, we usually use as a convient way to bind some logic to the lifecycle of the UI component. But here is not needed.

iByteABit256: Ok, changed it

app/src/main/res/values/strings.xml
```diff
@@ -502,4 +502,6 @@
     <string name="database_restored">Database Restored.</string>
     <string name="database_backed_up">Database Backed up.</string>
     <string name="settings_screen_backup_and_restore">Backup and restore</string>
+    <string name="donate">Donate</string>
+    <string name="donation_dialog_dismiss">No Thanks</string>
```
MV-GH, line 506: Maybe `No, thanks` ?

app/src/main/java/com/jerboa/ui/components/home/donation/DonationDialog.kt
```diff
@@ -0,0 +1,162 @@
...
+    DonationNotificationDialog({}, {})
+}
+
+fun markDonationNotificationShown(
+    scope: CoroutineScope,
+    onComplete: () -> Unit,
+) {
+    if (API.getInstanceOrNull()?.FF?.markDonationDialogShown() == true) {
+        scope.launch {
+            val result = API.getInstance().markDonationDialogShown()
```
MV-GH, line 148: Chain it directly?

MV-GH (commented):
How does this work with changelog dialog? Does it stack on top of each other?

Looks otherwise good to me.

app/src/main/java/com/jerboa/ui/components/home/donation/DonationDialog.kt
iByteABit256: Can't chain it because I'm using the feature flags on l146, I refactored it with an api variable instead

iByteABit256:
The changelog would conflict a bit with the donation dialog if both should be shown, I made a combined app startup dialogs component that manages the order they should be shown

app/src/main/java/com/jerboa/ui/components/home/AppStartupDialogs.kt
```diff
@@ -0,0 +1,142 @@
...
+    when (activeDialog) {
+        DialogType.Changelog -> {
+            val markdown = appSettingsViewModel.changelog
+            LaunchedEffect(appSettingsViewModel) {
+                appSettingsViewModel.loadChangelog(ctx)
+            }
+
+            ShowChangelog(
+                onClick = {
+                    appSettingsViewModel.updateLastVersionCodeViewed(currentVersionCode)
```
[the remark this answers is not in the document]
iByteABit256: Yes that's correct

app/src/main/java/com/jerboa/ui/components/home/AppStartupDialogs.kt
```diff
@@ -0,0 +1,142 @@
...
+                Log.e("ShowAppStartupDialogs", "Failed to parse donation time", e)
+                false
+            }
+        } else {
+            false
+        }
+    }
+
+    // Determine which dialog to show
+    LaunchedEffect(changelogNeedsToShow, donationNeedsToShow) {
```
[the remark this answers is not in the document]
iByteABit256: It works like you mentioned in the later comment, and I also made it to check activeDialog for null. This is because changelog changing to shouldShow = true for example would override the donation dialog currently showing. Now it waits until the current dialog is closed and then it changes the activeDialog.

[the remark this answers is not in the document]
MV-GH: But I don't see why its needed?

Couldn't you get the exact same result by just removing the effect?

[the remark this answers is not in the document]
iByteABit256: I can imagine a scenario where the dialog has closed and `activeDialog` is null again, but the shouldShow logic for one of the dialogs hasn't received updated data yet for the condition to change, making the dialog open up again a second time

app/src/main/java/com/jerboa/ui/components/home/AppStartupDialogs.kt
```diff
@@ -0,0 +1,151 @@
...
+    scope.launch {
+        api
+            .markDonationDialogShown()
+            .onSuccess { onComplete() }
+            .onFailure {
+                Log.e("markDonationNotificationShown", "Failed to mark donation shown", it)
+                onComplete()
+            }
+    }
+}
```
dessalines, line 146: LGTM

dessalines (approved):
Looks good to me.

The only potential change I can think of, is to prepend the donation dialog text to the changelog when both need to be shown, but that's up to you.

app/src/main/java/com/jerboa/ui/components/home/AppStartupDialogs.kt
```diff
@@ -0,0 +1,151 @@
...
+import com.jerboa.getVersionCode
+import com.jerboa.model.AppSettingsViewModel
+import com.jerboa.model.SiteViewModel
+import com.jerboa.ui.components.common.DonationNotificationDialog
+import com.jerboa.ui.components.common.ShowChangelog
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+import java.time.OffsetDateTime
+
+const val DONATION_LINK = "https://join-lemmy.org/donate"
```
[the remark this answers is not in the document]
iByteABit256: done

dessalines (approved):
Just tested this out, and it works well. Thx!

PR #1814, opened 2025-05-01
Add Image Proxy endpoint support to Share/Download actions

MV-GH:


dessalines (approved):
Thx, and for the test too.

Can I start a release of this, or do you still want to look at the navbar?

MV-GH:
I'll take another look at that issue this weekend if not we can do a release.

PR #1823, opened 2025-05-10
Add Android 15 SDK support, Configure Android lint, deprecation fixes

MV-GH:
Discovered Android Lint recently, Its pretty handy. I discovered some issues with the translations with it. I have added it to the CI.

It generates a report like this.
image

Also bump SDK to 35 and Compose Bom to 2025.05

Fixes LocalClipboardManager deprecations (new API)

Fixes autofill deprecations (new API)

the setColor for statusBar and navigationBar have been deprecated because in Android 15. Its fixed to some transparent mode.

So not sure yet how to handle this for pre android 15. Will check that out later.

### DIFF
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -29,6 +29,16 @@ steps:
     when:
       - event: pull_request
 
+  check_android_lint:
+    image: *android_image
+    commands:
+      - sudo chown -R circleci:circleci .
+      - ./gradlew lint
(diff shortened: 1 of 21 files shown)


dessalines:
Nice. It might be the same as the ones you can run inside android studio, I do run those occasionally.

dessalines (approved):

PR #1824, opened 2025-05-11
Fixes #1748 Extra padding above IME keyboard

### LINKED ISSUE #1748: Padding above IME keyboard in createPosts
### Jerboa Version

master

### Android Version + Phone

Happens on Android 11 + Android 14

### Describe The Bug

See image

Image

### To Reproduce

- create comment
- long enough text

_No response_

MV-GH:
The imePadding seemed to also apply the navigationBar padding. Using this `consumeWindowInsets` made it only apply once.

Also refactored `MarkdownTextField` slightly to not have that required maxFillSize default as part of params and apply the same modifier twice.


dessalines (approved):
👍

PR #1825, opened 2025-05-11
Fixes #1755 Keyboard still open after creating comment

### LINKED ISSUE #1755: virtual keyboard not closed after comment post
### Jerboa Version

0.77

### Android Version + Phone

11

### Describe The Bug

When user write a comment and send it, comment view is left but sometimes virtual keyboard is reopened.

It's not systematic, it feels like a short race with keyboard toggling. 

I am not Android UI specialist but it's maybe in callback event when comment editor has been destroyed.

### To Reproduce

Select post
Create comment
Send comment

_No response_

MV-GH:
When sending the comment, it goes into "loading" mode, here the CommentReplyView leaves the composition, then when it receives response it goes into ApiState Success. Now the previous ReplyView enters composition again for short period. Where it asks for focus again. Thus the open keyboard state after creating a comment.

This fixes it by not showing the CommentReply view after sending. Thus it doesnt ask for focus.

Also removed the focus removal. It doesn't seem to do anything.

https://github.com/user-attachments/assets/345e0554-3d78-4007-aed0-fff8438af54f

The other screens aren't affected because they don't show a loading bar and hide the body. Thus it doesn't leave composition. :)

Fixes #1755

### DIFF
--- a/app/src/main/java/com/jerboa/model/CommentReplyViewModel.kt
+++ b/app/src/main/java/com/jerboa/model/CommentReplyViewModel.kt
@@ -5,7 +5,6 @@ import android.util.Log
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.setValue
-import androidx.compose.ui.focus.FocusManager
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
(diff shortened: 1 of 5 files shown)


dessalines (approved):
Weird bug, I haven't gotten it on several devices.

PR #1830, opened 2025-05-19
Adding weblate translations url.

dessalines:
Switches to using weblate like all our other lemmy translations.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -77,9 +77,9 @@ prettier --write "*.md" "*.yml"
 
 ## Adding translations
 
-Translations via [weblate](https://weblate.join-lemmy.org/engage/lemmy/):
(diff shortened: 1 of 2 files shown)


MV-GH:
Can you directly link to this https://weblate.join-lemmy.org/projects/lemmy/jerboa/

Also is that all that is needed for the integration to work with weblate?

dessalines:
Yeah sure.

Yep, weblate should have push perms added for jerboa. We can also do:

```
git remote add weblate http://weblate.join-lemmy.org/git/lemmy/jerboa/
git fetch weblate
```

If its not working for some reason.

MV-GH:
Will the locale also be added automatically?

MV-GH:
That is true, but it wouldnt hurt to keep that pasage.

PR #1840, opened 2025-06-01
Add "Copy Image" action to Image context menu

### LINKED ISSUE #1839: Add "Copy Image" button to context menu for images
### Pre-Flight checklist

- [x] This is a single feature request. (Do not put multiple feature requests in one issue)
- [x] This is not a question or discussion. (Use https://lemmy.ml/c/jerboa for that)

In the current implementation of Jerboa, it has the following options on the context menu when pressing and holding images:

Image

Extensive, but missing the ability to copy an image to clipboard.

Firefox Mobile has this ability:

Image

Image

would be nice to have this to save a little time.

MV-GH:
https://github.com/user-attachments/assets/af97e4cc-8c40-4d3b-862a-7355b8d86e6c

(Android 13, with system info msg)

Add copy image action, for android 13 it won't display toast since system message should popup. Technically we should do the same for copy text to clipboard. But Idk seems to work rather very sporadic. Maybe its just my emulator but it barely pops up.

Also includes:
- (Slight) Refactor copyTextToClipboard, not sure why it was done previously like that, maybe was copied from fragments in Android Views UI. But Compose (single activity) this retrieval shouldn't ever fail AFAIK
- Add missing hu locale
- Some Pruning/refactor in String resource files, removed some duplicates

Tested on real device.

Fixes #1839


app/src/main/java/com/jerboa/ui/components/comment/CommentOptionsDropdown.kt
```diff
@@ -116,21 +100,7 @@ fun CommentOptionsDropdown(
...
-                            ).show()
-                    } else {
-                        Toast
-                            .makeText(
-                                ctx,
-                                ctx.getString(R.string.generic_error),
-                                Toast.LENGTH_SHORT,
-                            ).show()
-                    }
+                    copyTextToClipboard(ctx, content, "comment", R.string.comment_node_comment_copied)
```
dessalines, line 103: Much cleaner, thx.

app/src/main/java/com/jerboa/ui/components/common/LinkDropDownMenu.kt
```diff
@@ -80,30 +81,16 @@ fun LinkDropDownMenu(
...
-                            ).show()
-                    } else {
-                        Toast
-                            .makeText(
-                                ctx,
-                                ctx.getString(R.string.generic_error),
-                                Toast.LENGTH_SHORT,
-                            ).show()
-                    }
+                    copyTextToClipboard(ctx, link, "Link", R.string.ink_copied)
```
dessalines, line 88: `ink_copied` should be `link_copied`

app/src/main/res/values/strings.xml
```diff
@@ -52,26 +51,21 @@
     <string name="commentReply_send">Send reply</string>
     <string name="comment_body_deleted">*Deleted*</string>
     <string name="comment_body_removed">*Removed*</string>
-    <string name="comment_edit_edit">Edit</string>
```
dessalines, line 55: We need to get rid of all these pointless precursors at some point.

app/src/main/res/xml/locales_config.xml
```diff
@@ -21,4 +21,5 @@
     <locale android:name="bg"/>
     <locale android:name="tr"/>
     <locale android:name="zh"/>
+    <locale android:name="hu"/>
```
dessalines, line 24: Does the lint catch these now?

PR #1846, opened 2025-06-06
Fix edgecase saveImage failing causing crash

MV-GH:
A failure like this below in `saveMedia` action would cause the app to crash. Due to missing catch. I checked some other actions they seem good.

> java.lang.IllegalStateException: Failed to build unique file: /storage/emulated/0/Pictures/Jerboa image_proxy

Now the failure itself. Apparently you can only have like 32 duplicate filenames before it starts throwing that failure. For 

See screenshots below (This could be selective to only a few Android versions)

image

image

This was found because it would save all image_proxy images as "image_proxy" likely would not have been encountered else. But if we do seem to be encountering this elseware and it not being a specific android version specific. We could look into preventing this failure. (Randomized hash as filename or looking up how some other apps might deal with this)


dessalines (approved):

dessalines:
Weird error. Yep a random filename seems like it'd be best.

PR #1865, opened 2025-06-30
Fix LemmyAPI build

MV-GH:
LemmyAPI was published to OSSHR but that got sunset. And all my SNAPSHOT deployments got removed.
see  https://central.sonatype.org/pages/ossrh-eol/

So i have published it to mavencentral now.

Also fixed some deprecations regarding gradle DSL see https://kotlinlang.org/docs/gradle-compiler-options.html#migrate-from-kotlinoptions-to-compileroptions


dessalines (approved):
👍

PR #1936, opened 2025-08-30
Add option to disable video auto play

MV-GH:
Most work was propdrilling that setting.

I ve been thinking about using compositionprovider to push the settings down, and only pass by arg in the relevant composables.

But that would be a lot of refactor work to setup initially and I don't know how you feel about it.

https://github.com/user-attachments/assets/23d7871c-cfac-4bd2-aacb-f54be517a847

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/post/PostListingCard.kt
+++ b/app/src/main/java/com/jerboa/ui/components/post/PostListingCard.kt
@@ -1018,12 +1018,13 @@ fun PostTitleBlock(
         (postUrl != null && VideoHostComposer.isVideo(postUrl))
 
     when {
-        videoPost && expandedImage && !disableVideoAutoplay -> {
+        videoPost && expandedImage -> {
             PostTitleAndVideoLink(
                 postView = postView,
                 appState = appState,
                 showIfRead = showIfRead,
                 blurNSFW = blurNSFW,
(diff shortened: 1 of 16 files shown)


app/src/main/java/com/jerboa/ui/components/post/PostListingCard.kt
```diff
@@ -1006,7 +1018,7 @@ fun PostTitleBlock(
         (postUrl != null && VideoHostComposer.isVideo(postUrl))
 
     when {
-        videoPost && expandedImage -> {
+        videoPost && expandedImage && !disableVideoAutoplay -> {
             PostTitleAndVideoLink(
```
dessalines, line 1022: Shouldn't this be internal to that composable? This makes it seem like you're also hiding the post title and video, rather than just not playing the video.

It should probably display a frame, and just not autoplay, if that's possible.

dessalines:
The prop drilling is annoying I agree, but I much prefer it any obfuscated param injection methods, so ya I'd like to keep all the explicit props as they are.

I def would be willing to consider just passing the entire settings object to all the composables that need it tho (as a separate PR). Then at least we wouldn't have to define a ton of things. That also has drawbacks tho, like not being able to easily set up previews, and not being clear about what specifically the component needs.

PR #1968, opened 2025-09-16
Regenerate baseline profiles

MV-GH:
```
Comparison with previous startup profile:
  28087 Old rules
  28404 New rules
  385 Added rules (1.35%)
  68 Removed rules (0.24%)
  28019 Unmodified rules (98.41%)


> Task :app:generateBaselineProfile

BUILD SUCCESSFUL in 33m 43s
```

I still get the warning when building a release 🤷 

Ig you can ignore it.


MV-GH:
Maybe thats bc of the preprocessing removal changes in the build.gradle

WIll look into it later

Issue #1711, opened 2024-10-28
Weird scrolling behaviour in post comments


Ghoelian:
### Jerboa Version

0.0.77

### Android Version + Phone

Android 15, Pixel 8 pro

### Describe The Bug

On this one specific post, if I scroll all the way down to the bottom, and then scroll back up not too fast, the scroll position gets reset constantly.

I haven't seen this happen on any other posts yet.

I'll edit the issue with a video later.

### To Reproduce

1. Open https://lemmy.dbzer0.com/post/30338733
2. Scroll all the way to the bottom
3. Scroll back up kinda slowly, doesn't happen when I scroll quick

dessalines:
Probably has to do with the rendered picture its scrolling past. This is pretty complicated because compose is probably recalculating the picture dimensions when it scrolls back into view, causing the recompositions. It involves our markdown library, the picture renderer, and our own threaded tree components.

g1t3m:
Present in the following version:
version 0.0.87-fdroid

com.jerboa
versionCode 87

targetSdk 36
minSdk 26

Installed: Apr 3, 2026; 9:16 PM
Updated: May 2, 2026; 12:00 AM

dessalines:
Yes its still a known issue, having to do with recompositions.

PR #1979, opened 2025-10-02
Reduce recompositions in the comments

MV-GH:
Well I have reduced the recompositions in the comments, it no longer recomposes while it scrolls.

Cause was the topbar, the padding constantly changes, by moving the body into a separate composable I create a new "root" composable, This composable skips recompositions due to its params never changing. Before it everytime it would recreate everything below when the padding changed. It would recreate the commentsTreeList, the lambda, ... bc of that those comments all had changed params. and thus the whole tree down recomposed.

As you can see from the video when you do any action though, it fully recomposes the entire comments. To achieve only the vote itself recomposing, it would need to refactored to use something similar like we did for the postfeed. But thats too much work for now. Cuz currently it creates a new list each time an actions happens upon the comments. Thus they arent equal anymore.

Sadly this did not end up fixing #1711

So there is still something else going on that makes those images pop in and pop out constantly.

before (whole tree recomposes on every scroll)

https://github.com/user-attachments/assets/eac78dd6-5e5a-47ed-8a28-c1eba8392928

after (only the pulldownrefresh recomposes because padding changes, and mainbody skips every recompose)

https://github.com/user-attachments/assets/3b6d2058-5f92-464e-b407-831862498e71

### DIFF
--- a/app/src/main/java/com/jerboa/JerboaAppState.kt
+++ b/app/src/main/java/com/jerboa/JerboaAppState.kt
@@ -1,6 +1,7 @@
 package com.jerboa
 
 import android.os.Parcelable
+import android.util.Log
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.Stable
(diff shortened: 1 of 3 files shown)


dessalines (approved):
What a weird hidden behavior.... Its probably a good practice then to always create new composables whenever padding is used.

PR #2029, opened 2025-12-19
Disable fastlane changelogs.

### LINKED ISSUE #2028: fastlane changelogs too long
```
com.jerboa/metadata/en-US/changelogs/80.txt has 926 chars
com.jerboa/metadata/en-US/changelogs/84.txt has 1302 chars
```

fastlane's limit for changelogs is 500 chars, everything beyond that position is simply cut off (even mid-word). It's fine to include a link to the full changelog at their end – but simply summing up all commit messages in the changelogs is… well, not doing the job (with the last one, 80% were only relevant to the app developers even 🙈).

Can you please take care to stay within those limits with future releases? I've just manually cleaned-up the last log at our end, so the relevant things are still being shown.

Thanks in advance!

dessalines:
- Fixes #2028

### DIFF
--- a/scripts/generate_changelog.sh
+++ b/scripts/generate_changelog.sh
@@ -14,19 +14,15 @@ sed -i "s/versionName = .*/versionName = \"$new_tag\"/" $app_build_gradle
 
 # Writing to the Releases.md asset that's loaded inside the app
 tmp_file="tmp_release.md"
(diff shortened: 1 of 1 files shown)


dessalines:
I did, but the f-droid / izzy ppl let me know that just linking to the changelog in individual release files was also pointless, since the main fastlane file already links to it.

PR #2053, opened 2026-01-23
AGP 9 upgrade

### LINKED ISSUE #2052: Upgrade to AGP 9
I took a quick attempt at it but AGP 9 is a complex upgrade.

Lots DSL changes in the build.

Some of the plugins don't support AGP 9 yet either which is blocking me atm from going further.

There is also this block that is gone

```
         postprocessing {
                isRemoveUnusedCode = true
                isObfuscate = false
                isOptimizeCode = true
                isRemoveUnusedResources = true
                proguardFiles("proguard-rules.pro")
            }
```

There is also this

```
This version of the Baseline Profile Gradle Plugin was tested with versions below Android
Gradle Plugin version Android Gradle Plugin version 9.0.0-alpha01 and it may not work as intended.
Current version is Android Gradle Plugin version 9.0.0.

This warning can be disabled setting the following property:
baselineProfile {
    warnings {
        maxAgpVersion = false
    }
}
```

Disabling this warning does not work at all.

I recommend waiting with this upgrade for now.

MV-GH:
Current progress

see https://developer.android.com/build/releases/gradle-plugin-roadmap

and https://developer.android.com/build/releases/agp-9-0-0-release-notes

Fixes #2052


MV-GH:
I got it to build now. Would recommend still waiting, still have to do weird workarounds...

at least figured out to replace the postprocessing. it was just setting that dontobfuscate flag

MV-GH:
Man how is this still so painful after 6 months

MV-GH:
I think, i got it now

dessalines (approved):
Nice, thx for handling this. Any build tooling updates can be a nightmare, its always difficult to find what they changed.

PR #2112, opened 2026-04-10
Adding freedroidwarn

dessalines:


MV-GH:
Is this like a soft deprecation of this app then?

How does those requirements impact the distribution through fdroid and izzy store?

dessalines:
No its not a deprecation, its a warning most FOSS apps are adding about this upcoming android developer-ID scheme from google: https://keepandroidopen.org

~~It won't affect alternative app stores~~, although I'm trying to get more devs to collectively be willing to yank their apps from google play if this goes through: https://github.com/keepandroidopen/keepandroidopen.github.io/issues/115

MV-GH:
Since we also distribute through fdroid, should those also see the warning then?

dessalines:
Ya IMO, this also puts f-droid at risk, and its best to spread awareness about this everywhere we can.

To be specific: 

> On August 25th, 2025, [Google](https://consumerrights.wiki/w/Google) announced an upcoming application installation restriction on Google-certified [Android](https://consumerrights.wiki/w/Android) devices, requiring all developers to register and verify their real-life identity through the Developer Verification program and be approved by Google before their apps can be installed on Android devices. This requirement extends to all installation methods including "[sideloading](https://consumerrights.wiki/w/Sideloading)", third-party app repositories like [F-Droid](https://consumerrights.wiki/w/F-Droid), and direct APK installations. Google stated that this change "keeps the ecosystem open".[[1]](https://consumerrights.wiki/w/Android_Developer_Verification#cite_note-1)

MV-GH:
Alright, is this a onetime popup? or every startup?

dessalines:
One time per upgrade / version.

MV-GH (approved):

PR #2116, opened 2026-04-22
Consolidate gplay and fdroid versions using flavors

dessalines:
Google play required me to remove all donation info, so until now I'd been keeping a separate branch.

I recently learned how to use [product flavors](https://developer.android.com/build/build-variants#product-flavors), which are a better way to handle this, by building multiple variants.


MV-GH (approved):
Since when did gplay have this requirement?

dessalines:
At least a year I think. I pushed an update and they kept refusing to publish.

PR #2123, opened 2026-04-27
restrict low bandwidth auto mode to Data Saver

KiaraGrouwstra:
Drop the metered-network check from `LowBandwidthMode.Auto` so it activates only when the system Data Saver setting is enabled. Most mobile networks report as metered, which made Auto trigger far more often than users expected.

Addresses https://github.com/LemmyNet/jerboa/issues/1594#issuecomment-4330034240.

/cc @MV-GH


MV-GH (approved):

dessalines (approved):
I'll try to get a release out within a few days or so.

Issue #2142, opened 2026-05-25
Issue with Backup database.


SGaming700:
### Jerboa Version

0.0.87-gplay

### Android Version + Phone

Android 13 Go, FOXXD A55

### Describe The Bug

There seems to be a problem with backing up the app database because when I go to settings and I back up the files for it, it shows up as 0 bytes and when I try to restore it everything goes back to default and removes my saved settings so there is something wrong going on there.

### To Reproduce

1. Go to the menu on the top right.
2. Go to settings. 
3. Go to backup and restore.
4. Tap backup database

dessalines:
I just tested this, and can't verify. The backup wasn't 0 bytes, and restore worked fine.

Are you sure there isn't some issue with storage permissions for the folder you're backing up to?

SGaming700:
Here's some evidence https://drive.google.com/drive/folders/1qC0mkfuJJP-6tQfqgnssIxBJQeRO-giU

MV-GH:
we don't have access to that but you can directly upload files here

SGaming700:
Huh I thought that link would work let me try to send the files here.

https://github.com/user-attachments/assets/86fd9e58-df99-414b-8b33-d1e10ef233a2

<img width="384" height="768" alt="Image" src="https://github.com/user-attachments/assets/f3754461-e6bb-419e-a22a-6e5653672eea" />

dessalines:
This is gonna be a tough one. The DB library I wrote should be fine for android versions above `O` / `8.0` 

Here's the code for it: https://github.com/dessalines/room-db-export-import/blob/main/room-db-export-import/src/main/java/com/roomdbexportimport/RoomDBExportImport.kt

MV-GH:
There is a lot that could wrong there, it should return a result so that Jerboa can handle/know about it when it fails. So that it can inform and display the failure. Currently it swallows any IO exception. And the jerboa code will just save an empty file

```kt
    val exportDbLauncher =
        rememberLauncherForActivityResult(
            ActivityResultContracts.CreateDocument("application/zip"),
        ) {
            it?.also {
                dbHelper.export(ctx, it)
                Toast.makeText(ctx, dbSavedText, Toast.LENGTH_SHORT).show()
            }
        }
```

dessalines:
True, that'd be a good idea to have that return a result. 

Although it is printing a stacktrace, its not going to be easy to view it since it isn't causing a crash.

PR #2145, opened 2026-05-27
Upgrade room-db-export to toast failure messages

dessalines:
#2142

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/settings/backupandrestore/BackupAndRestoreScreen.kt
+++ b/app/src/main/java/com/jerboa/ui/components/settings/backupandrestore/BackupAndRestoreScreen.kt
@@ -26,6 +26,7 @@ import androidx.compose.ui.tooling.preview.Preview
 import com.jerboa.R
 import com.jerboa.db.AppDB
 import com.jerboa.ui.components.common.SimpleTopAppBar
+import com.jerboa.ui.components.common.apiErrorToast
 import com.roomdbexportimport.RoomDBExportImport
 import me.zhanghai.compose.preference.Preference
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/jerboa/ui/components/settings/backupandrestore/BackupAndRestoreScreen.kt
```diff
@@ -48,8 +48,13 @@ fun BackupAndRestoreScreen(onBack: () -> Unit) {
...
         ) {
             it?.also {
-                dbHelper.export(ctx, it)
-                Toast.makeText(ctx, dbSavedText, Toast.LENGTH_SHORT).show()
+                dbHelper
+                    .export(ctx, it)
+                    .onSuccess {
+                        Toast.makeText(ctx, dbSavedText, Toast.LENGTH_SHORT).show()
+                    }.onFailure { e ->
+                        Toast.makeText(ctx, e.message, Toast.LENGTH_SHORT).show()
```
MV-GH: This is a lot better but those toasts will only show tiny fraction of the error, so it still could be a problem.

Also is this all blocking IO on the main thread? bc toasts need to be fired from the main thread.

FYI there is apiErrorToast

but preferably you could wire up ApiErrorText?

dessalines: > apiErrorToast

Good call, I'll use this.

> Also is this all blocking IO on the main thread? bc toasts need to be fired from the main thread.

No I don't believe so, `rememberLauncherForActivityResult` doesn't seem to be affecting IO. The toasts do work correctly.

MV-GH (approved):

PR #2162, opened 2026-07-07
Fix scroll stuck/glitchy in comments when scrolling up to an image

### LINKED ISSUE #1711: Weird scrolling behaviour in post comments

MV-GH:
I have spent a lot of time on this issue. The issue is with the interaction of the lazycolumn and the markwon view (more specifically the drawable image). When you scroll down you pull the post into view -> it creates the comment composable which in turn creates the markdownview which instantiates the drawable. Has initially just the 'text' height,  image loads in with new height which forces the textview to relayout, then is propagated to its parents where compose has to do a remeasure pass. Which causes to shift outside the viewport boundry of the lazy column. Which then disposes of it and moves the feed to the previous item.

This isn't a "proper" fix but it works. its the same technique i used to prevent the videos from flickering when scrolling in the feed. Because we reuse an already init drawable, its dimension are immediately known. Thus bypassing this relayout, and keeps the scroll smooth.

Before:

https://github.com/user-attachments/assets/e5c8509e-e222-448d-acd0-9a2b7a81647d

After:

https://github.com/user-attachments/assets/79e62574-9096-459b-9158-1926302bbc56

This might also fix it, not sure though.

```
 * This overload of [AndroidView] does not automatically pool or reuse Views. If placed inside of a
 * reusable container (including inside a [LazyRow][androidx.compose.foundation.lazy.LazyRow] or
 * [LazyColumn][androidx.compose.foundation.lazy.LazyColumn]), the View instances will always be
 * discarded and recreated if the composition hierarchy containing the AndroidView changes, even if
 * its group structure did not change and the View could have conceivably been reused.
 *
 * To opt-in for View reuse, call the overload of [AndroidView] that accepts an `onReset` callback,
 * and provide a non-null implementation for this callback. Since it is expensive to discard and
 * recreate View instances, reusing Views can lead to noticeable performance improvements —
 * especially when building a scrolling list of [AndroidViews][AndroidView]. It is highly
 * recommended to opt-in to View reuse when possible.
 
 ```
 
 from AndroidView but last we tried onReset was used, It caused clickevents to go the wrong post :) (the factory is only called once and the update sets the new text but the factory holds the old lambdas which holds old ids which opens the some other post)
 
 AndroidView are expensive compared to Composables which are designed to be lightweight and constantly rerender (recompose)

Fixes #1711

### DIFF
--- a/app/src/main/java/com/jerboa/ui/components/comment/CommentNode.kt
+++ b/app/src/main/java/com/jerboa/ui/components/comment/CommentNode.kt
@@ -142,6 +142,7 @@ fun CommentNodeHeaderPreview() {
 @Composable
 fun CommentBody(
     comment: Comment,
+    postId: PostId,
     viewSource: Boolean,
     onClick: () -> Unit,
     onLongClick: ((View) -> Boolean),
@@ -159,6 +160,9 @@ fun CommentBody(
             onClick = onClick,
             onLongClick = onLongClick,
             modifier = Modifier.padding(0.dp, 0.dp, 0.dp, MEDIUM_PADDING),
+            // commentId alone isn't globally unique (it's only unique per post/instance), so
+            // combine it with postId for a cache key that's unique enough in practice.
+            cacheKey = "${postId}_${comment.id}",
         )
     }
 }
@@ -168,6 +172,7 @@ fun CommentBody(
 fun CommentBodyPreview() {
     CommentBody(
         comment = sampleCommentView.comment,
+        postId = sampleCommentView.post.id,
         viewSource = false,
         onClick = {},
         onLongClick = { true },
@@ -345,6 +350,7 @@ fun LazyListScope.commentNodeItem(
                             Column {
                                 CommentBody(
                                     comment = commentView.comment,
+                                    postId = commentView.post.id,
                                     viewSource = viewSource,
                                     onClick = { onCommentClick(commentView) },
                                     onLongClick = { v ->
--- a/app/src/main/java/com/jerboa/ui/components/comment/CommentNode.kt
+++ b/app/src/main/java/com/jerboa/ui/components/comment/CommentNode.kt
@@ -142,7 +142,6 @@ fun CommentNodeHeaderPreview() {
 @Composable
 fun CommentBody(
     comment: Comment,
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: app/src/main/java/com/jerboa/ui/components/comment/CommentNode.kt [hunk 1]
```diff
@@ -159,6 +160,9 @@ fun CommentBody(
             onClick = onClick,
             onLongClick = onLongClick,
             modifier = Modifier.padding(0.dp, 0.dp, 0.dp, MEDIUM_PADDING),
+            // commentId alone isn't globally unique (it's only unique per post/instance), so
+            // combine it with postId for a cache key that's unique enough in practice.
+            cacheKey = "${postId}_${comment.id}",
```
dessalines: This isn't right, comment ids are unique per instance, so the post id shouldn't be necessary. You'll never see a duped comment id from the same API.

dessalines (commented):
The caching scares me a little, mainly because comment content can be edited remotely, and the cache key would then potentially re-use stale data. 

However if this were re-instantiated on every comment section load, that wouldn't be a problem. 

We also have all the other cache-related concerns: would this be destroyed and memory freed after leaving a comment section, or would it keep piling onto memory.

app/src/main/java/com/jerboa/ui/components/comment/CommentNode.kt
MV-GH: Ah I didnt know, i'll change it

dessalines (approved):
We can merge this, with at least the caveat that evenually when we switch to multi-platform markdown-renderer, it shouldn't be needed.

PR #2172, opened 2026-07-13
Adds option to enable/disable infinite scroll

rodrigo-fm:
Here are a few pictures of the navigation without infinite scroll

image

image

image

image

image

And here's a video :)

https://github.com/user-attachments/assets/8f90b38d-37a2-4dec-a413-2238c40cb48b


app/src/main/java/com/jerboa/db/entity/AppSettings.kt
```diff
@@ -58,6 +59,11 @@ data class AppSettings(
         defaultValue = DEFAULT_SHOW_BOTTOM_NAV.toString(),
     )
     val showBottomNav: Boolean,
+    @ColumnInfo(
+        name = "enable_infinite_scroll",
+        defaultValue = DEFAULT_ENABLE_IFINITE_SCROLL.toString(),
+    )
+    val enableInfiniteScroll: Boolean,
```
dessalines: This shouldn't need to be added as a jerboa app setting, its already a back-end setting (even in `0.19`): `LocalUser.infinite_scroll_enabled` . 

It needs to be added as a saveable boolean setting in your user settings in the app, not the `Look and feel` section.

app/src/main/java/com/jerboa/feed/PaginationController.kt
```diff
@@ -1,18 +1,36 @@
...
+import java.util.Stack
+
+class PaginationController {
+
+    val previousPageCursors = Stack<PaginationCursor?>()
+    var currentPageCursor: PaginationCursor? = null
+    var nextPageCursor: PaginationCursor? = null
+    var page by mutableLongStateOf(1)
+
+    fun appendPage(nextPage: PaginationCursor?) {
```
dessalines, line 16: This seems to be a dupe of `nextPage` but without pushing to the stack. Why is that necessary?

app/src/main/java/com/jerboa/feed/PaginationController.kt
```diff
@@ -1,18 +1,36 @@
...
+import androidx.compose.runtime.setValue
 import it.vercruysse.lemmyapi.datatypes.PaginationCursor
+import java.util.Stack
+
+class PaginationController {
+
+    val previousPageCursors = Stack<PaginationCursor?>()
+    var currentPageCursor: PaginationCursor? = null
+    var nextPageCursor: PaginationCursor? = null
+    var page by mutableLongStateOf(1)
```
dessalines, line 14: Why did this need to change. I thought `mutableStateOf` only matter for functional components.

app/src/main/java/com/jerboa/feed/PaginationController.kt
```diff
@@ -1,18 +1,36 @@
...
 
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableLongStateOf
+import androidx.compose.runtime.setValue
 import it.vercruysse.lemmyapi.datatypes.PaginationCursor
+import java.util.Stack
+
+class PaginationController {
+
+    val previousPageCursors = Stack<PaginationCursor?>()
```
dessalines, line 11: In `1.0`, we won't need this stack, as the `PagedResponse` returns a `prev_page` and `next_page`, so please add a TODO comment about that.

app/src/main/java/com/jerboa/feed/PaginationController.kt
dessalines, line 1: On thinking about it a bit, we won't need this at all in `1.0` , and can do everything with your `PageButton` alone.

This is because the viewmodels are already storing the `PagedResponse` view, which contains the `prev_cursor` and `next_cursor`. So all that's needed is to pass those into your `PageButton` component, and do an `onPageChange: (cursor?: PaginationCursor)`

app/src/main/java/com/jerboa/model/PostsViewModel.kt
```diff
@@ -114,21 +119,49 @@ open class PostsViewModel(
...
+            ApiState.Loading,
+            onSuccess = pageController::nextPage
+        )
+    }
+
+    fun previousPage() {
+        if (!pageController.canMoveToPreviousPage()) return
+
+        pageController.page--
+        pageController.currentPageCursor = pageController.previousPageCursors.pop()
```
dessalines, line 134: Why not add these as functions to pageController, just like with the others.

app/src/main/java/com/jerboa/model/PostsViewModel.kt
```diff
@@ -51,6 +51,9 @@ open class PostsViewModel(
     private val pageController = PaginationController()
     private val postController = PostController()
 
+    val currentPage: Long
+        get() = pageController.page
```
dessalines, line 55: Why. Just use the pageController.page.

app/src/main/java/com/jerboa/model/PostsViewModel.kt
```diff
@@ -114,21 +119,49 @@ open class PostsViewModel(
...
             state,
+            onSuccess = { pageController.nextPageCursor = it }
         )
     }
 
     fun refreshPosts() = resetPosts(ApiState.Refreshing)
 
+    protected open fun getPreviousPageForm() = GetPosts(
+        page = pageController.page,
+        page_cursor = pageController.currentPageCursor,
```
dessalines, line 156: This looks like the current page, not the previous one. 

I don't think these extra form builders will be necessary in `1.0` also. It will be:

- The `next_page` and `prev_page` are already in the `PagedResponse`, and given to the `PageButton`
- User clicks next or previous on the page button, pushing up the correct `page_cursor`.
- The viewmodel does the fetch given that page cursor, and sets `next_page` and `prev_page` again.
- repeat

app/src/main/java/com/jerboa/ui/components/comment/CommentNodes.kt
```diff
@@ -68,57 +77,98 @@ fun CommentNodes(
...
+                isCollapsedByParent = isCollapsedByParent,
+                showActionBar = showActionBar,
+                enableDownVotes = enableDownVotes,
+                showAvatar = showAvatar,
+                blurNSFW = blurNSFW,
+                voteDisplayMode = voteDisplayMode,
+                swipeToActionPreset = swipeToActionPreset,
+            )
+            item {
+                if (enableInfiniteScroll) Spacer(modifier = Modifier.height(100.dp))
```
dessalines: I know it wasn't done before, but externalize this `dp` like all the others to `Sizes.kt` .

I'm gonna add an issue to do this with the rest.

app/src/main/java/com/jerboa/ui/components/common/Buttons.kt
```diff
@@ -29,3 +33,43 @@ fun RetryLoadingPosts(onClick: () -> Unit) {
         Text(stringResource(R.string.posts_failed_loading))
     }
 }
+
+@Composable
+fun PaginationButton(
```
dessalines: Rename this to `Paginator` or something like that. Especially since its not actually a button, but a row of several buttons.

app/src/main/java/com/jerboa/ui/components/common/Buttons.kt
```diff
@@ -29,3 +33,43 @@ fun RetryLoadingPosts(onClick: () -> Unit) {
...
+    currentPage: Long,
+    onNext: () -> Unit,
+    onPrevious: () -> Unit,
+    onNextEnabled: Boolean = true
```
dessalines: Define this sealed class to help with page types:


```kt
sealed class PageType {
    data class Number(val data: Long) : PageType
    data class Cursor(val data: PaginationCursor) : PageType
}
```

In anticipation of `1.0`, provide it with 3 things.

```kt
current: PageType?,
nextPage: PageType?,
prevPage: PageType?,
```

Then do 

```kt
onNext: (PageType) -> Unit,
onPrev: (PageType) -> Unit,
```

You can remove `onNextEnabled` because you'll know this from giving it the `nextPage`

app/src/main/java/com/jerboa/ui/components/person/PersonProfileScreen.kt
```diff
@@ -681,6 +683,14 @@ fun UserTabs(
                                     swipeToActionPreset = swipeToActionPreset,
                                     disableVideoAutoplay = disableVideoAutoplay,
                                     lowBandwidthMode = lowBandwidthMode,
+                                    enableInfiniteScroll = enableInfiniteScroll,
+                                    onNextPage = {
+                                        personProfileViewModel.navigatePagination(profileId, 1)
+                                    },
+                                    onPreviousPage = {
```
dessalines, line 690: You'll now be able to use the correct cursor from the on page button click

app/src/main/java/com/jerboa/ui/components/post/PostListings.kt
```diff
@@ -41,7 +49,7 @@ import it.vercruysse.lemmyapi.datatypes.PostView
 @OptIn(ExperimentalLayoutApi::class)
 @Composable
 fun PostListings(
-    posts: List<PostView>,
+    posts: List<PostView>?,
```
dessalines, line 52: Just show the message for an empty list, which is what's returned by the api anyway. No need for an option.

app/src/main/java/com/jerboa/ui/components/settings/lookandfeel/LookAndFeelScreen.kt
```diff
@@ -429,6 +431,17 @@ fun LookAndFeelScreen(
                         },
                     )
 
+                    SwitchPreference(
```
dessalines, line 432: Move this to the `AccountSettings` screen.

app/src/main/java/com/jerboa/feed/PaginationController.kt
rodrigo-fm: Seems great. This pull request would serve as legacy support for instances that are running versions older than 1.0.0 ?

app/src/main/java/com/jerboa/ui/components/post/PostListings.kt
rodrigo-fm: I did this to avoid showing "No more posts available" while loading the posts, I thought this could be seen as a bug by some users. This implementation allowed me to show the "No more posts available" only when the response returns an empty list of posts.

app/src/main/java/com/jerboa/db/entity/AppSettings.kt
rodrigo-fm: Only authenticated users would be able to edit this? Would the default be `true` or `false` for unauthenticated users?

Also, I tried fetching the /site endpoint, I even used the api documentation below passing `Authorization Bearer {{token}}` as a header, but I'm unable to retrieve the "my_user" key from the response. How do I fetch this information so that I have a reference on how to adapt the code?

https://mv-gh.github.io/lemmy_openapi_spec/#tag/Site

MV-GH: > Only authenticated users would be able to edit this?

yes

> Would the default be true or false for unauthenticated users?

infinite_scroll_enabled = true

same as anon, or the value when fresh account is created

> Also, I tried fetching the /site endpoint, I even used the api documentation below passing Authorization Bearer {{token}} as a header, but I'm unable to retrieve the "my_user" key from the response. How do I fetch this information so that I have a reference on how to adapt the code?

you prob used either 1.0.0 lemmy or wrong bearer token.

its already in the LemmyAPI, you can just use it

LocalUser.infinite_scroll_enabled

SaveUserSettings.infinite_scroll_enabled

app/src/main/java/com/jerboa/ui/components/post/PostListings.kt
dessalines: > while loading the posts

The screen above this shouldn't load this component until its `ApiState == success`, so loading should already be handled. IE the screen should have a `ApiState == Loading` indicator, that prevents even rendering the `PostListings` component otherwise.

app/src/main/java/com/jerboa/db/entity/AppSettings.kt
rodrigo-fm: ok, I moved it to use the infinite_scroll_enabled from the API. Let me know if something is still missing there

app/src/main/java/com/jerboa/ui/components/common/Buttons.kt
rodrigo-fm: Makes sense. Should I leave it inside `Buttons.kt` or move it to another file? Maybe create a new file inside `ui/components/common/` called `Paginator.kt`?

MV-GH: ye you can move it

app/src/main/java/com/jerboa/feed/PaginationController.kt
dessalines: That's fine. There's no way around me having to merge this into my `1_0_dev` branch anyway.

I'll have to remember to cherry pick this into the `release/v0.19` branch, because we're going to use `main` as a broken but fast-developing 1.0 branch from now on.

app/src/main/java/com/jerboa/db/entity/Account.kt
```diff
@@ -23,6 +23,11 @@ data class Account(
         defaultValue = "0",
     )
     val defaultSortType: Int,
+    @ColumnInfo(
+        name = "infinite_scroll_enabled",
+        defaultValue = "1"
+    )
+    val infiniteScrollEnabled: Boolean,
```
dessalines, line 30: Eventually I'm going to remove all these, but don't worry about that for now, this is fine.

app/src/main/java/com/jerboa/ui/components/comment/CommentNodes.kt
dessalines: This size needs to be defined somewhere, and it should be part of the paginator row margins.

app/src/main/java/com/jerboa/ui/components/comment/CommentNodes.kt
```diff
@@ -68,57 +77,98 @@ fun CommentNodes(
...
+            }
+            item {
+                if (!enableInfiniteScroll) {
+                    PaginationButton(
```
dessalines: Pretty sure I have a comment in the last review about renaming this.

dessalines (commented):
There are still some unaddressed comments from the last review.

Mark the PR as *draft*, then *ready for review* when you're ready for us again.

app/src/main/java/com/jerboa/model/PostsViewModel.kt
rodrigo-fm: The pageController is private and I was passing the currentPage as an argument to some screens (to display on the Paginator composable), but it might not be needed anymore after implementing all the changes suggested here.

app/src/main/java/com/jerboa/ui/components/post/PostListings.kt
rodrigo-fm: I don't know if I understood it correctly, but the `HomeActivity` loads the composable like this on the main branch:

```kotlin
val posts: List<PostView> = when (val postsRes = homeViewModel.postsRes) {
    is ApiState.Failure -> {
        apiErrorToast(ctx, postsRes.msg)
        listOf()
    }

    is ApiState.Holder -> {
        postsRes.data
    }

    else -> {
        listOf()
    }
}
```

The else branch is triggered when `postRes is ApiState.Loading`, so the message "No more posts available" is displayed while loading because the list is empty. Changing the argument type to `List<PostView>?` was the easiest way I found to avoid displaying the message while loading the posts.

The `CommunityScreen` does it differently, though. It only loads the `PostListings` composable when `postRes is ApiState.Holder`. I can apply this logic on the `HomeActivity`, you want it to be like that?

`CommunityScreen` example:

```kotlin
when (val postsRes = communityViewModel.postsRes) {
  ApiState.Empty -> {
      ApiEmptyText()
  }
  
  is ApiState.Failure -> {
      ApiErrorText(postsRes.msg)
  }
  
  is ApiState.Holder -> { /* load composable */ }
```

PR #2176, opened 2026-07-19
Expand Search Functionality

### LINKED ISSUE #27: Search

MV-GH:
Expands the existing search by adding by making its options configurable and having it show Posts, Comments and Users too.

Some additions:
- Prevents search triggers on empty text
- Search ime action closes the (soft) keyboard
- Can now also be used to do User/Posts/Comments selection (also disables changing selection in the filter)
- Scroll to load in more searches
- The errorText was white on white.  So changed the colorScheme key,  imo i think for most it should be that but left those as is for now.
- DEBOUNCE_DELAY fixes warnings as they prefer Duration now

Example

https://github.com/user-attachments/assets/188a36bd-d762-4f93-a85e-d1814b13a7e0

Fixes #27

And I'll finally close that draft PR


dessalines (approved):
Nice, it looks really good. 

The only minor nitpick I have, is that in the future I'd like to try to use [filter chip bars](https://m3.material.io/components/chips/guidelines) for filters and sorts (especially since pages have many more than 2 filtering options now), but that can always be done later. I'd also like your input on that, because it could make us rethink those `TopAppBars`.

There are also some `1.0` considerations for how search works now (IE get rid of the `All` search, show specific filters depending on the content, use the individual `/list` endpoints instead of `/search`). See the [lemmy-ui 1.0 search issue](https://github.com/LemmyNet/lemmy-ui/issues/3995) for context. That can also be left till later.

I'll merge this, then I'll freeze a `release/v0.19` branch.

PR #2185, opened 2026-07-25
Support more image formats (AVIF, HEIC, ...) plus refactors

MV-GH:
I noticed that we didnt support avif images. There is Android resource that lists the supported formats: https://developer.android.com/media/platform/supported-formats#image-formats

I have added these image formats to be recognized as image.

Avif is only supported on Android 14+ so I tweaked the error to be more readable if the image format isnt supported.

see (it shows in overview bc we do downsampling images using pictrs in webp format in feed)

https://github.com/user-attachments/assets/04f07b88-ca3a-4f10-8927-aa1dff928301

On android 16

https://github.com/user-attachments/assets/ea0b4854-7ff3-4365-b5f9-4872d1b1ac92

Then the refactors, I wasnt planning on doing this but it was starting to get very complicated.
I had some small refactors but they werent right bc of the image proxy stuff that was all over the place.
So I centralized that stuff al into PostLinkType, and removed all the logic we had everywhere.
_(I used ktor Url isntead of Android Uri bc jvm vs Android, else would need AndroidTests which is a pain as requires to be run in emulator and thus much slower)_

The video changes were due to the above refactors, I couldn't comprehend much of my earlier choices. The video stuff was being decided differently in three places. So aligned that using PostVideoSource. Also that lowBandwidthMode wasnt exactly correct. It would still fetch the video metadata for placeholder stuff.


dessalines (approved):
The refactors look good, and even though they're minor, its going to be a nightmare trying to get them into my dev branch. Unavoidable I spose.

